Issue #109, opened 2026-05-12
Builder /nix-store cache: block-device backing on macOS


auser:
## Context

We just shipped a persistent host-backed `/nix-store` cache for the
microsandbox builder VM (`mvm_build::builder_vm::run_build_async` +
`builder_store_dir()`), but on macOS the bind-mount path through
libkrun's virtio-fs proxy strips `setxattr`. nix uses `setxattr` to
mark chroot-store paths, so `nix build` fails on the very first
store-path write with EIO. As a stopgap, `run_build_async` now
auto-defaults `force_tmpfs: true` on macOS — correctness preserved,
but warm-cache benefit lost (every cold build is still ~25 min).

This issue tracks moving macOS onto the same warm-cache path Linux
gets, by attaching the cache as a virtio-blk device backed by a
host-resident raw ext4 file. ext4 inside the guest = native xattr
support, bypassing libkrun's filesystem proxy entirely.

## Precedent

Two existing call sites already use this pattern for the same reason:

- `crates/mvm-backend/src/microsandbox.rs:190` — rootfs `.ext4`
  attached via `.image_with(|i| i.disk(&disk_path).fstype(\"ext4\"))`
- `crates/mvm-providers/src/apple_container/macos.rs:588-616` — the
  dev VM's `~/.mvm/dev/nix-store.img` attached as `/dev/vdb` (Apple
  Virtualization Framework path). The comment there documents the
  exact same xattr motivation: \"virtiofs on macOS can't surface
  trusted.* xattrs, so an overlay over a virtiofs upper silently
  downgrades to read-only.\"

## API surface

microsandbox 0.4.5 exposes everything on `MountBuilder`
(`lib/sandbox/types.rs:310-340`):

```rust
m.disk(host_image_path)
 .format(DiskImageFormat::Raw)
 .fstype(\"ext4\")
```

agentd auto-mounts at the guest path; no virtio-fs proxy involved.

## Work shape (~1 day)

1. Rename `builder_store_dir() → builder_store_image()`, returning
   `~/.cache/mvm/builder-store.img` instead of a directory.
2. Add `ensure_builder_store_image(path, size)` — provisions a sparse
   ext4 file at the path the first time it's requested, runs
   `mkfs.ext4` on it once. Tests cover the create-and-format and the
   no-op-when-present paths.
3. In `run_build_async`, swap the `Some(...)` branch from
   `m.bind(host_dir.as_path())` to
[...]

PR #117, opened 2026-05-13
feat(dev-image): kernel modules in rootfs, drop custom kernel override

### LINKED ISSUE #110: Ship kernel modules in dev-image rootfs; drop custom kernel override
## Context

The dev-image flake at `nix/images/builder/flake.nix` overrides
`pkgs.linuxPackages.kernel` with `structuredExtraConfig` to compile
vsock support **built-in** (rather than as modules). The override
exists because the rootfs `mkGuest` emits is a flat busybox tree
with no `/lib/modules/`, so vsock-as-module can't load — which would
break the guest agent's vsock surface, and with it
`mvmctl console`, `mvmctl dev shell`, and `mvmctl build`.

The cost is severe: any kernel `.override` produces a derivation
hash that misses `cache.nixos.org`, so every cold build of the dev
image compiles the full Linux kernel from source. On aarch64 with
4 vCPUs inside the microsandbox builder, that's ~20-25 minutes per
cold build — the dominant 95% of total build time.

This issue tracks dropping the override and shipping a tiny
`/lib/modules/` tree in the rootfs so we can `modprobe` vsock at
boot from the stock binary-cached kernel.

## Expected impact

Cold build on macOS: **~25 min → ~2 min**.

- Kernel: stock `linuxPackages.kernel`, cache hit at
  `cache.nixos.org` (~30 s download).
- Rust toolchain + cargo-vendor closure: ~60 s (also cache-hits the
  stock derivation).
- Rootfs assemble + modules tree + nix evaluation: ~30 s.

This is **independent of the block-device cache work** (filed
separately). Combined with that issue, warm builds on macOS would
collapse to <30 s.

## Work shape (~half a day)

1. **mkGuest change** (`nix/lib/mk-guest.nix` or wherever the rootfs
   assembly lives): when `kmod` is in the closure, install
   `\${kernel}/lib/modules/\${version}/` into the rootfs at the
   matching path. The modules tree is ~5 MB (vsock + its
   transitive deps) — comfortably small enough that we don't need
   to gate it.
2. **Init script** (`crates/mvm-cli/resources/baseline.nix` or the
   matching mkGuest entrypoint stub): early-boot `modprobe vsock
   vhost_vsock vmw_vsock_virtio_transport` before the agent
   socket is opened. Order matters — vsock has to be loaded
   before the agent calls `socket(AF_VSOCK, ...)`.
3. **Drop the override**: revert
   `nix/images/builder/flake.nix:97-117` (the `devKernel` block)
   to `kernelPkg = pkgs.linuxPackages.kernel;`. Remove the now-
   obsolete vsock-built-in commentary.
4. **Re-run \`cargo xtask build-dev-image\`** to regenerate the
   vendored slot at `nix/images/dev-prebuilt/aarch64/` plus its
   `checksums-sha256.txt` (the new hashes prove the rebuild
   actually used the stock kernel).
5. **Trust-boundary doc**: update the flake header to remove the
   \"custom kernel ~25 min\" commentary and replace with the
   \"modules in rootfs, kernel from binary cache\" rationale.

## Acceptance criteria

- [ ] `cargo xtask build-dev-image` on a clean
      `~/.cache/mvm/builder-store/` completes in under 5 minutes on
      aarch64 macOS.
- [ ] Boot of the resulting image successfully `modprobe`s vsock
      modules before the guest agent starts.
- [ ] `mvmctl dev up && mvmctl dev shell` works end-to-end.
- [ ] `mvmctl console <vm>` works (validates the vsock console
      surface against the modular kernel).
- [ ] Rootfs size growth stays under 10 MB compared to the current
      built-in-vsock build.

## Risks

- **Stock kernel vsock module ABI drift.** The vsock kABI is
  stable in upstream Linux, but the *exact* set of modules to
  `modprobe` varies by version (e.g., `vmw_vsock_virtio_transport`
  on some, `vsock_loopback` opt-in on others). The init script
  needs to be defensive: \"modprobe each, log + continue on
  failure of any optional one.\"
- **Rootfs no-modules invariant.** Some security-posture code may
  assert no `/lib/modules/` exists (defence in depth against
  module-loading attacks). Audit `crates/mvm-guest/` for any such
  assertions; if present, scope-narrow them to runtime VMs (which
  still want no modules) and exempt the dev image.
- **Verified-boot interaction.** The dm-verity sidecar covers the
  rootfs Merkle tree — adding modules to the rootfs changes the
  hashes, which is fine, but the W3 verified-boot path needs a
  re-test to confirm the modules tree doesn't break verity
  semantics. (Likely orthogonal; dev image isn't verity-signed.)

## Related

- Block-device builder-store backing on macOS (filed separately) —
  warm-cache pathway, complementary to this cold-path optimisation.
- The 25-min cold-build pain that motivated this:
  `nix/images/builder/flake.nix:97-117` (current `devKernel`
  override block).

auser:
## Summary

- Drops the `pkgs.linuxPackages.kernel.override` (custom vsock-built-in config); ships matching `/lib/modules/<kver>/` in the rootfs and `modprobe`s vsock at boot instead. Cold dev-image build collapses from ~25 min to **~5 min** on aarch64 macOS (stock kernel = `cache.nixos.org` binary hit).
- Adds groundwork for persistent host-backed `/nix-store` cache (warm builds in seconds on Linux; macOS auto-defaults to tmpfs until the block-device follow-up — see #109).
- Salvages improvements from a parallel-session attempt at the same problem (reflink module copy, dead-symlink cleanup), bundles the worktree git-metadata fix + crates.io egress fix surfaced during verification.

Closes #110.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo test --workspace` passes (~667 tests)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo run -p xtask -- build-dev-image` cold: **274 sec** end-to-end (kernel cache-hit, crates.io reached, modules tree copied, vendored slot emitted with checksums)
- [x] `mvmctl dev up` boots cleanly; `mvmctl dev status` returns `Kernel: 6.12.87` via vsock RPC
- [x] `mvmctl console mvm-dev --command 'lsmod'` shows the vsock chain loaded:
      ```
      vsock_loopback                    Live
      vmw_vsock_virtio_transport        Live  (depends on common)
      vmw_vsock_virtio_transport_common Live  (depends on vsock)
      vsock                             Live  (4 deps)
      ```

## What's in the commit chain

```
74ddece fix(mk-guest): pull modules from kernel's `modules` output, not `$out`
2650318 fix(builder): bind-mount git-worktree metadata into the sandbox
7e02b3b chore(mk-guest): reflink module copy + drop nixpkgs source/build symlinks
ebb5125 chore: trim scratch.md to active open questions               ← orthogonal housekeeping
66d5e50 docs(research): park catalog & schema versioning notes        ← orthogonal housekeeping
5df788e chore(lima-cleanup): finish dev-VM rename in docstring        ← orthogonal housekeeping
811c2c3 fix(builder): allow_all egress + tighten xattr regex
7b79883 test(audit): live tests via mock-VM fixture                   ← duplicate of a87f0a1
8b299e2 feat(dev-image): ship vsock kernel modules in rootfs (#110)
2e87267 feat(builder): persistent shared /nix-store cache
```

**Note on `7b79883`**: it's a cherry-pick of the same content already on `feat/mock-backend` as `a87f0a1`. The diff is empty (GitHub may collapse it) but it shows in the commit list. Happy to rebase-drop it if it's noise.

## Known follow-ups

- **Rootfs size**: grew from ~770 MB → 1.05 GB (~280 MB modules tree, most of which we never load). A trim to just `vsock` + deps + regenerated `modules.dep` would bring it back below the previous floor; tracking as a TODO if you want a follow-up PR.
- **#109** (block-device builder cache for macOS) is unaffected by this — separate issue, separate work.

## What this doesn't change

- Runtime microVMs (production path) — they have their own `mkGuest` calls without `kernel = …`, so the modules-install block is gated off.
- The trust boundary / security claims in CLAUDE.md — same posture, just different kernel-vsock provisioning.


auser:
## Modules-tree trim landed (73319aa)

Resolved the rootfs-size concern from the original PR body. Walks
`modprobe --show-depends` for the boot-time vsock closure, prunes
everything outside, regenerates `modules.dep` via `depmod -b`.

### Results

| Layer | Before trim | After trim |
|---|---|---|
| Rootfs (`rootfs.ext4`) | 1.05 GB | **774 MB** (−279 MB) |
| In-guest `/lib/modules/` | ~280 MB | **812 KB** |
| Cold build | 274 sec | 295 sec (+21 sec for modprobe-walk + depmod) |
| `vmlinux` SHA | unchanged | unchanged |

Net rootfs is now within **~4 MB** of the pre-#110 floor (770 MB) —
i.e. we get the cold-build speedup (~25 min → ~5 min) at essentially
the same rootfs size as the old custom-kernel build.

### Re-verified end-to-end

`mvmctl dev up` boots cleanly. `mvmctl dev status` returns
`Kernel: 6.12.87` via vsock RPC. `lsmod` shows the same four modules
loaded as before:
```
vsock_loopback                    Live
vmw_vsock_virtio_transport        Live  (depends on common)
vmw_vsock_virtio_transport_common Live  (depends on vsock)
vsock                             Live  (4 deps)
```

In-guest `ls /lib/modules/*/kernel/net/` returns just `vmw_vsock` —
the trim removed every other kernel subtree.

### To extend the closure

If we ever need to `modprobe` something else at boot, append the
module name to the `keepModules` loop in `mk-guest.nix`. The
closure walk picks up all transitive deps automatically.

### Open follow-up

`/init`'s modprobe loop currently tries `vsock vmw_vsock_virtio_transport
vsock_loopback` — but the trim's closure-walk only seeds from
`vmw_vsock_virtio_transport` + `vsock_loopback` (the latter pulls in
`vsock` via deps). Adding more modules to /init's loop without also
adding them to the trim's keep list would be a silent boot-time
failure. I'd consider centralising the list in one place (a let-binding
near the top of mkGuest) as a small follow-up cleanup, but not
load-bearing.

PR #176, opened 2026-05-14
Plan 72 W2+W3: builder VM image + mvm-builder-init

auser:
## Summary

- **Plan 72 W2 (commit `89b0444`):** New `nix/images/builder-vm/` flake produces the Layer-1 builder VM image (`vmlinux` + `rootfs.ext4` + `manifest.json`) that `LibkrunBuilderVm` (Plan 72 W1, blocked on Plan 57 W3) will boot into to run `nix build` for mvmctl. Replaces microsandbox + `nixos/nix:2.24.10` OCI bootstrap on the user-facing path per ADR-046. New `builder-vm-image` job in `release.yml` publishes the artifacts.
- **Plan 72 W3 (commit `6846115`):** New `crates/mvm-builder-init/` Rust crate is PID 1 inside that VM — mounts pseudofs, formats + mounts `/dev/vdb` as the persistent `/nix` store on first boot, runs `/job/cmd.sh`, writes the exit code to `/job/result`, powers off via `reboot(POWER_OFF)`. `nix/packages/mvm-builder-init.nix` swaps from the W2 shell stub to `pkgsStatic.rustPlatform.buildRustPackage`. Same `$out/sbin/mvm-builder-init` contract so the flake doesn't move.

## Notes for review

- **No `mkGuest`.** The workload-oriented contract (guest-agent fork under uid 990, `/etc/mvm/entrypoint`, setpriv-to-uid-1000) doesn't fit a single-shot build VM. Direct rootfs assembly is ~50 lines (extracted to `files/assemble-rootfs.sh`) and avoids baking ~30 MiB of unused `mvm-guest-agent` into the image.
- **Templates instead of inline strings.** Every multi-line shell body and every templated artifact lives in its own file under `nix/images/builder-vm/files/` or `nix/packages/`. The `.nix` files are orchestration only.
- **macOS compatibility.** `mvm-builder-init` is Linux-only at runtime but compiles to a stub on macOS so `cargo build --workspace` keeps working. Init logic in `src/init.rs` is gated `#[cfg(target_os = "linux")]`.
- **Uses `libc` directly**, not the `nix` crate. The syscall surface (`mount`, `reboot`) is two FFI calls; the workspace already pins `libc`.

## Known follow-ups (intentionally deferred)

- **`/job` mount inside the guest.** Production uses virtio-fs (the host attaches it pre-boot via `LibkrunBuilderVm`). The qemu smoke test would need a `stage_job_mount_9p` to make `/job` writable in that test context. Documented in `tests/qemu_smoke.rs` as a TODO; the test file is a skeleton with `#[ignore]`.
- **Compressed size budget.** The flake enforces the 1.2 GiB *uncompressed* cap from W2 §Acceptance. The 300 MiB compressed budget needs a `zstd` step in `release.yml`, which isn't there today for any of the existing image jobs (dev-image, builder-image, default-microvm). Aligns better with a workflow-wide compression pass than a one-off.

## Test plan

- [ ] `cargo build --workspace` passes on macOS aarch64 (stub `mvm-builder-init` path)
- [ ] `cargo build --workspace` passes on Linux aarch64 (real init build)
- [ ] `cargo clippy --workspace -- -D warnings` clean
- [ ] `cargo test -p mvm-builder-init` passes the non-ignored tests (`status_truncation_respects_char_boundary`, `blkid_fstype_returns_none_when_blkid_absent`)
- [ ] On a Linux runner: `nix build ./nix/images/builder-vm#packages.aarch64-linux.default` produces `vmlinux` + `rootfs.ext4` + `manifest.json`
- [ ] On a Linux runner: `nix build ./nix/images/builder-vm#packages.x86_64-linux.default` does the same
- [ ] `manifest.json` validates as JSON and contains the expected schema/system/kernelCmdline/initIsStub fields
- [ ] `rootfs.ext4` stays under 1.2 GiB uncompressed (enforced by build; build fails if exceeded)
- [ ] Release workflow run on a tag uploads `builder-vm-vmlinux-{arch}`, `builder-vm-rootfs-{arch}.ext4`, `builder-vm-{arch}.manifest.json`, and the checksums file

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -245,6 +245,59 @@ jobs:
             staging/builder-image-${{ matrix.arch }}-checksums-sha256.txt
             staging/builder-image-${{ matrix.arch }}.manifest.json
 
+  # Plan 72 W2 / ADR-046: the Layer-1 builder VM image. This is the VM
+  # `LibkrunBuilderVm` boots into to run `nix build` on behalf of mvmctl,
+  # replacing the microsandbox `nixos/nix:2.24.10` OCI bootstrap on the
+  # user-facing path. Same per-arch matrix as dev-image; uploads the
+  # vmlinux + rootfs + manifest.json the flake emits.
+  builder-vm-image:
+    strategy:
+      matrix:
+        include:
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+          - arch: x86_64
+            runner: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
+
+      - name: Build builder VM image
+        env:
+          ARCH: ${{ matrix.arch }}
+        run: |
+          SYSTEM="${ARCH}-linux"
+          nix build "./nix/images/builder-vm#packages.${SYSTEM}.default" \
+            --no-link --print-out-paths > /tmp/store-path.txt
+          STORE_PATH=$(head -1 /tmp/store-path.txt)
+          mkdir -p staging
+          cp "$STORE_PATH/vmlinux"       "staging/builder-vm-vmlinux-${ARCH}"
+          cp "$STORE_PATH/rootfs.ext4"   "staging/builder-vm-rootfs-${ARCH}.ext4"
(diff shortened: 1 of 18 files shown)


auser:
Closing — main already has Plan 72 W2/W3/W4 via #172 (commit 31a4b41). This PR was written in parallel and is duplicate/conflicting work. See the PR discussion for details.

PR #280, opened 2026-05-15
plan 77 W0 — Stage 0 invariant + LibkrunBuilderVm image override + plan doc

auser:
## Summary

Lays the groundwork for **Plan 77** — restoring a local Stage 0 path for the builder VM image after PR #230's microsandbox removal, without falling back to a published prebuilt artifact. W0 is the invariant + refactor + plan; W1+ wires the actual `bootstrap_via_dev_image_stage0` function.

## What's in W0

**Hard invariant recorded** in both `AGENTS.md` and `CLAUDE.md`:

> No prebuilt builder VM artifact is downloaded when developing locally, and none is published until we explicitly cut a release.

This is stricter than ADR-046's existing "source-checkout builds never depend on mvm-published artifacts" — it also bans the *fallback* download path that previously silently took over on cache miss, and bans publishing the artifact in CI until release.

**`LibkrunBuilderVm` refactor:** `BuilderVmImage::new` is now public, and `LibkrunBuilderVm::with_image_override(image)` lets `run_build` skip the `~/.cache/mvm/builder-vm/<arch>/` lookup when the caller supplies an explicit kernel + rootfs + cmdline. No behavior change — nothing calls it yet.

**Plan 77 spec** at `specs/plans/77-stage0-bootstrap-via-dev-image.md`:
- W0..W4 sequencing
- **Twelve numbered security considerations** (trust boundary, r/o workspace mount, staged-then-promoted cache, no agent surface enlargement, libkrun TSI posture, advisory lock for concurrency, bounded exec timeout, audit chain entries, no source-checkout download fallback, `do_exec` scope, verity-sidecar handling, flake-injection threat model)
- Test matrix (unit + integration)
- Rollback plan
- Explicit out-of-scope (mvm-oci Stage 0 is Plan 75 territory; DNS-in-VM is Plan 72 W5.D territory)

## Why this slice ships first

W1's implementation will be ~200–300 lines that boot the dev image, use vsock RPC, validate output, atomically promote the cache, and rip out the source-checkout download path. That work needs:
- A landed invariant doc so the patch can reference it as the binding constraint
- A landed refactor of `LibkrunBuilderVm` so the W1 patch is purely additive to `apple_container.rs`
- Reviewers to agree on the security posture *before* the implementation lands

W0 establishes all three.

## What's NOT in this PR

- The `bootstrap_via_dev_image_stage0` function itself — that's W1
- Any change to `mvmctl dev up` behavior — current "no Stage 0 path" failure mode is unchanged
[...]

PR #293, opened 2026-05-15
plan 77 W1 — Stage 0 seeds from ~/.mvm/dev/current/

auser:
## Summary

- Extends `find_local_fallback_image` in `crates/mvm-cli/src/commands/env/apple_container.rs` to also consider `~/.mvm/dev/current/{vmlinux,rootfs.ext4}` as a Stage 0 seed candidate
- Updates two error messages (Stage 0 hard-error + `download_dev_image` dead-letter context) to name all three searched paths
- Adds three unit tests in `dev_status_image_tests` covering the load-bearing case, mtime-based ordering against `prebuilt/`, and the empty-case smoke test

## Context

Most of Plan 77 W1 already shipped: commits 843ef18, 4bbb615, a8f47e9, 0aac0f2, 68ae7db wired up `bootstrap_builder_vm_image_via_dev_image_stage0` on top of `LibkrunBuilderVm::run_build` + `with_image_override`, plus atomic staging, source-fingerprint binding, and the artifact-digest manifest.

The remaining bug: `find_local_fallback_image`, the helper that locates a Stage 0 seed image, only searched `~/.mvm/dev/prebuilt/v*/` and `~/.mvm/dev/builds/<hash>/`. It missed `~/.mvm/dev/current/` — the canonical "live" path that `build_image_via_libkrun` writes and `resolve_dev_status_image` reads. So a contributor whose only dev image was at `dev/current/` (typical after a manual `rm -rf ~/.cache/mvm/builder-vm/`) got a hard error even though a valid seed was sitting right next to it.

The fix shares one `consider(dir, label)` validation+record path between the new `current/` candidate and the existing `prebuilt/` + `builds/` loops, so the size-floor + ext4-magic gate stays consistent across all three sources.

## Why not the prompt's literal design

The prompt described a brand-new `bootstrap_via_dev_image_stage0(workspace_root, arch)` built on `LibkrunBackend::start` + `dispatch_in_session` + `tear_down_session_vm`. That approach pre-dates 843ef18, which shipped equivalent functionality through `BuilderVm::run_build` + `with_image_override` (a thinner, already-tested layer). Re-implementing it the prompt's way would duplicate that machinery and break the W2/W3 hardening (4bbb615, 0aac0f2, 68ae7db) that hangs off the `BuilderVm` path.

## Test plan

- [x] `cargo test -p mvm-cli --lib -- dev_status_image_tests builder_vm_bootstrap_tests` — 27 pass (incl. 3 new)
- [x] `cargo test --workspace` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — zero warnings
- [ ] **End-to-end on the host that hits the bug** (cannot run from CI):
  - `rm -rf ~/.cache/mvm/builder-vm/aarch64/`
  - confirm `~/.mvm/dev/current/{vmlinux,rootfs.ext4}` exists
  - `cargo run -- dev up`
  - expected: Stage 0 picks `current/` as the seed, builds the builder VM image into a staging dir, atomically promotes to `~/.cache/mvm/builder-vm/<arch>/`, then the normal dev-up dispatch runs
  - (DNS-in-VM failures during the inner `nix build` are Plan 72 W5.D, out of scope here)


auser:
End-to-end verified on this host (macOS arm64, source checkout):

```
$ touch ~/.mvm/dev/current/rootfs.ext4 ~/.mvm/dev/current/vmlinux  # force current/ to win mtime
$ cargo run -- dev up
[mvm] Building dev image via libkrun builder VM (Plan 72 W5) from: …/nix/images/builder
[mvm] Builder VM image not in cache; building locally from …/nix/images/builder-vm...
[mvm] Using local dev image cache (current) as Stage 0 bootstrap image.
```

`(current)` is the label this PR adds. Before the fix, the same invocation hard-errored with "no local dev image cache was found under ~/.mvm/dev/prebuilt/v*/ or ~/.mvm/dev/builds/*/" even though `~/.mvm/dev/current/{vmlinux,rootfs.ext4}` were sitting right there.

A separate run without the `touch` correctly picked the freshest candidate (`prebuilt/v0.14.0`, which had an mtime ~1 hour newer than `current/` on this host) — confirming the mtime ordering across all three search paths works as the `fallback_image_prefers_most_recent_candidate` test asserts.

The dev-up then hit an unrelated `mvm-libkrun-supervisor binary not found` error in the downstream `LibkrunBuilderVm::run_build` step — that's a host setup gap (`cargo install --path crates/mvm-libkrun --features libkrun-sys` or `MVM_LIBKRUN_SUPERVISOR_PATH`), not a regression introduced by this PR. The Stage 0 seed-image selection itself works end-to-end.

Also added: `specs/plans/77-stage0-bootstrap-via-dev-image.md` (rescued from the now-closed W0 PR #280; status header refreshed to reflect what's actually on main).

Issue #95, opened 2026-05-07
ADR-0018 in-guest binaries: mvm-mesh-dns + mvm-mesh-vsock-bridge


auser:
Implement the two new in-guest binaries that make `db.mesh.local:5432`-style addressing work for consumers without any iroh code in the guest. Paired with mvmforge ADR-0018 (composable attested addons) and ADR-0020 (workload mesh).

Cross-repo tracker: tinylabscom/mvmforge#38.

## Contract

The contract spec lives at `specs/contracts/in-guest-mesh-dns.md` (currently untracked in this repo's working tree; landed during the mvmforge addon work and ready to commit). Both binaries are iroh-free; vsock stays the only host↔guest data path (preserves the existing `mvm-guest-agent` vsock-narrow-surface).

### `crates/mvm-mesh-dns/` (new)
- Thin wrapper over `hickory-dns`.
- Listen on `127.0.0.1:53` and `::1:53` only — no TAP, no public interface.
- Authoritative for `*.mesh.local`; forward everything else upstream.
- Per-instance zone loaded from the config disk's `mesh_dns_zone`; SIGHUP reload via the existing guest-agent vsock RPC.
- Supervised by minimal-init alongside `mvm-guest-agent`.
- Target ~200 LOC + tests.

### `crates/mvm-mesh-vsock-bridge/` (new)
- TCP↔vsock bridge for outbound mesh traffic.
- Binds loopback-range IPs from \`mesh_loopback_bindings\` (config disk).
- On accept: opens a vsock stream to mvmd-agent, writes a length-prefixed JSON peer header, then proxies bytes both ways. Half-close semantics preserved.
- **Capability tokens NEVER appear on the guest** — they're attached by mvmd-agent on the host side.
- Target ~100 LOC + tests over libc + tokio.

### Init plumbing
- Extend `nix/lib/minimal-init/lib/05-networking.sh.in` to set `/etc/resolv.conf` listing `127.0.0.1` first when mesh is enabled.
- Bake both binaries unconditionally into `mkGuest` (always-install + no-op when zone empty); no new `mkGuest` argument.
- Extend `crates/mvm-core/src/config.rs` to formalize the config-disk JSON shape for `mesh_dns_zone` + `mesh_loopback_bindings`.

## Acceptance criteria

- [ ] Boot a consumer with no `addons[]`: both binaries start but are no-ops; existing networking unchanged.
[...]

PR #350, opened 2026-05-19
[codex] Serve exact addon DNS records

auser:
Summary:
- Adds a loopback-only UDP DNS server to mvm-addon-dns.
- Answers exact configured production-equivalent hostnames authoritatively as A records.
- Forwards every non-configured name to an explicit upstream resolver snapshot.
- Refuses non-loopback bind addresses and self-upstream recursion.
- Removes the unused hickory-server dependency.

Security posture:
- Guest-loopback DNS only.
- No host-side DNS listener.
- No remote microVM DNS exposure.
- No distributed resolver or mesh plane.

Validation:
- cargo fmt --check
- cargo test -p mvm-addon-dns -p mvm-addon-vsock-bridge
- cargo clippy -p mvm-addon-dns -p mvm-addon-vsock-bridge --all-targets -- -D warnings
- cargo check --workspace
- cargo test --workspace
- cargo tree -p mvm-addon-dns showed no hickory-server or iroh matches

Refs #95


auser:
## Security hardening follow-up

I reviewed the DNS addition specifically for listener exposure, recursive forwarding behavior, and upstream response handling. I pushed commit 547c64b to tighten the forwarding path:

- UDP forwarding sockets now connect to the selected upstream before sending, so the socket only accepts packets from that peer.
- Upstream replies are decoded and must be DNS responses with the same transaction ID as the forwarded query.
- If an upstream reply is malformed, from the wrong transaction, or otherwise mismatched, the addon DNS server treats that upstream attempt as failed and returns SERVFAIL if no valid upstream answers.
- Responses only advertise recursion availability when upstream resolvers are configured.
- Added a regression test for wrong-transaction upstream replies.

Local validation after this hardening commit:

- cargo fmt --check
- cargo test -p mvm-addon-dns
- cargo clippy -p mvm-addon-dns --all-targets -- -D warnings
- cargo check --workspace

PR #351, opened 2026-05-19
fix(libkrun): enable libkrun-sys by default for mvmctl binary

auser:
## Summary

Plan 86 / Plan 72 W5.D bullet 10 added `extract_bundled_kernel()` calling libkrunfw's `krunfw_get_kernel` C API. That call site is the only path to a known-good TSI-patched kernel on macOS — the in-repo port (`nix/images/builder-vm/kernel/`) applies to nixpkgs 6.12.87 cleanly but kernel-oops's at `tsi_release` on socket close.

`mvm-cli` keeps `libkrun-sys` as an opt-in library feature (commit e7a445e5) so mvmd and other library consumers don't pull libkrunfw headers. The root `mvmctl` binary has no such constraint — every contributor `cargo run -- dev up` needs the FFI live. This adds `libkrun-sys` to the root `default` feature so the binary path is wired end-to-end without a manual `--features libkrun-sys` flag.

## Test plan

- [x] `cargo build` (default features) — clean
- [x] `cargo run -- dev up` reaches `extract_bundled_kernel` (visible in `[mvm] Extracted libkrunfw kernel (22675456 bytes)` log line)
- [x] libkrunfw kernel boots in libkrun without the SP/PC alignment exception the in-repo TSI port hit
- [x] `mvm-builder-init` runs through pseudofs mounts → /nix-store seed → cmd.sh dispatch

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -311,7 +311,13 @@ which.workspace = true
 # libkrun isn't installed on the host, so a default-features binary works
 # everywhere. Library consumers (mvmd) that want a lean closure build with
 # `default-features = false`.
-default = ["builder-vm"]
+default = ["builder-vm", "libkrun-sys"]
+# Plan 86 / Plan 72 W5.D bullet 10: extract libkrunfw's bundled
+# TSI-patched kernel at runtime. Wired here on the root `mvmctl`
+# binary by default so the contributor `cargo run -- dev up` path
+# reaches `extract_bundled_kernel`. `mvm-cli` keeps it as an
+# opt-in library feature (library consumers don't link libkrunfw).
+libkrun-sys = ["mvm-cli/libkrun-sys"]
 contributor-bootstrap = [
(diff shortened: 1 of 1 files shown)


auser:
Closing in favor of the Plan 87 implementation series (PR #352).

The libkrun-sys default-on for the mvmctl binary breaks CI on Linux runners that don't have libkrunfw headers installed (build.rs panics in mvm-libkrun). Making it conditional on macOS via Cargo features alone isn't clean (Cargo features aren't cfg-gated).

The proper fix lives in the Plan 87 work — that plan needs CI to install libkrun + libkrunfw + passt anyway for the passt-supervisor integration tests, so the libkrun-sys default-on lands there alongside the CI dep install. Contributors who need the libkrunfw kernel extraction today can still pass `--features libkrun-sys` explicitly.

The mvm-cli library opt-in stance (commit e7a445e5) is preserved.

PR #363, opened 2026-05-19
feat(builder): add Cachix substituter for Stage 0 acceleration (Plan 89)

auser:
## Summary

- Adds `https://mvm.cachix.org` to the in-VM `cmd.sh` `NIX_CONFIG` (`crates/mvm-build/src/libkrun_builder.rs`) so a fresh `mvmctl dev up` substitutes the four self-built derivations (TSI/passt kernel, `mvm-builder-init`, `mvm-egress-proxy`, `mvm-guest-agent`) from our public Cachix cache instead of source-building them. Stage 0 has free egress; steady state falls back to local `/nix` warmed at Stage 0.
- Adds `connect-timeout = 5` to the same `NIX_CONFIG` so a steady-state cache miss (firewall-dropped) fails fast instead of hanging on the default ~63s TCP SYN-retry.
- Adds `.github/workflows/cache-push.yml` — on main merge, build the `nix/images/builder-vm` closure for both arches and push to Cachix. Path-filtered so doc-only edits skip the job.
- Adds `specs/plans/89-cachix-substituter-stage0.md` with the scope (Stage 0 only), trust model (signed closures), and what this does NOT help (steady-state user-flake builds, contributor input-hash changes, host-side Rust edits).

## One-time user provisioning before this is functional

Plan 89 §\"One-time user provisioning\" — short version:

1. Create cache `mvm` at https://app.cachix.org/ on the open-source plan (free).
2. `cachix generate-keypair mvm` — save private key in 1Password.
3. Replace `<USER_PROVISIONS_THIS>` in `crates/mvm-build/src/libkrun_builder.rs` with the printed **public** key. (Single place — the workflow uses `cachix-action` with the auth token, not the literal key.)
4. Add `CACHIX_AUTH_TOKEN` (from `cachix authtoken`) to Settings → Secrets → Actions.

Until step 3, signed closures from the real cache get rejected (wrong public key) — same effective behavior as no Cachix at all. Until step 4 the CI push job fails closed.

## Why no ADR

Three reasons in Plan 89 §\"Why no ADR\" — same trust mechanism as the existing `cache.nixos.org` substituter, bounded blast radius, narrow Stage-0-only scope. ADR-056 is the next free slot if reviewers want one.

## Stacks on top of

Independent of PR #358 (`max-jobs = auto`) — same `NIX_CONFIG` block but different lines. Either order merges cleanly; both wins are additive.

## Test plan

- [x] `cargo test -p mvm-build --lib --features builder-vm libkrun_builder::tests::stage_job_dir_writes_cmd_sh_with_escaped_inputs` — pass (asserts the three new `NIX_CONFIG` entries are emitted)
- [x] `cargo test --workspace --all-features` — green
- [x] `cargo clippy --workspace --all-features --tests -- -D warnings` — clean
- [ ] After user provisioning lands, verify a fresh `mvmctl dev up` hits the cache (closure substituted, not source-built)

### DIFF
--- a/.github/workflows/cache-push.yml
+++ b/.github/workflows/cache-push.yml
@@ -1,0 +1,79 @@
+name: cache-push
+
+# Plan 89 — push the builder VM closure to Cachix on every `main`
+# merge so contributor `mvmctl dev up` and fresh CI runners can
+# substitute instead of source-building the four heavy derivations
+# (TSI/passt kernel + mvm-builder-init + mvm-egress-proxy +
+# mvm-guest-agent). Stage 0 has free egress and reads from the
+# cache; steady state cannot reach it (iptables OUTPUT default-deny
+# per ADR-047) and falls back to its warm local /nix.
+#
+# Requires the `CACHIX_AUTH_TOKEN` repo secret (Plan 89 one-time
+# user provisioning). Without the secret the job fails closed,
+# which is preferable to silently shipping unsigned closures.
+#
+# Security: this workflow consumes no PR/issue/comment fields and
+# nothing user-controlled. The only ${{ … }} interpolations are
+# matrix.arch (set by this file) and secrets.CACHIX_AUTH_TOKEN
+# (repo secret). Per the GitHub Actions injection hardening guide,
+# `matrix.arch` is forwarded through `env:` to a shell variable
+# rather than spliced into the run script.
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


auser:
Scope rolled back: keeping the parallelism win in #358 only; not adding any remote cache. Branch retained in case we revisit.

PR #537, opened 2026-06-03
fix(mkFunctionService): function workload PID 1 must idle, not be the runner (Plan 120 Task 4)

auser:
## The bug (Plan 120 Task 4 blocker)

The libkrun core demo's `up → vsock Ping` step intermittently failed with `Guest agent not reachable`. Root-caused (via `MVM_KRUN_LOG=trace`) to the **workload's PID 1 exiting at ~5s** → kernel reboot (`reboot: Failed to start orderly reboot`):

```
/init: /etc/mvm/entrypoint: line 1: /usr/lib/mvm/wrappers/runner: not found
mvm: /etc/mvm/entrypoint returned without exec — kernel will panic
```

A **function** workload's PID 1 is supposed to be the *idle* bootScript (`exec sleep infinity`) that the compile path passes via mkGuest `entrypoint.command` (`crates/mvm-sdk/src/compile/flake.rs:140`); the **agent** then execs the runner *per call* over vsock `RunEntrypoint`. But `mkFunctionService` *also* wrote `/etc/mvm/entrypoint = /usr/lib/mvm/wrappers/runner` in its `extraFiles`, and that override won — so PID 1 became the one-shot runner, whose `#!/usr/bin/env python3` shebang doesn't resolve in the busybox rootfs (no `/usr/bin/env`). `/init` fell through to `sleep 5` and the kernel rebooted at ~5s.

The "flakiness" was a **race**: the agent forks into the background (mk-guest.nix) and answers `Ping` during the ~5s window before reboot — so the demo "passed" only when the host pinged in first.

## The fix

Drop the `/etc/mvm/entrypoint` entry from `mkFunctionService.nix` `extraFiles`. The idle bootScript stands as PID 1 → the VM stays up → the agent answers `Ping` reliably. The runner stays at `/usr/lib/mvm/wrappers/runner` for the agent to exec per call.

Verified by code analysis: `funcBootScript` = `buildPreStart` (mkdir+ln, safe) → `/etc/mvm/hooks/before_start.sh` (baked 0755, `set -eu; export FOO=1`, exits 0) → `exec sleep infinity`.

## Validation status

Live end-to-end validation (`dev up → compile → up → agent_ping at t=20s`) was **blocked by a concurrent session holding the shared builder-store flock** (`nix-store-aarch64.img` already attached) — environmental, not the fix. The Plan 120 owner can do the final live boot→ping confirmation on a quiet host.

## Follow-up (separate, for `mvmctl invoke`)

`specs/research/function-runner-nix-artifact-followup.md` — the runner is still inlined verbatim, so its shebang/mode break actual function invocation. Make it a per-language nix-built artifact (Python `writePython3`, Node `makeWrapper`), install via `source=` (preserves nix's 555 mode), and relax the agent's `required_mode` `0o755`→`0o555`. This PR fixes boot→ping (Task 4); the follow-up fixes `invoke`.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,89 @@
+name: Publish SDK to npm
+
+# Publishes the TypeScript SDK (sdks/typescript, package `mvm-sdk`) to npm.
+#
+# Trigger:
+#   - automatically when a GitHub Release is published (same event as
+#     publish-crates.yml / publish-pypi.yml, so all three ship at one
+#     version — see the version-vs-tag guard below);
+#   - manually via workflow_dispatch (dry_run=true does `npm publish
+#     --dry-run`, no upload — rehearse before the first real publish).
+#
+# Auth: NPM_TOKEN repo secret (a granular automation token with publish
+# rights on `mvm-sdk`). Provenance is attested via OIDC. See
+# specs/runbooks/publish-sdks.md.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "npm publish --dry-run only; no upload"
+        required: true
+        default: true
+        type: boolean
+
+concurrency:
+  group: publish-npm
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # npm provenance via OIDC
+      contents: read
+    defaults:
+      run:
+        working-directory: sdks/typescript
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Check token
+        if: ${{ github.event_name == 'release' || !inputs.dry_run }}
+        run: |
+          if [ -z "${NODE_AUTH_TOKEN:-}" ]; then
+            echo "::error::Missing NPM_TOKEN secret."
+            exit 1
+          fi
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Version coupling: refuse to publish an SDK whose version doesn't
+      # match the release tag (the SDK emits IR the same-version toolchain
+      # consumes). GITHUB_REF_NAME is read as a shell env var, never
+      # interpolated into the script — no workflow injection.
+      - name: Assert SDK version matches the release tag
+        if: github.event_name == 'release'
+        run: |
+          set -euo pipefail
+          sdk_ver=$(node -p "require('./package.json').version")
+          tag="${GITHUB_REF_NAME#v}"
+          echo "sdk=$sdk_ver tag=$tag"
+          if [ "$sdk_ver" != "$tag" ]; then
+            echo "::error::sdks/typescript/package.json version ($sdk_ver) != release tag ($tag). Bump it in the release commit."
+            exit 1
+          fi
+
+      - name: Install + build
+        run: |
+          npm ci
+          npm run build
+
+      - name: Publish to npm (dry-run)
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run }}
+        run: npm publish --provenance --access public --dry-run
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish to npm
+        if: ${{ github.event_name == 'release' || !inputs.dry_run }}
+        run: npm publish --provenance --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,73 @@
+name: Publish SDK to PyPI
+
+# Publishes the Python SDK (sdks/python, package `mvm`) to PyPI.
+#
+# Trigger:
+#   - automatically when a GitHub Release is published (same event that
+#     drives publish-crates.yml, so the SDK ships at the same version as
+#     the toolchain — see the version-vs-tag guard below);
+#   - manually via workflow_dispatch (dry_run=true builds + checks only,
+#     no upload — use this to rehearse before the first real publish).
+#
+# Auth: PyPI Trusted Publishing (OIDC) — no long-lived token. Configure a
+# trusted publisher on PyPI first (see specs/runbooks/publish-sdks.md):
+#   project `mvm` → Publishing → add GitHub publisher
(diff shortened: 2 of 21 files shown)


auser:
## Update — boot→ping now validated live ✅

Tracing the first fix's live boot (`MVM_KRUN_LOG=trace`) revealed it was **necessary but not sufficient**: removing the runner-as-entrypoint override correctly made the idle `funcBootScript` PID 1, which then exposed a **second** bug one layer down —

```
/nix/store/…-greet-prestart: line 3: dirname: command not found
/nix/store/…-greet-prestart: line 3: mkdir:   command not found
mvm: /etc/mvm/entrypoint returned without exec — kernel will panic
```

`buildPreStart` (inlined into the bootScript) staged the user source with **bare** `mkdir`/`dirname`/`ln`; the busybox workload rootfs has no coreutils on the early-boot PATH, so `set -eu` aborted PID 1 before `exec sleep infinity` → the same ~5s reboot. Fixed to absolute `${pkgs.coreutils}/bin/…` paths (commit d742131d), matching the pattern the next bootScript line already used for `sleep`.

**Live result (libkrun, isolated rebuild so the binary is unambiguously this branch):**
- New workload revision `2zwlrwn4…` (not the broken `lw7f7bmf7…`)
- Trace shows **no** `command not found` / `returned without exec`
- `REBOOTED=no` — VM stays up past the old 5s window
- `PING_AT_25s=OK` — `AGENT PING OK … capabilities: [Ping]` at t=25s

Note: `up`'s exit code alone is **not** a reliable signal — the agent forks in the background and answers Ping inside the 5s window, so `UP=0` can race-pass even when the VM reboots. The `t=25s` ping is the trustworthy check.

Boot→ping (Plan 120 Task 4) is green. The runner-as-nix-artifact follow-up (`specs/research/function-runner-nix-artifact-followup.md`) remains for `mvmctl invoke` (`RunEntrypoint` still returns `EntrypointInvalid` — the agent's "entrypoint marker contents not absolute" + runner shebang), which is separate from boot→ping.

auser:
## #2 follow-up: function `invoke` — marker decouple + mode, validated live

Picked up the `mvmctl invoke` (vsock `RunEntrypoint`) follow-up. Tracing the agent boot log peeled two layers the original note didn't capture:

**1. Primary blocker — `/etc/mvm/entrypoint` overloaded (marker vs boot command).** The agent's `EntrypointPolicy` read `/etc/mvm/entrypoint` as a marker whose contents must be a *bare absolute path* to the runner; but that file is also PID 1's idle boot command (a `#!/bin/sh` script `/init` sources). The two requirements are mutually exclusive on one file — the old override satisfied the agent by breaking PID 1; the boot fix satisfies PID 1 by breaking the agent. Fixed by giving the agent its **own** marker, `/etc/mvm/runner` (`mkFunctionService` writes it; `EntrypointPolicy.marker_path` points at it). `/etc/mvm/entrypoint` stays the boot command.

**2. Next blocker (revealed once #1 cleared) — runner mode.** Trace then showed `runner has mode 555 (must be 755)`. mkGuest's rootfs hardening pass strips owner-write from baked files (`install -m 0755` doesn't survive), and the agent binary itself is `0555` — so `0555` is the correct expected mode and the `0o755` requirement was the bug. Set `required_mode = 0o555` + declare the runner at `0555`.

Plus the python runner shebang: `#!/usr/bin/env python3` → `#!${pkgs.python3}/bin/python3` (no `/usr/bin/env` in the busybox rootfs; `oneshot.py` is stdlib-only on the JSON path).

**Live result** (rev `47l5zg67…`, isolated agent rebuild from `/work`):
```
mvm-guest-agent: entrypoint validated at /usr/lib/mvm/wrappers/runner (held open for fexecve)
```
Both prior errors gone; boot still green (`REBOOTED=no`, `PING_AT_20s=OK`). The agent now accepts the runner marker and is ready to serve `RunEntrypoint`.

**Remaining (not yet validated live):** an actual `RunEntrypoint` returning `hello ari` — exercises the runner's `python3` shebang *exec* + `greet` dispatch. That needs the `template create/build/invoke` lifecycle (invoke boots from a registered template, not the compiled flake). Node ESM needs a shell shim + the `runnerFiles` registry generalization. Both are specced in `specs/research/function-runner-nix-artifact-followup.md`.

The CI `factory_shape.nix` assertion and code comments were updated to the new marker name; historical ADR/plan/guide references are flagged for a coordinated reconciliation pass.

auser:
## ✅ `mvmctl invoke` works end-to-end — `hello ari`

```
$ agent_invoke <vm> ari
STDOUT: "hello ari"
TERMINAL: Exit { code: 0 }
```

Function invocation never ran E2E before. Brought it up layer by layer (each found by driving a real `RunEntrypoint` against an `up --flake` VM via the `agent_invoke` example, then reading the failure):

1. **Marker decouple** — agent reads a dedicated `/etc/mvm/runner` (not `/etc/mvm/entrypoint`, which is PID 1's boot command).
2. **Mode 0555** — the rootfs hardening pass strips owner-write; the agent's `required_mode` was wrong at 0o755.
3. **Shebang** — `#!/usr/bin/env python3` → `#!${pkgs.python3}` (no `/usr/bin/env` in the busybox rootfs).
4. **fexecve vs shebang script** — the agent execs the runner via a CLOEXEC `/proc/self/fd/<n>`; a shebang re-exec can't reopen it. Dropped CLOEXEC (`F_DUPFD`); TOCTOU preserved (same pinned inode).
5. **`import mvm` (the decision point)** — the user module did `import mvm` for the `@mvm.app` decorator, but the decorator is **build-time metadata already lowered to IR**, and returns the function unchanged. So rather than shipping the SDK to the guest, `compile` now **strips** `import mvm` + `@mvm.*` from the bundled source (tree-sitter pass, `strip_framework.rs`). The guest imports a bare `def greet` — the SDK never enters the rootfs.

Stripped bundle (live):
```python
def greet(name: str) -> str:
    return f"hello {name}"
```

### Follow-ons (tracked, not in this PR)
- **Default per-app-type hooks** (`explicit > registry default > no-op`): Python `before_build` default = `pip install -r requirements.txt` in the builder, for *real* user deps. Needs the `before_build`-in-builder consumer wired (stubbed today). hello-app's `requirements.txt` documents the convention.
- **Publish the SDKs** — `mvm` (PyPI) + `mvm-sdk` (npm) are packaged; need CI workflows + a release trigger.
- **Node** invoke: strip pass + runner shim are analogous (tree-sitter-typescript), not yet wired.

auser:
Superseded by #538. All of this PR's validated work — the function-workload boot fix, the invoke E2E fixes (runner mode 0555, CLOEXEC drop, python shebang, @mvm.app decorator strip → `hello ari`), the PyPI/npm publish workflows (npm scoped to @runmvm/mvm), and the agent_invoke harness — was reconciled onto the feat/artifact-model line (adapted to its /etc/mvm/boot design, dropping the redundant marker-rename) and folded into #538. Also fixed a latent reboot in #538's bootScript (bare `mkdir`) that this reconciliation surfaced. Stage 0 nix-store persistence stays as its own follow-up.

Issue #549, opened 2026-06-03
Vz builder VM fails to launch: VZErrorCode 2 "storage device attachment is invalid"


auser:
## Summary

`mvmctl dev up` / builds on the **Vz** builder backend (`MVM_BUILDER_BACKEND=vz`,
the auto-detect default on macOS 26+ Apple Silicon) fail to launch the **builder
VM** with:

```
mvm-vz-supervisor: VM failed to start: Error Domain=VZErrorDomain Code=2
  "The storage device attachment is invalid."
  UserInfo={NSLocalizedFailure=Invalid virtual machine configuration.,
            NSLocalizedFailureReason=The storage device attachment is invalid.}
Error: vz builder VM: nix build failed inside builder sandbox:
  supervisor exited with non-zero status (3)
```

This is raised at host-side Vz config `validate()` (VZErrorCode 2 =
`.invalidVirtualMachineConfiguration`) — **before the guest kernel boots** (the
builder VM `console.log` is empty). The Stage 0 `nix build` of the builder-VM
flake completes fine; only the subsequent Vz builder-VM launch fails.

## Environment

- macOS 26+ Apple Silicon (arm64), Vz / Apple Virtualization.framework
- `mvm-vz-supervisor` built + codesigned via `crates/mvm-vz-supervisor/tools/build.sh`
  (carries `com.apple.security.virtualization`); path pinned via `MVM_VZ_SUPERVISOR_PATH`

## What it is NOT

- **Not disk-size alignment.** All attached images are 512-byte aligned:
  `nix-store-aarch64.img` / `nix-store-stage0-aarch64.img` (`mod 512 == 0`),
  `rootfs.ext4` (`mod 512 == 0`), `vmlinux` (31 MB, `mod 512 == 0`).
- **Not branch-specific.** Reproduces identically on `main` and
  `feat/artifact-model` (no delta in `crates/mvm-vz-supervisor` or `crates/mvm-backend`
  between them).
- **Not kernel-related.** Independent of guest kernel content — it's a host-side
  disk-attachment config rejection.

## Where to look

`crates/mvm-vz-supervisor/Sources/mvm-vz-supervisor/Supervisor.swift` builds
`VZDiskImageStorageDeviceAttachment(url:, readOnly:)` per `config.disks` and the
failure surfaces at `vm.start()` validation. Candidate causes: a disk URL that
doesn't exist yet at validate time, the same image attached twice, an exclusive-lock
[...]

PR #550, opened 2026-06-03
chore(builder-vm-kernel): drop unused dm-verity/conntrack/ext2 from enables

auser:
## What

Slim the builder-VM kernel config (`nix/images/builder-vm/kernel/default.nix`)
by moving symbols nothing in the builder-VM runtime uses out of `enables` and
into `disables`. This extends Plan 95's empirical method (previously applied
only to the SoC `disables` clusters) to the `enables` list.

Removed (audited against `crates/mvm-host-vm-init`):

- **`MD` / `BLK_DEV_DM` / `DM_VERITY`** — the builder VM is a dev-tier image that
  boots `root=/dev/vda ro` with no roothash and never opens a dm-verity device.
  Verified boot is a *workload-kernel* concern; the builder only
  `veritysetup format`s sidecars (userspace Merkle hashing, no kernel `dm` target).
- **`NF_CONNTRACK` / `NF_DEFRAG_IPV4` / `NETFILTER_XT_MATCH_STATE` /
  `NETFILTER_XT_MATCH_CONNTRACK` / `NETFILTER_XT_MARK`** — the egress lockdown
  (`network.rs` `install_egress_lockdown`) is an OUTPUT owner-match + default-DROP
  ruleset with no conntrack/state/mark match. Kept `NETFILTER_XT_MATCH_OWNER`.
- **`EXT4_USE_FOR_EXT2`** — nothing mounts ext2/ext3; the only ext4 mount is `/dev/vdb`.

Symbols are moved to `disables` (not merely dropped from `enables`) so
`olddefconfig` actually turns them off instead of leaving a defconfig default `=y`.

## Why these are safe / kept

`VIRTIO_CONSOLE` (backs `console=hvc0`), `HW_RANDOM_VIRTIO` (early entropy/TLS),
`SECCOMP` (nix `filter-syscalls` + Firecracker), and all virtio/fs/namespace/cgroup
symbols are load-bearing and untouched. The change is orthogonal to KVM/virtio/vsock,
so it cannot regress microVM generation.

## Verification

- ✅ **Builds cleanly.** The builder-VM `nix build` produces a valid slim
  Linux 6.12.87 arm64 `Image` (`buildPhase completed`, `olddefconfig` accepted the
  moved `disables`), verified locally across multiple runs via the Vz/libkrun
  Stage-0 build path.
- ⏳ **Boot deferred to CI (Linux/Firecracker).** The slim builder-VM kernel is only
  booted by Firecracker (Linux) or Vz; on macOS, libkrun boots the libkrunfw-bundled
  kernel (the built `vmlinux` is unused there), and the Vz builder path is currently
  blocked by an unrelated bug — see #549. CI's Linux/Firecracker lane exercises the
  real boot + the egress lockdown where it matters at runtime.

## Related

- #549 — Vz builder VM "storage device attachment invalid" (separate, pre-existing;
  blocks local Vz boot verification on macOS).

### DIFF
--- a/nix/images/builder-vm/kernel/default.nix
+++ b/nix/images/builder-vm/kernel/default.nix
@@ -28,6 +28,18 @@
 #   would otherwise force `=m`). Everything we need is `=y`. The
 #   rootfs ships no `/lib/modules` tree; mkGuest's
 #   `copy_module_closure` is unused for the builder-VM rootfs.
+#
+# Why no dm-verity / netfilter conntrack:
+#   This is the *builder VM* kernel — a dev-tier image (ADR-002
+#   dev-vs-prod tiers). It boots `root=/dev/vda ro` with no roothash
+#   and never opens a dm-verity device; verified boot is a
+#   workload-kernel concern, and the `veritysetup` the builder ships
+#   only runs `format` (userspace Merkle hashing, no kernel `dm`
+#   target). The egress lockdown (mvm-host-vm-init `network.rs`) is an
+#   OUTPUT owner-match + default-DROP ruleset with no conntrack/state/
+#   mark match. So MD/BLK_DEV_DM/DM_VERITY and the netfilter conntrack
+#   symbols live in `disables`, not `enables`. Audited against
+#   crates/mvm-host-vm-init.
 
 { pkgs }:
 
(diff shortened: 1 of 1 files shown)


auser:
Superseded by #554, which folds in this exact slimming (builder kernel config is directive-identical) plus the shared kernel base, the `mvmctl kernel build` command, and the dual-arch publish workflow. Closing to avoid a conflict on the kernel config.

PR #574, opened 2026-06-04
Plan 158 — dual dev/prod default microVM image (design refinement)

auser:
Refines Plan 158 (merged in #570) to the **dual-image** design we settled on, and is the contract the flake + resolver implementation will follow.

## The model

The `nix/images/default-tenant` flake produces **two** variants, keyed on the existing `mvm_build::pipeline::BuildMode` (`--dev`/`--prod`, default prod):

| Variant | mkGuest | Tier | Verity | exec/console | Ships? |
|---|---|---|---|---|---|
| **dev** | `entrypoint.shell` | dev | no | **yes** | no — built locally in dev mode |
| **prod** | `entrypoint.command` | prod | **yes** | no | **yes** — the `default-microvm-*` release assets |

A single image can't be both interactive and sealed, so dev = the scratch image you `exec` into; prod = the hardened image that ships for hosting.

## Security posture (verified, not asserted)

The microVM **isolation boundary holds for both** variants (hypervisor, seccomp-jailed VMM, no host-fs beyond shares, default-deny egress) — the host is protected either way. The **dev** variant relaxes only *in-guest* defense-in-depth (entrypoint uid 0 via `defaultEntrypointUid`, `do_exec` via `withDevShell`, no rootfs verity) — exactly the dev tier ADR-002 explicitly exempts from verified boot. Untrusted/prod workloads must run in the prod variant or a sealed user flake.

## Honest status of the implementation

This PR is the **design/plan only**. The flake itself (Task 1) genuinely needs a **nix/Linux environment** — I can't generate a valid `flake.lock` or eval/build Nix from a macOS host, so that piece is CI/nix-loop work. The grounded facts to write it are all captured (workload kernel w/ `DM_VERITY`, the runtime-overlay `veritysetup` recipe, `mvm-meta.json` via `builtins.toJSON`). The `BuildMode`-aware resolver (Task 3) + release job (Task 2) are locally validatable and follow once the flake's asset contract is fixed.

### DIFF
--- a/nix/images/default-tenant/flake.lock
+++ b/nix/images/default-tenant/flake.lock
@@ -1,0 +1,65 @@
+{
+  "nodes": {
+    "microvm": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "spectrum": "spectrum"
+      },
+      "locked": {
+        "lastModified": 1778669912,
+        "narHash": "sha256-WT2iimtOBZM/6AcZeBoJU2EgUSaywtlItsEgNkZBda0=",
+        "owner": "microvm-nix",
+        "repo": "microvm.nix",
+        "rev": "a7a7009064cec75d9da652c6723412ce27b9bc44",
+        "type": "github"
+      },
+      "original": {
+        "owner": "microvm-nix",
+        "repo": "microvm.nix",
+        "type": "github"
+      }
+    },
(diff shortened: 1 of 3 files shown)


auser:
Superseded by the full feature PR (plan + flake draft + host wiring on feat/158-host-wiring).

PR #679, opened 2026-06-07
docs(plan-169): host-side lifecycle convergence + single-host density

auser:
## What

Two new specs, grounded in a review of an external single-machine sandbox control plane — an MIT-licensed Go service pairing a container runtime, a reverse proxy, and an embedded SQLite store. Its **isolation/security** model does not transfer (hardened container isolation, default-off auth, open egress — its own docs say so). Its **host-daemon lifecycle** shape does: a boot-time reconciler (converge runtime → persistent store), an idle-timeout **and** host-memory-pressure reaper, and a wake-on-request handler. (External project named obliquely per repo policy.)

The finding: **mvm already owns every primitive** — `VmNameRegistry`, the TTL `supervisor::reaper`, `instance_wake`/`auto_resume`, `cache prune --reap-orphans` — but they're fragmented, and two of the three only run inside mvmd's resident supervisor, never on the plain `mvmctl` path. Reconciliation is manual + lazy, so the recurring stale-host-state failures (libkrun.pid race, Stage 0 stale-crate bail, degraded-builder-store `dev up` loop) surface at use-time instead of self-healing.

## Files

- **`specs/plans/169-host-lifecycle-convergence-and-density.md`** — WS-A reconcile-at-CLI-entry convergence (+ `mvmctl reconcile`, self-heals the builder-store loop); WS-B activity-driven idle reaper (sleep, not kill); WS-C host-memory-pressure LRU eviction (escalates after the existing balloon lever); WS-D wake-on-request completion. mvm-side mechanism only — fleet policy stays in mvmd (Plan 140 / mvmd Plan 55).
- **`specs/adrs/074-registry-source-of-truth-convergence.md`** — registry wins on conflict (adopting orphans would resurrect a workload outside signed admission); converge at CLI entry, not a resident daemon; fail-open.
- **`specs/plans/140-…md`** — one-line pointer separating the *trigger* (169) from the snapshot *correctness* gaps (140).

## Scope

No code. Both docs **Proposed**. No change to the isolation posture or claims 1–15 — host-side lifecycle bookkeeping only. `xtask check-spec-numbers` clean (169/074 unique).

### DIFF
--- a/specs/adrs/074-registry-source-of-truth-convergence.md
+++ b/specs/adrs/074-registry-source-of-truth-convergence.md
@@ -1,0 +1,94 @@
+---
+title: "ADR-074: VM name-registry is the source of truth; converge at CLI entry, not a resident daemon"
+status: Proposed
+date: 2026-06-07
+related: Plan 169 (host-lifecycle convergence + density); ADR-045 (hermetic VM-lifecycle testing); ADR-044 (audit_emit! macro); Plan 140 (snapshot/restore productionization); Plan 99 (Stage 0 reaper / cache prune)
+---
+
+## Status
+
+Proposed. Records the architectural posture behind Plan 169; no code lands
+with the ADR. Supersedes nothing — it formalizes a model the codebase already
+half-implements (`VmNameRegistry` + `cache prune --reap-orphans` + the TTL
+`reaper`) and makes the resolution rule explicit.
+
+## Context
+
+mvm's local runtime state lives in two places that can drift apart:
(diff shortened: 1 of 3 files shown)


auser:
Superseded by #688, which folds in this plan + ADR-074 (cherry-picked, renumbered **169 → 170** to clear the `check-spec-numbers` collision with the merged agent-RPC transport plan) alongside the Workstream A implementation. Closing in favor of #688.

PR #721, opened 2026-06-08
feat(secrets): Plan 129 #1b — egress-substitution inject leg (handoff + QEMU serve_vsock + guest proxy)

auser:
Wires the **inject/bearer leg** of Plan 129 egress secret substitution (ADR-067) end-to-end on the QEMU backend: a secret-bearing workload's egress is routed through a guest-local forward proxy to a host substitution endpoint that swaps an opaque placeholder for the real credential — the guest only ever holds `mvm-secret-<hex>`.

### What's wired (4 commits)
- **`mvm-core::substitution_handoff`** — per-VM sidecar (`<vm_state_dir>/substitution.json`, mode 0600) carrying the ready-to-inject workload env (placeholders + `HTTP_PROXY`), parallel to `runtime_meta`'s `mode.json`. **`invoke`** injects it into the `RunEntrypoint` env (was `Vec::new()`); absent sidecar → today's no-env behavior.
- **`prepare_substitution` (mvm-hostd)** — boot-time producer: from a plan's `SecretBinding`s → `from_plan` (mints the service) → builds the env → writes the sidecar. `Ok(None)` when the plan binds no secrets.
- **QEMU bridge** (`__qemu-vsock-bridge`) — reads the admitted plan stashed at `<vm_state_dir>/plan.json`; if it binds secrets, `prepare_substitution` + `serve_vsock` on `SUBSTITUTION_PORT` on a background runtime thread alongside the agent bridge. Self-gating; `--name` added so the bridge can locate the plan + sidecar. Plus `SubstitutionService::serve_vsock_port`.
- **Guest agent** — starts `start_forward_proxy` at boot (unconditional bind, dial-on-demand; only used when `HTTP_PROXY` is set).

### Validation
- Unit tests for every new unit (handoff roundtrip/0600/gating, `prepare_substitution`, invoke consumer) — green on macOS **and** the Linux x86_64 box.
- `cargo fmt --all` + `clippy -D warnings` clean (incl. the cfg(linux) paths, on the box).
- **Live QEMU boot on a KVM box:** a real `--hypervisor qemu --builder qemu` boot confirms the full chain — plan admitted (`backend=qemu`) → `plan.json` stashed (0600) → bridge spawned with `--name` → secret-gating verified (`substitution.json` correctly absent for a no-secret workload).

### Deferred follow-up (separate workstream)
The **live positive substitution demo** is gated on the **Python/TS SDK toolchain** item, not this leg: (1) `mvm.secret()` doesn't yet expose `auth_type`/`allowed_hosts`, so a compiled secret app emits an unbound `SecretRef` that validation rejects; (2) the forward proxy parses absolute-form HTTP, so the workload needs the SDK egress client. Tracked for the next slice. The other backends' bin glue (libkrun/vz/firecracker) follow the same shape as the QEMU bridge here.

### DIFF
--- a/crates/mvm-backend/src/qemu.rs
+++ b/crates/mvm-backend/src/qemu.rs
@@ -590,6 +590,8 @@ fn spawn_vsock_bridge(name: &str, cid: u32, state_dir: &Path) -> Result<()> {
         .arg(cid.to_string())
         .arg("--port")
         .arg(mvm_guest::vsock::GUEST_AGENT_PORT.to_string())
+        .arg("--name")
+        .arg(name)
         .arg("--watch-pid-file")
         .arg(state_dir.join(QEMU_PID_FILE))
         .stdin(std::process::Stdio::null())
--- a/crates/mvm-cli/src/commands/qemu_bridge.rs
+++ b/crates/mvm-cli/src/commands/qemu_bridge.rs
@@ -25,11 +25,103 @@ pub(in crate::commands) struct Args {
     /// Guest vsock port to dial.
     #[arg(long)]
     port: u32,
+    /// VM name — used to locate the stashed admitted plan and write the
+    /// per-VM substitution handoff sidecar (Plan 129).
+    #[arg(long)]
+    name: String,
     /// Exit when this PID file's process is no longer alive (the VM is gone).
     #[arg(long)]
     watch_pid_file: PathBuf,
 }
 
 pub(in crate::commands) fn run(args: &Args) -> Result<()> {
+    // Plan 129 #1b: if this VM's admitted plan binds secrets, serve the egress
+    // substitution endpoint over AF_VSOCK alongside the agent bridge. The guest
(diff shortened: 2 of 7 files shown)


auser:
Closing as superseded. The Plan 129 inject leg landed on `main` in parallel via #713/#715/#717/#718 with a stronger architecture — a separate jailed `mvm-substitution-endpoint` subprocess moat (vs this branch's in-bridge `serve_vsock`). Every piece here is already on main in a different shape (endpoint spawn, `invoke.rs::substitution_env` HTTP_PROXY/placeholder injection, guest forward-proxy start), so merging would duplicate it. The box run on this branch independently validated the QEMU boot + admission + plan-stash + secret-gating pipeline (captured in memory for whoever boot-validates the merged moat). Remaining Plan 129 work — the SDK toolchain (`mvm.secret()` `auth_type`/`allowed_hosts` + the absolute-form egress client) — is unaffected and still open.

PR #700, opened 2026-06-08
feat(plan-152 WS-B): Rust-native vz supervisor — boot + vsock proxy + gvproxy network (slices 1-3)

auser:
Plan 152 WS-B (Stage 3) — the Swift→Rust port of the per-VM Vz supervisor, as an
additive, behind-the-parity-gate slice series. **Production stays on the Swift
`crates/mvm-vz-supervisor` until the parity matrix is green and the bin is
codesigned.** The new `mvm-vz-supervisor` `[[bin]]` is reachable for testing via
`MVM_VZ_SUPERVISOR_PATH` (resolver priority 1) — no `vz.rs` resolver change.

Threading per the recorded WS-B decision (#697 / ADR-056 addendum): a private
serial dispatch queue + `VZVirtualMachineDelegate` (objc2 `define_class!`),
bridged to a tokio current-thread runtime via a oneshot-per-hop
`SerialQueue::dispatch` — not a main-thread `CFRunLoop`. All `unsafe` lives in
`mvm-vm-host/src/vz_objc.rs`. The objc2 stack (objc2 / objc2-virtualization /
block2 / dispatch2) is already in the workspace lock via mvm-backend's
`apple_container` path — no new dependency.

### Slices in this PR
- **1 — boot + lifecycle** (`81a0e015`): config translation (cpu/mem/kernel/
  disks/virtio-fs/vsock-device/console/entropy/balloon/platform), cold boot,
  SIGTERM/SIGINT → graceful `requestStop`, guest exit-code propagation (0 clean
  / 1 error) via a terminal-state watch.
- **2 — vsock host proxy** (`e833a25b`): per-port `<socket_dir>/vsock-<port>.sock`
  (0700) UNIX listeners → `connectToPort` on the VM queue → `copy_bidirectional`
  splice (`VsockStream` = dup'd guest fd in `AsyncFd`). Makes the guest agent
  reachable for mvmctl/mvmd agent RPC.
- **3 — direct gvproxy network** (`c15a94ec`): SOCK_DGRAM vfkit socket →
  `VZFileHandleNetworkDeviceAttachment` with pinned MAC. Unblocks a real
  networked guest boot.

### Fail-closed (still on Swift behind the gate)
- `Restore` startup mode.
- **Flow-audited** networking (`events_ingest` set) — refused, not direct-attached,
  so the claim-10 egress audit is never silently bypassed. Porting it as an
  in-process `payload_tap` (absorbing Plan 141's Vz arm) is the next slice.

[...]

PR #730, opened 2026-06-08
test(plan-152): gate — control-verb + save/restore as absolute Rust-correctness

auser:
Follow-up to #703 (the merge stranded this last commit). #703 landed the gate
with the harness fixes (writable rootfs copy + STATUS-based boot detection) that
make it runnable live — but its control-verb and save/restore checks still assert
`== Swift`, which is the wrong bar.

## Why `== Swift` is wrong for these
Running the gate live against the real Swift + Rust (slice-6) supervisors on
macOS 26 showed the **Swift** supervisor *deadlocks* on PAUSE/RESUME/SAVE — its
`synchronousVZCall` blocks the VM's dispatch queue waiting for a completion that
fires on that same queue, wedging the control socket. Swift is the broken,
soon-to-be-deleted baseline, and the Rust port *fixes* the deadlock (the Stage-2
async-dispatch decision). So "matches Swift" can never go green there.

## This change
- **`control_verbs_rust_correct`** — asserts the Rust control socket answers
  STATUS/PAUSE/RESUME correctly (absolute, no Swift). **Passes live.**
- **`save_restore_rust_correct`** — asserts Rust SAVE returns OK + writes the
  snapshot blob and `.machine-id` sidecar, then restores to running. **Expected
  red at slice 6**: the SAVE handler calls `saveMachineStateToURL` while the VM
  is *running*, but VZ requires it paused first (`VZErrorDomain:3`). The fix is
  the `pause → save` order from the ADR-056 addendum / WS-B plan; this test is
  the live "save/restore not safe yet → don't delete Swift" signal and greens
  when that lands.

Boot (P1) and vsock (P2) stay parity-vs-Swift — Swift works there. No conflict
with the Swift→Rust rewrite: this is a test file in `mvm-build`, separate from
the `mvm-vm-host` supervisor; it verifies the replacement.

Verified live on macOS 26: P3 ok, P4 red with the VZErrorDomain:3 message.
Offline: `cargo test -p mvm-build --test vz_supervisor_parity` 9 pass (live
skips), nightly fmt, clippy clean.

### DIFF
--- a/crates/mvm-build/tests/vz_supervisor_parity.rs
+++ b/crates/mvm-build/tests/vz_supervisor_parity.rs
@@ -32,14 +32,20 @@
 //! Both supervisor binaries share the name `mvm-vz-supervisor`, so they are
 //! addressed by explicit path, never by name resolution.
 //!
-//! Slices: P1 boot + graceful-stop + exit-code parity; P2 vsock proxy round-trip
-//! parity (identical request bytes → identical reply bytes through each
-//! supervisor's per-port proxy); P3 control-verb parity (STATUS/PAUSE/RESUME
-//! replies match step-for-step); P4 save/restore parity (`SAVE` a snapshot, then
-//! restore it in a fresh supervisor). P3/P4 are staged ahead of the Rust control
-//! socket + snapshot slices, so they report the gap until those land. Plus the
-//! pure config/contract/round-trip/codec helpers. (BALLOON joins P3 once the
-//! boot config carries a balloon device.)
+//! Two kinds of live check, because the Swift baseline is uneven on macOS 26:
+//!
+//! - **Parity** (P1 boot, P2 vsock round-trip) — compares Rust against Swift on
+//!   surfaces where Swift works, a no-regression check during the migration.
+//! - **Absolute Rust-correctness** (P3 control verbs, P4 save/restore) — Swift
+//!   *deadlocks* on PAUSE/RESUME/SAVE here (`synchronousVZCall` blocks the VM's
+//!   dispatch queue), so it is a broken, soon-to-be-deleted baseline; "matches
+//!   Swift" is the wrong bar. These assert the Rust supervisor is correct on its
+//!   own terms. P4 is expected to fail until the slice-6 SAVE handler adds the
+//!   `pause → save` step (ADR-056 addendum) — it's the live "not safe to delete
+//!   Swift yet" signal.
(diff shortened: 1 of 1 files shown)


auser:
Reached the identical conclusion independently while driving #700 to merge (validated live on macOS 26 / Apple Silicon against both supervisors). Confirming your Swift-deadlock root cause, and adding three findings from the live run that this PR will want:

**1. `save_restore_rust_correct` is red for *two* reasons, and neither is a Rust defect — Rust is save/restore-correct.** With a caller-side `PAUSE` before `SAVE`, Rust returns `OK` and writes the snapshot + `.machine-id` sidecar; and a fresh supervisor restores it **to running**. I verified the test goes fully green that way (9/9 live). So the "red signal" is purely the missing pause + a harness bug (below), not "save/restore not safe yet" on the Rust side.

**2. Independent harness bug — restore phase ENOENT.** `probe_save_restore` never `create_dir_all`s `<root>/restore`, so even once SAVE works the restore supervisor aborts before VZ with:
```
Error: boot vz guest
  create console log .../restore/console.log: No such file or directory (os error 2)
```
That alone was the entire cause of `restore_reached_running=false` in my first runs. You'll hit it regardless of where the pause lives — `std::fs::create_dir_all(&restore_dir)` before `build_boot_config` fixes it.

**3. The dev `/init` self-exits on console EOF**, so control/save verbs can race a powered-off guest. An `MVM_VZ_PARITY_CMDLINE` override (defaulting to the current cmdline) lets the runner point at a long-lived PID 1 (e.g. `... init=/bin/busybox sleep 1000000`); with that, the sequence is deterministic.

**On the design question (pause in supervisor vs caller):** the slice-6 live validation and the real backend path do caller-side `PAUSE`→`SAVE` (the supervisor's SAVE saves the already-paused state), so a harness `PAUSE` mirrors real usage. But that's your call as the WS-B owner — flagging the evidence either way. Happy to hand over the `create_dir_all` + cmdline-env diffs if useful.

auser:
Superseded by #736 (Plan 152 WS-B finalization). #736 deletes the Swift supervisor entirely and makes this gate Rust-only (asserting Rust correctness), which folds in this PR's intent — there's no Swift left to compare against. Carries the same control-verb / save-restore Rust-correctness pivot plus the pause-before-save + restore-dir mkdir fixes, validated live on macOS 26.

PR #731, opened 2026-06-08
test(plan-152): harden vz parity gate — validate Rust, fix save/restore probe

auser:
## Summary

Hardens the Plan 152 WS-B parity gate (`crates/mvm-build/tests/vz_supervisor_parity.rs`, merged in #703) after running it **live on macOS 26 / Apple Silicon** against both supervisors. The live run found that the gate's `rust == swift` byte-equality contract is unachievable — **the Swift control socket self-deadlocks** — and surfaced two harness bugs.

### Why byte-equality can't hold
The Swift control socket runs on the VM's *own serial `DispatchQueue`* and `synchronousVZCall` does `sema.wait()` on that queue while `vm.pause/resume/save`'s completion handler is dispatched back to the same queue → **self-deadlock**. `STATUS` works (pure read); the first async verb wedges the queue, so every later connection times out. The Rust supervisor's serial-queue→tokio bridge avoids this — that's the bug WS-B exists to fix. Asserting equality with a broken baseline would block the *correct* implementation, so the live control-verb and save/restore tests now **assert Rust correctness**, with Swift probed informationally (logged, not failed) since it's being retired.

### Harness bugs fixed (both surfaced by the live run)
- `probe_save_restore` now sends `PAUSE` before `SAVE` — VZ rejects `saveMachineStateToURL` on a running VM (`VZErrorDomain:3`).
- `probe_save_restore` now `create_dir_all`s the restore state dir — otherwise the supervisor's console-log open fails with `ENOENT` before it reaches VZ (this was the sole cause of the earlier "restore didn't reach running").

### Also
- New `MVM_VZ_PARITY_CMDLINE` env to supply a long-lived guest cmdline (the dev `/init` self-exits on console EOF, too short for the control/save sequence).
- Renamed the two pivoted tests: `control_verbs_rust_correct`, `save_restore_rust_correct`.
- Refresh `specs/REFACTOR-STATUS.md`: Plan 152 WS-B (rewrite #700, gate #703, this hardening) + WS-E folded into #700; Plan 159 WS-1 now unblocked; doctor follow-up #728.

## Test Plan
- [x] Live on macOS 26 / Apple Silicon (both supervisors built + signed): **9/9 pass** — `boot_parity`, `control_verbs_rust_correct`, `save_restore_rust_correct` (snapshot+sidecar+restore-running), config/contract/proxy tests. Rust validated on boot/control/save/restore.
- [x] `cargo clippy -p mvm-build --tests -- -D warnings` clean; nightly `cargo fmt` clean.
- [x] Live tests skip cleanly without the env vars (CI Linux / plain Mac), as before.

## Follow-ups (tracked, not in this PR)
- Flip `resolve_supervisor_path` to the Rust bin + delete the Swift crate (then the Swift-informational probes here are removed and the gate is Rust-only).
- Security-review items from #700: stale SAVE doc, `tenant_id`-without-`plan` fail-open (mirrors libkrun), exit-listener leak, a Vz-bridge deny-path CI test.

### DIFF
--- a/crates/mvm-build/tests/vz_supervisor_parity.rs
+++ b/crates/mvm-build/tests/vz_supervisor_parity.rs
@@ -68,6 +68,20 @@ const VSOCK_REQUEST: &str = "MVM_VZ_PARITY_VSOCK_REQUEST";
 /// real launch path does.
 const DEFAULT_CMDLINE: &str = "console=hvc0 root=/dev/vda rw init=/init";
 
+/// Override for [`DEFAULT_CMDLINE`]. The committed default boots the dev `/init`,
+/// whose shell self-exits on console EOF — fine for boot parity, but too
+/// short-lived for the control-verb / save-restore gates (the guest can power
+/// off mid-test). Export this to a long-lived PID 1 (e.g.
+/// `console=hvc0 root=/dev/vda rw init=/bin/busybox sleep 1000000`) so the guest
+/// stays up while the verbs run.
+const CMDLINE_OVERRIDE: &str = "MVM_VZ_PARITY_CMDLINE";
+
+/// Resolve the boot cmdline: [`CMDLINE_OVERRIDE`] if exported, else
+/// [`DEFAULT_CMDLINE`].
+fn cmdline() -> String {
+    std::env::var(CMDLINE_OVERRIDE).unwrap_or_else(|_| DEFAULT_CMDLINE.to_string())
+}
+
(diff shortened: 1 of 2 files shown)


auser:
Closing as a duplicate of #730 (the WS-B owner's gate-hardening, opened first). I reached the same conclusion independently — Swift deadlocks on async control ops, so the gate must assert Rust correctness, not `== Swift`. Folding my net-new findings into #730 instead. Apologies for the rival PR.

PR #748, opened 2026-06-09
feat(supervisor): prelaunched libkrun supervisor — base/attach split + attach-time plan re-verify (Plan 118 WS-1 1a)

auser:
## Summary

Plan 118 WS-1 **Layer 1a** — the security-critical *supervisor primitive* of the
macOS "instant up" warm pool. Adds a **prelaunched** mode to
`mvm-libkrun-supervisor`: it does the workload-independent setup at spawn
(codesign re-exec, dylib load), then **blocks on a control UDS holding no rootfs
and no plan before `start_enter`**. On a verified `attach` it re-verifies the
signed `ExecutionPlan`, merges the workload config, and only then boots.

Scope is the primitive only — **no pool, no `up` integration, no
`warm_pool_size`** (those are 1b). Design:
`specs/notes/plan-118-ws1-layer1a-prelaunched-supervisor-design.md`;
implementation plan:
`specs/notes/plan-118-ws1-layer1a-implementation-plan.md`.

### The security crux

The cold/legacy path *deliberately skips* plan re-verify — the host admitted the
plan and stdin is a private parent→child pipe, trusted under ADR-002 (extract,
don't verify). The warm path's control UDS is **same-uid-reachable**, so it is
**not** a trusted private channel. The supervisor therefore **independently
re-verifies** before any `start_enter`:

- Ed25519 signature via `mvm_core::plan::verify_plan` against the host-signer
  **public** key derived from the on-disk secret (claim 8; **no new key**),
- the G4 validity window (`check_window`) + per-signer nonce-replay
  (`NonceStore`), mirroring `mvm-hostd::supervisor::aggregate::launch`'s order.

Cross-standby replay is closed by a **per-spawn binding nonce** the attach must
echo; combined with **one-shot attach** (accept exactly one connection, then
boot-or-die) and the plan's G4 window. The verify+merge is a **pure function**
(`mvm_vm_host::prelaunch::verify_and_merge_attach`) so the rejection ladder is
unit-testable with no VM. The verifiers are reused, not forked.

### Changes

- **`libkrun-sys`**: `SupervisorBaseConfig` (workload-independent) /
  `SupervisorAttachConfig` (workload-specific) split, both
  `deny_unknown_fields`, + `SupervisorConfig::from_base_and_attach` (binding-nonce
  echo + field merge; refuses a base that already carries a rootfs). Leaf crate —
  no `mvm-core` dep, so the plan re-verify lives in `mvm-vm-host`.
- **`mvm-vm-host`**: `prelaunch::verify_and_merge_attach` (the crux) + bin wiring
  — a `prelaunch_base` stdin envelope routes to `run_prelaunched`: bind a `0700`
  control UDS (`0700` dir, binding nonce in path), accept one connection with a
  per-conn timeout, read the length-prefixed attach frame, verify+merge, hand off
  to the existing `run_with_bridge`. The legacy bare-`SupervisorConfig` path is
  byte-for-byte unchanged; its exit-capture + routing is factored into a shared
  `dispatch_config`.
- **`mvm-hostd::framing`**: sync `read/write_json_frame_sync` siblings (the
  prelaunch accept path runs before `start_enter`, no tokio runtime).
- **`mvm-core`**: promote the `sample_plan` fixture to a feature-gated
  `plan::signing::test_support` so the supervisor crate signs valid plans without
  duplicating it.
- **Fuzz**: `fuzz_attach_message` (the only attacker-reachable-post-spawn
  surface). Also fixes the `security.yml` fuzz job's stale
  `crates/libkrun-sys` working-directory (the crate moved to
  `crates/deps/libkrun-sys` in the plan-121 consolidation, so the existing
  `fuzz_supervisor_config` step could not have found its fuzz dir).

### Design note — stdin dispatch

The design doc named a serde-tagged enum `SupervisorStdin { Whole, Base }`. This
PR instead uses a `prelaunch_base`-wrapper-key probe (try the wrapped envelope,
fall back to the bare legacy config). Reason: `deny_unknown_fields` is
incompatible with internally-tagged enums, and an `#[serde(untagged)]` fallback
could *silently* route a botched prelaunch to the permissive legacy arm (booting
without the audit bridge). The wrapper-key keeps the legacy bytes unchanged
**and** fails closed on a malformed prelaunch.

## Test Plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace -- -D warnings` (+ feature-gated bin + `libkrun-live` tests)
- [x] `cargo nextest run --workspace -E 'not package(mvm-backend)'` — 4844 passed
- [x] `cargo test --workspace --doc`
- [x] `cargo build -p mvm-vm-host --bin mvm-libkrun-supervisor --features libkrun-sys`
- [x] `xtask check-core-runtime-free` + `check-spec-numbers` clean
- [x] Unit rejection ladder (a–e): wrong binding nonce, wrong/unsigned key,
      out-of-window, replayed nonce, + happy merge
- [x] `libkrun-live` process integration: real supervisor spawn → control-UDS
      bind → accept → frame read → wrong-nonce refusal → exit 6, no boot
- [ ] Manual: `valid_attach_boots_and_agent_reachable` (`#[ignore]`d live-boot
      harness — run on a libkrun-capable host)

## Out of scope (1b — follow-up PR)

`warm_pool_size`/`--warm-pool-size`, the `~/.mvm/pool/<id>/`
`SupervisorStandbyPool`, `up` claim + cold-boot fallback, base-compat (kernel
match), reaper + `cache prune` TTL, the bench delta.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -270,7 +270,7 @@ jobs:
         run: cargo fuzz run fuzz_authed_path -- -max_total_time="$DURATION"
 
       - name: Fuzz SupervisorConfig (host-side)
-        working-directory: crates/libkrun-sys
+        working-directory: crates/deps/libkrun-sys
         env:
           DURATION: ${{ steps.duration.outputs.secs }}
           RUSTUP_TOOLCHAIN: nightly
@@ -281,6 +281,18 @@ jobs:
         # contract as `fuzz_guest_request` for the guest-side
         # `GuestRequest` parser.
         run: cargo fuzz run fuzz_supervisor_config -- -max_total_time="$DURATION"
+
+      - name: Fuzz SupervisorAttachConfig (prelaunch attach)
+        working-directory: crates/deps/libkrun-sys
+        env:
+          DURATION: ${{ steps.duration.outputs.secs }}
+          RUSTUP_TOOLCHAIN: nightly
+        # Plan 118 WS-1 1a — the prelaunched supervisor reads a
+        # `SupervisorAttachConfig` off a same-uid control UDS. It is the
+        # only attacker-reachable surface after spawn; any panic is a
+        # hard process death before the plan re-verify can fail closed.
+        # Same "never panic on any input" contract as fuzz_supervisor_config.
+        run: cargo fuzz run fuzz_attach_message -- -max_total_time="$DURATION"
 
       - name: Fuzz Vz SupervisorConfig (host-side)
         working-directory: crates/mvm-build
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3389,6 +3389,7 @@ version = "0.16.1"
 dependencies = [
  "anyhow",
  "block2",
+ "chrono",
  "dispatch2",
  "ed25519-dalek",
  "libc",
@@ -3405,6 +3406,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "tempfile",
+ "thiserror 1.0.69",
  "tokio",
  "toml",
  "tracing",
--- a/crates/deps/libkrun-sys/fuzz/Cargo.toml
+++ b/crates/deps/libkrun-sys/fuzz/Cargo.toml
@@ -42,3 +42,13 @@ path = "fuzz_targets/fuzz_supervisor_config.rs"
 test = false
 doc = false
 bench = false
+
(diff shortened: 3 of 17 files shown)


auser:
Superseded by #757 — consolidated the 1a→1b-iii stack into one PR (main requires up-to-date branches, which made the 4-deep stacked merge impractical). Same code, rebased clean onto main.

PR #751, opened 2026-06-09
feat: supervisor standby pool mechanism — backend-agnostic trait seam + libkrun impl (Plan 118 WS-1 1b-i)

auser:
## Summary

Plan 118 WS-1 **Layer 1b-i** — turns the 1a prelaunched-supervisor *primitive* (#748)
into an actual warm-pool **mechanism**, behind a **backend-agnostic `VmBackend` trait
seam** so Firecracker / Vz / cloud-hypervisor implement the same methods later, not
one-offs. **Stacked on #748** (base `feat/plan-118-ws1-layer1a`).

Plan: `specs/plans/176-supervisor-standby-pool-ws1-1b.md` (numbered 176 only for the
unique-prefix gate; it *is* Plan 118 WS-1 1b).

### The trait seam (parallel to the existing snapshot axis)

`mvm-core` gains, on `VmBackend`, opt-in + fail-closed (mirroring `snapshot_capability`/
`warm_start`):

```rust
fn supports_standby_pool(&self) -> bool { false }
fn spawn_standby(&self, spec: &StandbySpec)  -> Result<StandbyHandle, StandbyError>;  // default: Unsupported
fn claim_standby(&self, h: &StandbyHandle, claim: &StandbyClaim) -> Result<VmId, StandbyError>;
```

Backend-agnostic `Standby{Spec,Handle,Claim,Compat,State,Error}` types + `warm_pool_size`
(default 0 = feature off) live next to `VmStartConfig`. The pool is a *different axis*
from snapshot (snapshot = restore a booted VM; standby = pre-pay spawn/codesign/setup
latency), so it's a separate seam — they compose.

### What's in this slice

- **mvm-core:** the trait methods + `Standby*` types + `warm_pool_size`; `mvm_pool_dir()`
  / `pool_standby_dir()` config helpers.
- **mvm-backend:** the backend-agnostic `SupervisorStandbyPool` registry (`~/.mvm/pool/`
  state-dir; record / select-compatible / mark-claimed / remove, with pid liveness) +
  the libkrun `spawn_standby`/`claim_standby` impls (translate to the 1a
  `SupervisorBaseConfig`/`SupervisorAttachConfig`; factor a shared `krun_context_base` so
  the cold + warm paths can't drift).
- **mvm-cli:** the launch glue — kernel-sha256 base-compat key, per-spawn binding nonce,
  `StandbySpec` builder, `claim_or_cold` (fail-**open** to cold boot), `warm_to_target`.
- **libkrun-live test:** `spawn_standby` drives the real supervisor bin to bind its
  control UDS and block before boot — passes on the dev host.

### Two correctness refinements made during execution

1. **Base-compat = kernel sha256 + vcpus + mem** (`StandbyCompat`). libkrun fixes
   resources at spawn (`set_vm_config` runs from the base KrunContext, before attach), so
   a claim must match them or it'd silently misconfigure the VM. "No extra volumes" is the
   launch-path half (1a's attach only threads the rootfs).
2. **Sync framing relocated to `libkrun_sys::framing`.** `mvm-backend` can't depend on
   `mvm-hostd` (cycle), so the sync length-prefixed framing the control channel uses moved
   next to the `Supervisor*Config` wire types it frames — single source, reachable by both
   the `claim_standby` writer and the supervisor-bin reader. `mvm-hostd` keeps its async
   variants.

### Fail-open posture

The pool never breaks a launch: no compatible standby → cold boot; a claim error → log +
cold boot + remove the spent standby (never left idle).

## Re-scope note

`up.rs` has **five** `backend.start` sites across direct-boot/re-exec/wait paths. Wiring
auto-claim into that admission-heavy path — plus the `mvmctl pool` command — both need the
same default-kernel + hypervisor resolution and deserve isolated review, so they move to
**1b-ii** (along with replenish-on-use, the reaper + `cache prune`, and the bench delta).
This slice ships the mechanism + tested orchestration helpers; the helper module is
`#[allow(dead_code)]` until 1b-ii wires it. Recorded in the plan doc.

## Test Plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo nextest run --workspace -E 'not package(mvm-backend)'` — 4855 passed
- [x] `cargo test -p mvm-backend standby` (lib tests; mvm-backend test bin SIGKILLs under
      nextest --list on macOS) — 8 passed
- [x] `cargo test --workspace --doc`
- [x] feature-gated supervisor bin builds; `libkrun-live` spawn-binds-socket test passes
- [ ] Manual: the `#[ignore]`'d full spawn→claim→boot→agent harness on a libkrun host

## Out of scope (1b-ii)

`mvmctl pool warm/status`; `--warm-pool-size` + claim-on-launch + replenish wired into
`up`/`run`; reaper TTL + `cache prune`; `doctor` standby column; bench state-dir fix +
cold-vs-warm delta. **Deferred further:** multi-kernel pool keying; honouring an explicit
`--name`/extra volumes for warm launches (→ SPRINT.md in 1b-ii).

### DIFF
--- a/crates/deps/libkrun-sys/src/framing.rs
+++ b/crates/deps/libkrun-sys/src/framing.rs
@@ -1,0 +1,143 @@
+//! Sync length-prefixed JSON framing for the same-uid supervisor control channels.
+//!
+//! Wire format: a 4-byte big-endian length prefix followed by the JSON body. The cap is
+//! enforced on read **before** allocating the body buffer (a hostile `length_prefix =
+//! u32::MAX` must not OOM the reader).
+//!
+//! Lives here — next to the `SupervisorBaseConfig`/`SupervisorAttachConfig` wire types it
+//! frames — because both ends need it without a dependency cycle: the **writer** is
+//! `mvm-backend`'s libkrun `claim_standby` (Plan 118 WS-1 1b) and the **reader** is the
+//! `mvm-libkrun-supervisor` bin's prelaunch attach path (Plan 118 WS-1 1a). `mvm-backend`
+//! cannot depend on `mvm-hostd` (that crate depends on `mvm-backend`), so this can't live
+//! in `mvm_hostd::framing` (which keeps the *async* tokio variants for its own channels).
+//!
+//! This is the no-auth transport — correct for the same-uid control UDS. The trust
+//! boundary is the supervisor's independent plan re-verify on the attach (1a), not this
+//! frame.
+
+use std::io::{Read, Write};
+
+/// Width of the big-endian length prefix.
+pub const FRAME_LEN_BYTES: usize = 4;
+
+/// Errors from the sync framing layer. Hand-rolled (no `thiserror`) to keep this leaf
+/// crate's dependency surface minimal.
+#[derive(Debug)]
+pub enum FrameError {
+    /// Underlying stream read/write failed.
+    Io(std::io::Error),
+    /// Body failed to (de)serialize.
+    Json(serde_json::Error),
+    /// The length prefix exceeded the caller's cap — rejected before allocating.
+    TooLarge { size: usize, cap: usize },
+    /// Body length did not fit the u32 length prefix.
+    LengthOverflow,
+}
+
+impl std::fmt::Display for FrameError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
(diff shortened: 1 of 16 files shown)


auser:
Superseded by #757 — consolidated the 1a→1b-iii stack into one PR (main requires up-to-date branches, which made the 4-deep stacked merge impractical). Same code, rebased clean onto main.

PR #917, opened 2026-06-15
docs(rollup): Plan 197 2a default-path plan-persist gap closed (#909)

auser:
Keeps `specs/REFACTOR-STATUS.md` current after #909 landed (the code fix didn't touch the rollup).

#866 wired Phase 2a (macOS egress secret-substitution endpoint over vsock-5253) but it was **silently non-functional** on a plain `up`/`invoke --hypervisor vz`/`libkrun`: `VzBackend::start`/`LibkrunBackend::start` decode the secret bindings from `<state_dir>/plan.json`, but the CLI persisted that file pre-start only for firecracker. So the endpoint never spawned unless `MVM_GATEWAY_BRIDGE=1` happened to thread/persist the plan. #909 closed the gap with `persists_plan_before_start(hyp) = matches!(hyp,"firecracker"|"vz"|"libkrun")` on both the `up` and `invoke`/`run` launch arms (QEMU excluded — in-memory config).

This PR:
- Marks the **2a** checkbox done and records the #909 default-path fix.
- Updates the Plan 197 summary line accordingly.
- Notes the control plane is live-proven on vz (endpoint spawns; guest receives the placeholder only) and that the data-plane e2e remains gated on the separate Vz host→guest 5252 early-boot vsock race — not this code.

Docs-only.


auser:
Follow-on: #921 carries the **vz Phase 2a data-plane live proof** (httpbin reflects the real cred, guest holds only the `mvm-secret-…` placeholder = claim 13, non-allowed host refused = claim 12, repeated dials confirm the 5253 listener re-accepts) plus the matching REFACTOR-STATUS / SPRINT / blueprint updates. This PR had already merged, so the proof landed as a sibling rather than an addition here.

PR #957, opened 2026-06-16
feat(plan-193): WS-2.2a — lower egress policy into rvproxy's native [policy] config

auser:
## Plan 193 WS-2.2a — policy lowering (the config half)

WS-2 (port claim-10 egress enforcement off mvm's in-line splice onto rvproxy's native flow API) is now **unblocked**: rvproxy R2 shipped (flow-lifecycle events, flow-context + `FlowKill` + sticky teardown + the over-MTU/unserializable guard, and the rule-carrying redaction filter — all merged in `rvproxy`).

This is the first, parity-first slice. rvproxy runs as a supervised subprocess (`rvproxy run --config`), so the binding is **config**, not an in-process trait.

### What's here
`supervisor::network::rvproxy_policy::lower_policy` projects a resolved `CanonicalEgress` (+ DNS hostname allow-list) into rvproxy's `[policy]` table:
- `default_egress_deny` — claim-10's coarse deny-by-default gate
- `cidr_denylist` — the always-on mandatory-deny set (link-local + cloud metadata)
- `cidr_allowlist` — per-tenant L4 grants coarsened to their destination CIDRs

rvproxy's `[policy]` is **IP-level only**, so this is a **sound coarse pre-filter**: rvproxy denies everything outside the allowed CIDRs + mandatory-deny, and the splice's `L4PolicyScan` / `DnsSinkholeScan` / byte scans refine *within*. The composition never over-permits (rvproxy's allow is a superset; the splice still drops wrong-port/proto/host) and never under-blocks. `RvproxyPolicyGaps` enumerates what does **not** lower (port/proto-scoped rules, DNS hostnames, byte scans) so the splice is kept for it — **the coarsening is explicit, never silent**.

`RvproxyPolicy::permits_dest` mirrors rvproxy's `policy_destination_reason` precedence exactly, giving a **unit parity oracle** — the foundation for making the WS-1.5 gate binary-discriminating later.

### Tests (7)
Verdict-agreement with `CanonicalEgress::permits` across a probe battery (incl. mandatory-deny IPs); soundness (coarse filter never under-blocks a permitted destination); the gap enumeration; `[policy]` TOML field-name serialization. **RED-verified** by dropping the mandatory-deny denylist (metadata/link-local then leak). nightly fmt + `clippy -D warnings` + `check-no-spec-refs-in-comments` clean.

### Scope / what's next
Pure lowering — no launch-path change, no live boot, **deletes nothing** (parity-first per the WS-2 failing-test plan). Remaining sub-slices, tracked in Plan 193:
- **2b** — config emission + `rvproxy run --config` native launch behind a flag (overlaps WS-3)
- **2c** — subscribe to rvproxy's `FlowEvent` export → re-feed the chain-signed audit (needs the rvproxy JSONL/UDS `FlowEvent` export follow-up)
- **2d** — extend the WS-1.5 parity gate to the native path, then delete the splice + Plan-141 `on_packet` hooks once green

Two rvproxy-side follow-ups this surfaced: a JSONL/UDS **`FlowEvent` export** (rvproxy slice-2 deferral) and **port/proto + DNS-hostname policy config** (rvproxy slice-1 covers only `default_egress_deny` + CIDR allow/deny).

### DIFF
--- a/crates/mvm-hostd/src/supervisor/network/mod.rs
+++ b/crates/mvm-hostd/src/supervisor/network/mod.rs
@@ -34,6 +34,9 @@ pub mod flow_byte_log;
 pub mod latency;
 pub mod packet;
 pub mod pipeline;
+/// Lowering of a resolved egress policy into rvproxy's native `[policy]` config
+/// (claim-10 enforcement onto the rvproxy substrate).
+pub mod rvproxy_policy;
 /// Substitution/scan egress-proxy seams.
 pub mod stages;
 
--- a/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
+++ b/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
@@ -1,0 +1,306 @@
+//! Lower a resolved egress policy into rvproxy's native `[policy]` config table.
+//!
+//! The config half of moving claim-10 egress enforcement off mvm's in-line
+//! splice and onto rvproxy's native flow API. rvproxy runs as a supervised
+//! subprocess (`rvproxy run --config`), so the binding is *config*, not an
+//! in-process trait — mvm emits the `[policy]` table rvproxy enforces.
+//!
+//! ## What lowers, and what stays in the splice
+//!
+//! rvproxy's `[policy]` table is an **IP-level** allow/deny: `default_egress_deny`
+//! + `cidr_allowlist` + `cidr_denylist`. It expresses exactly:
+//!   - deny-by-default (claim-10's coarse gate),
+//!   - the always-on mandatory-deny set (link-local + cloud metadata) as a
+//!     denylist,
+//!   - per-tenant L4 grants *coarsened to their destination CIDRs*.
+//!
+//! It cannot express port/proto scoping, DNS hostname allow-lists, or the byte
+//! scans (placeholder-leak, undeclared redaction). So this lowering is a **sound
+//! coarse pre-filter**: rvproxy denies everything outside the allowed CIDRs and
+//! the mandatory-deny set, and the splice's `L4PolicyScan` / `DnsSinkholeScan` /
+//! byte scans refine *within* what rvproxy admits. The composition never
+//! over-permits (rvproxy's allow is a superset of the precise policy; the splice
+//! still drops wrong-port/proto/host) and never under-blocks (every destination
+//! some rule could permit is admitted at the IP layer).
+//!
+//! [`RvproxyPolicyGaps`] enumerates what did NOT lower so the caller keeps the
+//! splice for it — the coarsening is explicit, never silent.
+
+use mvm_core::network_policy::MANDATORY_DENY_RANGES;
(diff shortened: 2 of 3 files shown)


auser (commented):
**Review (WS-2.2a — policy lowering).** Solid and sound. The coarse-IP-pre-filter framing is the right shape, the gap enumeration makes the coarsening explicit, and the parity oracle holds up. I checked the soundness crux against rvproxy `main`:

- **`permits_dest` faithfully mirrors rvproxy's enforcement.** rvproxy's current `policy_destination_reason` (gateway `mod.rs`) has exactly the three-clause precedence `permits_dest` reproduces — including the `default_egress_deny && cidr_allowlist.is_empty()` deny-by-default clause. So the oracle isn't drifting from the binary. (And rvproxy's new L4-aware decision fn "reduces exactly to `policy_destination_reason` when `l4_allowlist` is empty" — which this PR leaves empty, so the IP-only oracle stays valid.)
- **Never-under-blocks holds by construction**, not just by the probe battery: `cidr_allowlist` = ∪ rule CIDRs, so any `(proto, ip, port)` a rule permits has `ip ∈ rule.net ⊆ allowlist`; and `egress.permits` excludes mandatory-deny IPs, which are exactly the denylist set — so no permitted destination is caught by the denylist. The composition is a clean AND (rvproxy coarse ∧ splice precise); over-permitting at the IP layer is safe *while the splice is retained*. ✅

Three forward-looking notes (none block 2a — it deletes nothing):

1. **The L4 config you flagged as a needed follow-up has landed** — rvproxy #115 added `l4_allowlist` (`{protocol, cidr, port_lo, port_hi}`). 2a stays IP-only (correct for a coarse pre-filter), but 2b can now lower into `l4_allowlist` to actually *close* the `l4_scoped_rules` gap rather than leaving it to the splice. (Same for the FlowEvent JSONL/UDS export — landed as rvproxy #113 — which 2c needs.)

2. **🔴 2d must gate splice-deletion on `RvproxyPolicyGaps` being closed.** The coarse filter is sound *only because the splice still refines within*. Once 2d deletes the splice, rvproxy-only over-permits — whole CIDR at all ports/protos, and no DNS-hostname enforcement at all. So 2d can only fire when `l4_scoped_rules` → `l4_allowlist`, `dns_hostnames` → a rvproxy DNS allow-list (**still missing** — only `allow_dns_*` booleans exist; tracked in rvproxy #118), and `byte_scans_in_splice` → the host-side terminator. `RvproxyPolicyGaps` is the perfect tripwire; recommend 2d hard-assert all gaps are closed/accounted before removing the splice.

3. **For 2c, wire the `secret-redaction-filter` / audit export with `include_payload_bytes: false`** (rvproxy #108 review finding) so the FlowEvent/decision stream can't carry cleartext.

Nits (non-blocking): `permits_dest` reparses the CIDR strings on each call — fine for a config/oracle path; a `proptest` over random IPs would harden the by-construction soundness argument beyond the fixed probe set.

Nice slice.

PR #964, opened 2026-06-16
feat(plan-193): WS-2 full-fidelity policy lowering (L4 proto/port + DNS sinkhole)

auser:
## Plan 193 WS-2 — full-fidelity policy lowering

**Stacked on #957** (the coarse IP lowering). Review/merge after it; base is `feat/plan-193-ws2-policy-lowering`.

#957 lowered only the coarse IP layer because rvproxy's `[policy]` table was IP-only. rvproxy has since shipped **L4 port/proto rules** (rvproxy #115) and a **DNS hostname allow-list** (rvproxy #120), so this completes the lowering to **full claim-10 fidelity**:

- L4 grants → `l4_allowlist` (proto + CIDR + inclusive port range), replacing the coarse `cidr_allowlist` projection — the decision is now port/proto-precise.
- DNS hostnames → `dns_allowlist` (dotted-suffix sinkhole).
- mandatory-deny → `cidr_denylist` (unchanged).

`RvproxyPolicy::permits_flow(proto, ip, port)` mirrors rvproxy's `policy_flow_reason` and is **proven verdict-identical to `CanonicalEgress::permits` for every probed proto/ip/port** — full L4 parity, not just the IP layer. `dns_permits` mirrors `DnsSinkholeScan`'s dotted-suffix match (lookalikes rejected).

With L4 + DNS lowering natively, `RvproxyPolicyGaps` collapses to a single residual: the **byte scans** (placeholder-leak + undeclared redaction), which ride the transform path, not `[policy]`.

### Tests
Full proto/ip/port parity vs `CanonicalEgress::permits` (Unrestricted / Rules / deny-all / mandatory-deny-overrides-grant); DNS dotted-suffix incl. lookalike rejection; TOML shape (`[[l4_allowlist]]`, `dns_allowlist`). Port precision **RED-verified**. 7 tests green; nightly fmt + clippy + `check-no-spec-refs` clean. Pure — deletes nothing.

### Remaining WS-2
- **2b** — wrap the lowered `[policy]` in the full rvproxy config (network/api/forward/audit; the audit section points the `FlowEvent` JSONL/UDS export at mvm for 2c) + flagged `rvproxy run --config` launch. Real config-parse validation lands here via the parity gate.
- **2c** — subscribe to rvproxy's `FlowEvent` export → re-feed `signer_task`'s chain.
- **2d** — extend the WS-1.5 parity gate to the native path, then delete the splice.

### DIFF
--- a/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
+++ b/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
@@ -5,29 +5,27 @@
 //! subprocess (`rvproxy run --config`), so the binding is *config*, not an
 //! in-process trait — mvm emits the `[policy]` table rvproxy enforces.
 //!
-//! ## What lowers, and what stays in the splice
+//! ## Fidelity
 //!
-//! rvproxy's `[policy]` table is an **IP-level** allow/deny: `default_egress_deny`
-//! + `cidr_allowlist` + `cidr_denylist`. It expresses exactly:
-//!   - deny-by-default (claim-10's coarse gate),
-//!   - the always-on mandatory-deny set (link-local + cloud metadata) as a
-//!     denylist,
-//!   - per-tenant L4 grants *coarsened to their destination CIDRs*.
+//! rvproxy's `[policy]` table expresses the full coarse + L4 + DNS shape of
+//! claim-10's egress gate:
+//!   - `default_egress_deny` — the deny-by-default coarse gate,
+//!   - `cidr_denylist` — the always-on mandatory-deny set (link-local + cloud
+//!     metadata),
+//!   - `l4_allowlist` — per-tenant L4 grants at full proto + CIDR + port-range
+//!     precision (rvproxy `policy_flow_reason`),
+//!   - `dns_allowlist` — the upstream DNS hostname sinkhole (dotted-suffix).
 //!
-//! It cannot express port/proto scoping, DNS hostname allow-lists, or the byte
-//! scans (placeholder-leak, undeclared redaction). So this lowering is a **sound
-//! coarse pre-filter**: rvproxy denies everything outside the allowed CIDRs and
-//! the mandatory-deny set, and the splice's `L4PolicyScan` / `DnsSinkholeScan` /
-//! byte scans refine *within* what rvproxy admits. The composition never
-//! over-permits (rvproxy's allow is a superset of the precise policy; the splice
-//! still drops wrong-port/proto/host) and never under-blocks (every destination
-//! some rule could permit is admitted at the IP layer).
-//!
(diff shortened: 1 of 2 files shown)


auser (commented):
**Review (WS-2 full-fidelity lowering — stacked on #957).** Strong approve. This closes the L4 + DNS gaps my #957 review flagged, and the parity oracle holds up to independent checking.

**Verified `permits_flow` is verdict-identical to rvproxy's `policy_flow_reason`** (read against rvproxy `main`, not just the in-PR `CanonicalEgress::permits` tests). Walked every branch of both:

| case | rvproxy `policy_flow_reason` | `permits_flow` |
|---|---|---|
| denylist match | deny | deny ✓ |
| `l4_match \|\| cidr_match` | allow | allow ✓ |
| non-empty allow-list, no match | deny (`*_allowlist_miss`) | deny ✓ |
| both allow-lists empty + `default_egress_deny` | deny | deny ✓ |
| both empty + not deny-by-default | allow | allow ✓ |

They agree on every path — including the subtle "empty allow-lists + deny-by-default" corner. And `dns_permits` matches rvproxy #120's `dns.rs` dotted-suffix *and* mvm's `DnsSinkholeScan` (all three normalize + `query == suffix || query.ends_with(".{suffix}")`), so the three parties agree by construction.

**`RvproxyPolicyGaps` now collapses to `byte_scans_in_splice` only** — exactly right: L4/DNS enforce natively, leaving only the byte scans (placeholder-leak + undeclared redaction). Under Option A those move to mvm's host-side TLS terminator, not the gateway. So after this slice the splice's `L4PolicyScan` + `DnsSinkholeScan` are redundant and 2d can delete them once the parity gate is green.

Notes (none blocking):

1. **The oracle is hand-mirrored, so it can silently drift** if rvproxy changes `policy_flow_reason`. You've got the durable guard planned (2d's binary-discriminating WS-1.5 gate runs the witnesses against the live native path) — just calling out that until 2d lands, `permits_flow`≡rvproxy is comment-asserted + verified-by-eye, not CI-enforced. Worth a `// keep in sync with rvproxy policy_flow_reason` pointer.
2. **Minor DNS normalization nit:** `dns_permits` uses `trim_matches('.')` (both ends) on query + suffix; rvproxy's `normalize_dns_name` trims trailing only. Diverges only for malformed leading-dot inputs (`.foo.com`) — harmless for valid names, and the 2d gate would catch a real divergence, but aligning the trim avoids a theoretical parity blip.
3. **Forward (2d gating):** with gaps == {byte_scans}, the splice's L4/DNS scans can go, but the placeholder-leak + redaction residual must land in the terminator (Option A) before the splice is *fully* removed. The `RvproxyPolicyGaps` residual is the right tripwire — keep 2d asserting on it.

Verified-by-source on the oracle; nice slice. (Review/merge after #957, per the stack.)

PR #970, opened 2026-06-16
fix(plan-125): bind the host-services broker to the backend-specific BROKER_PORT socket (vz)

auser:
## What

`spawn_broker` bound the **libkrun-shaped** `vm_vsock_port_socket` (`<state>/vsock-<port>.sock`) unconditionally. vz nests its per-port vsock listeners under `<state>/vsock/` and splices `BROKER_PORT` to `vm_vz_vsock_port_socket` — the same layout the egress-substitution endpoint already uses (`vz.rs` substitution path). On vz the broker therefore bound a path the supervisor never forwards the guest's `connect_host_vsock(BROKER_PORT)` dial to, leaving the in-guest `host.audit.v1` leg **unreachable**.

## Fix

Thread the broker listen socket into `BrokerServicesSpawnParams` / `BrokerSpawnParams`, computed by each backend's `start()`:
- **libkrun** → `vm_vsock_port_socket(name, BROKER_PORT)`
- **vz** → `vm_vz_vsock_port_socket(name, BROKER_PORT)`

`spawn_broker` now binds exactly the path it's handed instead of assuming one layout. `vm_name` (only ever used to compute that path) drops out of `BrokerSpawnParams`; the audit-signer keeps it for the per-VM chain key.

## Why this matters

Prerequisite for the Plan 125 E5.3b-4 live-VM `host.audit.v1` round-trip on the vz backend (the in-guest variant-B proof). Without it the guest can boot and dial but never reach the broker on macOS 26 / vz.

## Tests

- The admit + direct `spawn_broker` unit tests now point the supplied socket at a **vz-style `vsock/` subdir** and assert the broker bound *there* — a regression guard proving the path is threaded through, not recomputed as the libkrun flat layout. Readiness polls the supplied path, so a recompute regression would time out the spawn.

## Gates (local)

`cargo test -p mvm-backend --lib broker_services_spawn` (6 green) · `clippy -p mvm-backend --all-targets -D warnings` · `cargo check -p mvm-backend --all-targets` · nightly `fmt --all --check` · `xtask check-no-spec-refs-in-comments` — all clean.

### DIFF
--- a/crates/mvm-backend/src/broker_services_spawn.rs
+++ b/crates/mvm-backend/src/broker_services_spawn.rs
@@ -221,9 +221,13 @@ pub struct BrokerSpawnParams<'a> {
     pub workload_id: &'a str,
     /// Tenant id.
     pub tenant_id: &'a str,
-    /// VM name; the broker binds the per-VM `BROKER_PORT` socket keyed on it
-    /// (the VMM forwards the guest's `connect_host_vsock(BROKER_PORT)` there).
-    pub vm_name: &'a str,
+    /// The per-VM `BROKER_PORT` UDS the broker binds — the path the active
+    /// backend's VMM splices the guest's `connect_host_vsock(BROKER_PORT)` dial
+    /// to. libkrun binds `<state>/vsock-<port>.sock`
+    /// (`vm_vsock_port_socket`); vz nests it under `<state>/vsock/`
+    /// (`vm_vz_vsock_port_socket`). The caller computes it so the broker binds
+    /// where its own VMM actually listens, instead of assuming one layout.
+    pub broker_listen_socket: &'a Path,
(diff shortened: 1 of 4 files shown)


auser:
Closing as a duplicate of #971 — an independent parallel session landed the identical fix (backend-supplied `broker_listen_socket`, libkrun→`vm_vsock_port_socket` / vz→`vm_vz_vsock_port_socket`). Both were opened within ~4 min and neither existed when I started. Deferring to #971 to keep review focused.

One bit worth grafting from here: my admit + `spawn_broker` tests point the supplied socket at a vz-style `vsock/` subdir and assert the broker bound *there* — a direct regression guard for the vz path (readiness polls the supplied path, so a recompute regression times out the spawn). #971 relies on the existing `handle.uds_path == passed path` assertion, which is also adequate.

PR #971, opened 2026-06-16
fix(plan-125): vz broker binds the wrong BROKER_PORT socket (round-trip broken on vz)

auser:
**Bug found while grounding the E5.3b-4 live E2E: the host-services broker round-trip is silently broken on vz.**

`broker_services_spawn::spawn_broker` bound `vm_vsock_port_socket(vm, BROKER_PORT)` (`<state>/vsock-5300.sock`) **unconditionally**. That's correct for libkrun (the VMM proxies the guest's `connect_host_vsock(BROKER_PORT)` straight to it) but **wrong for vz**, which splices `host_listen_ports` to `vm_vz_vsock_port_socket` (`<state>/vsock/vsock-5300.sock` — the `vsock/`-subdir path the substitution endpoint binds, `vz.rs:1797`). So on vz the broker was bound one directory up from where the supervisor forwards the dial → the **vz guest leg was unreachable**, despite "round-trip live on vz".

### Why it slipped through
The host-spine integration test (#969) and the variant-A host-side proof both connect to the broker's **bound** UDS directly, bypassing the VMM splice — so neither exercises the path the guest actually traverses. Only the guest leg (VMM-forwarded) hits the mismatch.

### Fix
The broker's listen path is now **backend-supplied**, not derived:
- `BrokerServicesSpawnParams.broker_listen_socket` → `BrokerSpawnParams.broker_uds_path`
- `LibkrunBackend::start` passes `vm_vsock_port_socket`; `VzBackend::start` passes `vm_vz_vsock_port_socket`
- `spawn_broker` binds the passed path (no internal derivation)

### Tests / gates
`cargo test -p mvm-backend` green (329; the `spawn_broker` test already asserts `handle.uds_path == the passed path`, so it now guards that the param is honored — updated the two `BrokerServicesSpawnParams` constructions for the new field). clippy `-D warnings`, nightly fmt `--all`, `check-no-spec-refs-in-comments`, `check-core-runtime-free`, `cargo check --workspace --all-targets` clean.

Off `main`. Independent of #969 (test-only) and #965/#967 (codegen/veneer). E5.3b-4 variant A (host-side round-trip) was proven live separately this session.

### DIFF
--- a/crates/mvm-backend/src/broker_services_spawn.rs
+++ b/crates/mvm-backend/src/broker_services_spawn.rs
@@ -221,9 +221,12 @@ pub struct BrokerSpawnParams<'a> {
     pub workload_id: &'a str,
     /// Tenant id.
     pub tenant_id: &'a str,
-    /// VM name; the broker binds the per-VM `BROKER_PORT` socket keyed on it
-    /// (the VMM forwards the guest's `connect_host_vsock(BROKER_PORT)` there).
-    pub vm_name: &'a str,
+    /// The per-VM `BROKER_PORT` socket the broker binds — the **backend-specific**
+    /// path the VMM forwards the guest's `connect_host_vsock(BROKER_PORT)` to:
+    /// `vm_vsock_port_socket` on libkrun (the VMM proxies straight to it),
+    /// `vm_vz_vsock_port_socket` on vz (the supervisor splices to it). Passed in
+    /// rather than derived because the two backends use different socket paths.
+    pub broker_uds_path: &'a Path,
     /// Per-VM state dir; holds the broker PID file.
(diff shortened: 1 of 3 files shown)


auser:
Heads up: this duplicates #970 (independent parallel session, same fix + same backend-supplied-socket design; both opened within ~4 min). I've closed #970 in favor of this one. Optional graft from #970: its admit/`spawn_broker` tests point the listen socket at a vz-style `vsock/` subdir as an explicit regression guard for the vz path.

This is the prerequisite for the vz leg of the E5.3b-4 in-guest round-trip — see #973, which proves the round-trip live on **libkrun** (the vz leg needs this fix).

PR #973, opened 2026-06-16
feat(plan-125): in-guest audit-probe + opt-in mkGuest bake (E5.3b-4 in-guest driver)

auser:
## Why

The Plan 125 E5.3b-4 live-VM `host.audit.v1` round-trip (variant B — the in-guest headline) had **no in-guest driver baked into any guest rootfs**. The host spine is proven (PR #969, host-side, real bins, no VM); this supplies the missing in-guest half.

## What

Option (a) — the E2E-proof path. (Option b, baking the Python SDK so `import mvm; mvm.audit.emit(...)` runs in-guest, remains the productized route and is tracked under b3.)

- **`crates/mvm-guest/src/bin/audit-probe.rs`** — calls `mvm_guest::host_audit::emit` from inside the guest, dialing `connect_host_vsock(BROKER_PORT)` so an entry travels the full spine: guest → backend vsock splice → per-VM `mvm-broker` → `mvm-audit-signer` → `~/.mvm/audit/<tenant>.<vm>.workload.jsonl`. Modes:
  - `emit` — one normal record lands (prints chain head)
  - `oversized` — >4 KiB record → host refuses with `BadRequest`
  - `rate` — burst past the 20/s token bucket → at least one `RateLimited`
  - `all` — all three; exit 0 only if every step met its success condition, so the workload exit status is itself an assertion.
  It **cannot express a category** (`EmitRequest` has no such field), so the host's forced `workload_audit` stamp is the only category a workload entry can carry.
- **`nix/packages/mvm-audit-probe.nix`** — cross-compiles the bin from the workspace, mirroring `mvm-exit-report.nix`.
- **`nix/lib/mk-guest.nix`** — opt-in `withAuditProbe` flag bakes it at `/usr/local/bin/audit-probe`. **Off by default** → the production guest closure never carries it (sibling boundary to `mvm-guest-agent.nix` not selecting it).
- **`examples/audit-probe/`** — fixture flake that runs the probe (mode `all`) as the sealed PID-1 workload, so an admitted `mvmctl up` (tenant defaults to `local`) drives the whole round-trip.

## Independence

Independent of the vz broker-socket fix (#970): this works on libkrun today. The vz path additionally needs #970.

## Gates (local)

`clippy -p mvm-guest --bin audit-probe -D warnings` · cross-compiles `aarch64-unknown-linux-musl` · nightly `fmt --all --check` · `xtask check-no-spec-refs-in-comments` — clean. The `build_record` unit test runs on the Linux test lane (`cfg target_os = "linux"`). The nix package + flake build through the builder VM (no host Nix); the live boot is the E5.3b-4 closeout (separate).

### DIFF
--- a/crates/mvm-guest/Cargo.toml
+++ b/crates/mvm-guest/Cargo.toml
@@ -61,6 +61,13 @@ path = "src/bin/mvm-guest-netinit.rs"
 name = "syscall-probe"
 path = "src/bin/syscall-probe.rs"
 
+# In-guest host.audit.v1 driver for the live-VM round-trip proof. Baked only
+# by the opt-in `withAuditProbe` mkGuest flag (a test fixture image), never by
+# the production guest closure.
+[[bin]]
+name = "audit-probe"
+path = "src/bin/audit-probe.rs"
+
 # Test fixture for the warm-process worker pool. Speaks the
 # `worker_protocol` framed multi-call loop on stdin/stdout. Behavior
 # selectable via `MVM_FAKE_RUNNER_BEHAVIOR` env var so tests can
--- a/crates/mvm-guest/src/bin/audit-probe.rs
+++ b/crates/mvm-guest/src/bin/audit-probe.rs
@@ -50,6 +50,9 @@ mod linux {
     /// Records to burst in `rate` mode — comfortably over the 20/s bucket so at
     /// least one must be refused once the initial capacity drains.
     const RATE_BURST: usize = 40;
+    /// Attempts the first `emit` makes before giving up — covers the brief
+    /// boot window before the BROKER_PORT host-listen splice is wired.
+    const AUDIT_CONNECT_RETRIES: usize = 20;
 
     /// Build the workload record for `action`/`seq`. Pure so the field shape is
     /// unit-testable without a broker. `ts` is a workload field the host passes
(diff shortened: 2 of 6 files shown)


auser:
## Proven live (libkrun, macOS 26 / Apple Silicon)

Booted an admitted VM from `examples/audit-probe` and ran the probe in `all` mode end to end:

- per-VM `mvm-broker` + `mvm-audit-signer` spawned (`vsock-5300.sock` / `broker.pid` / `audit-signer.sock` present)
- **22 entries** landed in `~/.mvm/audit/local.e53b4probe.workload.jsonl` — `probe-emit: 1` + `probe-rate: 21`
- every entry `category: workload_audit` (host-forced — the probe carries no category field), server-authoritative `brk-*` correlation id, valid `prev_hash` chain
- `verify_workload_chain` → **clean: 22 entries, rc=0** under the host-signer key
- 20/s rate-limit observable: 22 of 40 burst emits landed; the rest refused
- workload exits `0` (emit ok + oversized→BadRequest + rate→limited all pass)

The retry commit (`run_emit` retries on transport errors) was added after the first live run: the standalone first emit lost a race against the supervisor wiring up the BROKER_PORT host-listen splice at boot. With the retry, the round-trip is clean and `workload.exit=0`.

### Two findings worth a maintainer's eye (not blockers for this PR)

1. **Broker-spawn is gated behind `MVM_GATEWAY_BRIDGE=1`.** On a plain `mvmctl up --hypervisor libkrun --tenant local`, `should_thread_signed_plan(false, "libkrun")` is `false`, so `populate_audit_substrate` never runs, `config.tenant_id` stays `None`, and the per-VM broker is a defused no-op — `host.audit.v1` is silently unavailable. The live proof needed `MVM_GATEWAY_BRIDGE=1` to thread the plan. Decoupling broker-spawn from the egress bridge would let `host.audit.v1` work on a normal admitted `up`.
2. **The bridge supervisor's claim-10 `validate_audit_substrate` hard-pins the host-signer key under `~/.mvm/keys`**, rejecting an isolated `MVM_DATA_DIR` ("must canonicalize under ~/.mvm/keys"). Fine for the trust boundary, but it means the broker path can't run under a relocated data dir.

PR #980, opened 2026-06-16
fix(plan-193): match rvproxy's dns_hostname_allowlist field name (+ WS-2 status reconcile)

auser:
## Defect fix — DNS sinkhole field-name mismatch

The rvproxy DNS hostname sinkhole shipped via a maintainer PR (rvproxy #119) that named the `[policy]` field **`dns_hostname_allowlist`**. mvm's policy lowering (#964) serialized it as **`dns_allowlist`**.

rvproxy parses its config with `#[serde(default)]`, so the mismatched field would be **silently ignored** once mvm emits a config in 2b — leaving the DNS sinkhole **disabled on the native path**, a latent claim-10 hole with no error. This renames mvm's `RvproxyPolicy` field (and its TOML output) to `dns_hostname_allowlist` to match rvproxy's actual schema.

Caught in review **before 2b wires config emission**, so the field was never actually emitted yet — pure rename, no behavior change.

Also reconciles the Plan 193 WS-2 notes: the DNS sinkhole landed as rvproxy **#119** (the original #120 was closed), and the field name is now correct.

### Tests
`rvproxy_policy` parity + serialization tests updated for the field name; 7 green; nightly fmt + clippy + `check-no-spec-refs` clean.

### DIFF
--- a/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
+++ b/crates/mvm-hostd/src/supervisor/network/rvproxy_policy.rs
@@ -14,7 +14,7 @@
 //!     metadata),
 //!   - `l4_allowlist` — per-tenant L4 grants at full proto + CIDR + port-range
 //!     precision (rvproxy `policy_flow_reason`),
-//!   - `dns_allowlist` — the upstream DNS hostname sinkhole (dotted-suffix).
+//!   - `dns_hostname_allowlist` — the upstream DNS hostname sinkhole (dotted-suffix).
 //!
 //! So the lowering is an exact projection of the resolved `CanonicalEgress`
 //! (proven by the parity tests: `permits_flow` agrees with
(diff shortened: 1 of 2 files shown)


auser (commented):
**Review (DNS field-name fix).** Correct and complete — approve. Verified against rvproxy `main`:

- rvproxy's TOML `[policy]` schema (`rvproxy_policy::PolicyConfig`) names the field **`dns_hostname_allowlist`**, so #964's `dns_allowlist` would have serialized a field rvproxy's parser ignores — the DNS sinkhole would have silently not taken effect on the native path. This rename fixes that, and all references are updated consistently (struct field, `dns_permits` + doc, `lower_policy`, the two tests, and the plan doc). Good catch.
- **My #964 review missed this** — I verified the match *logic* was identical but didn't check the serialized field *name* against rvproxy's schema. Mea culpa; this is the right fix.
- **L4 names are fine** (I checked, since it's the same bug class): rvproxy's `L4RuleConfig` is `{protocol, cidr, port_lo, port_hi}`, exactly matching `RvproxyL4Rule` — no analogous mismatch there.
- `dns_permits` still mirrors rvproxy's `hostname_matches_allowlist` (`dns.rs`) and mvm's `DnsSinkholeScan` — dotted-suffix, normalized. The `#119` doc reference is correct (the merged DNS PR; #120 was a closed duplicate).

One residual edge-nit (pre-existing, not introduced here): `dns_permits` normalizes with `trim_matches('.')` (both ends) while rvproxy's `normalize_dns_name` trims trailing only — diverges only for malformed leading-dot inputs (`.foo.com`), never for valid qnames. Harmless, and 2d's parity gate would catch any real divergence; flagging only for completeness.

Nice — this closes a real latent gap.

PR #1005, opened 2026-06-17
docs(plan-200): correct network-policy note (VmStartConfig has no network_policy field)

auser:
An implementation attempt of the transient `--net`/`--allow-host` slice disproved the key claim in #1004's note: **`VmStartConfig` has no `network_policy` field** (the struct with one is `mvm_backend::microvm::FlakeRunConfig`). The run path's egress is enforced by the **network provider's `deny_all()` default**, not a `VmStartConfig` field.

**Consequence:** enforcing a transient `--net` is **not** a one-line thread — it requires adding the field to `VmStartConfig` + wiring each workload backend's `start()` to apply it (FC iptables, libkrun/vz gateway), which is per-backend, claim-10-critical, and needs **live egress testing**. The CLI resolve/flags/audit-label parts of the note remain correct; added a hard note that enforce-less `--net` must not ship.

The CLI-side wiring was attempted and compiled cleanly; it was reverted (not shipped) because the backend enforcement is the larger, box-gated half. Docs-only.

### DIFF
--- a/specs/plans/200-machine-ux-dx-layer.md
+++ b/specs/plans/200-machine-ux-dx-layer.md
@@ -528,18 +528,33 @@ Implementation shape:
 >   `deny_all()`. `up` already exposes `--network-preset`/`--network-allow`
 >   through it. Mirror that; the ergonomic `--net`/`--allow-host` names map onto
 >   the same `(preset, allow)` signature.
-> - **`VmStartConfig.network_policy` already exists** and is applied by the
->   backend (the run path defaults it to `deny_all` via `..Default::default()`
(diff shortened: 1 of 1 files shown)


auser:
Obsolete: a parallel session shipped the full --net/--allow-host enforcement slice in #1003 (WS-B) — including adding VmStartConfig.network_policy and the per-backend application, and live-validated it. This note's correction ("VmStartConfig has no network_policy field") was true when written but #1003 made it false, so merging would inject a misleading note. The stale #1004 'not yet built' note is being replaced with a 'shipped via #1003' pointer in a follow-up.

PR #1026, opened 2026-06-17
docs(plan-200): refresh Session 3 prompt + hand off OCI-boot blocker

auser:
Folds the last unclaimed WS-B deferred item — `run --image <oci>` won't boot (missing `mvm-meta.json` sidecar) — into the **Session 3** prompt as item **1a**, the natural owner (the `MachineImageSource` → admission leg).

## Why it's handed off rather than fixed inline

Investigation (confirmed unclaimed: no active worktree/PR; only Session 1 `mcpwarm` and Session 2 `p200-l4` are running) showed this is **not** the "write a sidecar" fix it looked like:

- `admit_overlay_aware(rootfs_dir)` runs on **every** backend start and refuses any rootfs without a host-side `mvm-meta.json` saying `overlay_aware: true`. The OCI materialize path (`rootfs.rs`) writes only `rootfs.ext4`.
- `overlay_aware: true` means the rootfs carries `/mvm/runtime` + an overlay-preferring `/init` — an arbitrary OCI image has neither. The gate also refuses `overlay_aware: false`, so there's **no safe partial fix**.
- The OCI run path attaches **no** runtime overlay (`run_inner`'s `ImageSource::Prebuilt` leaves `runtime_overlay_path = None`), so the gate is a **false positive** on OCI.

So it needs a **design decision** (claim-adjacent gate) — (A) scope `admit_overlay_aware` to actual overlay attachment vs (B) make the OCI rootfs genuinely overlay-aware — plus **live builder-VM + boot verification** (`tests/oci_image_runner_smoke.rs` is gated off and doesn't exercise the gate). That belongs with Session 3's admission work and its STEP-0 owner confirmation, not a drive-by.

## Changes (docs only)

- Session 3 prompt: added **item 1a** with the full findings + the design question + the live-verification requirement.
- Deferred-follow-up list: marked the item **handed to Session 3 item 1a** with a pointer.

No code.

### DIFF
--- a/specs/plans/200-machine-ux-dx-layer.md
+++ b/specs/plans/200-machine-ux-dx-layer.md
@@ -998,9 +998,10 @@ WS-B change:
       OCI materialize path produces a rootfs without the W6.2 `mvm-meta.json` sidecar /
       W1.4b runtime-overlay mount point, so `admit_overlay_aware` refuses it. **Handed to
       Session 3 item 1a** (the `MachineImageSource` admission leg) — investigation showed it
-      is a design decision (scope the gate to actual overlay attachment vs. make OCI rootfs
-      overlay-aware), not a sidecar write, and needs live builder-VM verification. See the
-      Session 3 prompt below for the full findings + the design question.
+      is not a sidecar write: `run --image` drives the guest via the in-guest agent, which an
+      OCI image lacks, so the runtime overlay (carrying the agent) must be ATTACHED to OCI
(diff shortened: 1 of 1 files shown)


auser:
Closing as superseded by #1036. This PR refreshed the Session 3 OCI handoff prompt; #1036 now implements that handoff (run --image boots OCI images end-to-end) and carries the current Plan 200 docs/status updates. Keeping both open would preserve stale prompt-only guidance after the implementation exists.

PR #1038, opened 2026-06-17
Fix libkrun/Vz up egress enforcement

auser:
## Summary
- default admitted libkrun/Vz up boots onto the signed-plan gateway bridge instead of gvproxy-direct
- lower non-deny resolved NetworkPolicy values into generated signed-policy bundles with host-pinned TCP L4 rows, and carry those bundles through up and transient run
- keep deny-all on local-default, keep Firecracker on its nftables default path, and document the claim-10 posture in ADR/Plan/Sprint rollups

## Tests
- cargo fmt --check
- git diff --check
- cargo test -p mvm-cli admission_weaves --lib
- cargo test -p mvm-cli libkrun_and_vz_thread_signed_plan_by_default --lib
- cargo test -p mvm-hostd open_mode_lowers_to_unrestricted_egress --lib
- cargo test -p mvm-cli --lib
- cargo check --workspace
- cargo clippy --workspace -- -D warnings
- cargo test --workspace -- --test-threads=1

## Notes
- default parallel cargo test --workspace exposed the existing mvm lib global-state race; the failed tests passed individually and the full workspace passed serially
- live VM revalidation was not run in this pass due the known builder-VM flakiness; deterministic bridge/admission coverage is included


auser:
Live validation update (2026-06-17):

Ran the egress-probe on this branch with isolated MVM_DATA_DIR and the shared builder cache:

- First run: `up --hypervisor libkrun --flake examples/egress-probe --wait --network-allow one.one.one.one:443` reached boot but failed before PID with `resolve observer chain from admitted plan`. Root cause: generated in-memory policy refs were still being resolved through disk policy observer lookup. Fixed in `f613764a`: bridge entrypoints now use supplied `bundle_json.network.observers` when present.
- Rerun with hostname allow-list: supervisor starts, generated bundle is present in audit, exits `3` (both blocked). Audit shows early `l4-policy` drop.
- Rerun with raw IP allow-list: `--network-allow 1.1.1.1:443` also exits `3`.
- Control run with `MVM_ACK_UNRESTRICTED_NETWORK=1 --network-preset unrestricted` also exits `3`. Audit shows egress/ingress bridge flows open and no L4 drop under unrestricted.

Conclusion: the signed-plan/generated-bundle path is wired far enough for the bridge to start and audit the generated bundle, but the live libkrun gateway-bridge data path is still not delivering outbound TCP, even when policy is open. This draft PR should not be marked ready until that bridge data-path blocker is fixed or the default-bridge-on change is reconsidered.

Issue #1119, opened 2026-06-20
Plan 205: builder-residency Step 2 — live-coupled mechanism (snapshot-park, idle keeper, warm auto-start)


auser:
Tracking the remaining **live-coupled** half of the builder-residency story. Step 1 (PR #1114, merged) made `MVM_RESIDENCY` govern the builder *routing* (`cold` → ephemeral single-shot builder; `warm`/`parked` → persistent) and added a `builder residency` line to `mvmctl doctor` — all pure and CI-tested. Step 2 is the lifecycle *mechanism*: it boots, snapshots, suspends, and reaps a real builder VM, so it can only be validated on a **macOS-26 Apple Silicon box** (the same live-gating as the Plan 118/159 standby-pool live lanes).

**Full plan:** `specs/notes/plan-205-builder-residency-step2-followup.md`

### What Step 1 deliberately left as degrade-to-current
- `MVM_RESIDENCY=warm` does **not** auto-start a persistent builder — it only keeps one already started via `persistent-builder start`.
- `MVM_RESIDENCY=parked` **degrades to warm** (keeps the persistent builder) — no builder snapshot-park yet.
- `MVM_RESIDENCY=cold` skips the persistent builder for *new* builds but does **not** tear down an already-running one.
- No builder **idle timeout** — a persistent builder stays up until explicit `dev down` / `persistent-builder stop`.

### Workstreams
- [ ] **S2.1 — Builder VM snapshot-park.** Wire vz saved-state (Plan 159 snapshot path) into the persistent-builder boot path: snapshot an idle builder, restore sub-second on the next build. `parked` stops degrading-to-warm; doctor reports `parked (snapshot present|none)`.
- [ ] **S2.2 — Idle-timeout keeper.** Demote a persistent builder after `ResidencyPolicy::idle_timeout()` idle (`parked`→snapshot+suspend, `cold`→teardown, `warm`→keep). Prefer an invocation-driven check over a new daemon. Pure decision fn unit-tested; live action gated.
- [ ] **S2.3 — `dev up` / warm auto-start.** When policy is `warm` and no persistent builder is active, auto-start one; explicit `persistent-builder start`/`stop` always win.
- [ ] **S2.4 — Active teardown on `cold`.** Stop a running persistent builder when policy is `cold`, not just skip routing.

### Acceptance (measured on a macOS-26 box)
- `warm`: a second build reuses the builder with no boot.
- `parked`: idle builder parks to a snapshot and restores sub-second on the next build.
[...]

PR #1142, opened 2026-06-20
feat(plan-205): snapshot park/resume the Vz dev builder (S2.1 live, unblocks #1119)

auser:
**Slice 2** of builder snapshot-park — the live payoff of issue #1119: the Vz dev builder is now **snapshot-parked on `dev down` and resumed on `dev up`** (no cold boot), gated on the residency policy. Builds on Slice 1 (#1135, the snapshot-capable builder).

## Behavior (residency-gated)
- **`dev down`** — when residency allows a persistent builder (warm/parked, not `cold`) and the supervisor is alive and not `--reset`: SAVE the builder to `<state_dir>/parked.vzsnapshot` (+ `.machine-id`), then stop. Otherwise: plain stop + clean any stale snapshot.
- **`dev up`** — when allowed and a parked snapshot is present: RESTORE from it (re-acquire nix-store flock, respawn gvproxy, restore the Vz machine state) instead of cold-booting; the snapshot is consumed. Otherwise: cold-boot as before.
- **`cold` residency / `--reset`** — unchanged behavior (cold boot / full stop), stale markers cleaned so a later `dev up` never wrongly resumes.

## Mechanism (all `mvm-build`-local — no `mvm-backend` dep)
- `builder_control` module: a std-only control-socket client + a pure `builder_restore_config` (flips `startup_mode` to `Restore`, preserves all other fields).
- `builder_snapshot_save`/`builder_snapshot_restore`/`is_builder_parked` in `vz_builder.rs` — restore mirrors the bootable tail of `start()` (lock → gvproxy → supervisor in Restore mode → handle).
- `dev_vz.rs` wiring via tested `should_park`/`should_resume` gates; `--reset` threaded to all three `cmd_dev_vz_down` call sites.
- `SessionRecord.snapshot_path` (`#[serde(default)]`) for the dispatch-routing path.

## Tests / review
53 new unit tests (control round-trip, restore-config preservation, parked-state, gating truth tables); root binary builds with `builder-vm`. Whole-branch reviewed (opus): live-coupled correctness verified — no gvproxy leak window, park-before-stop with safe fallback, resume detaches + waits for guest-agent readiness, marker hygiene on every path. Review also caught + I reverted an unrelated deletion of the stale-supervisor boot diagnostic.

Live macOS-26 validation (cold `dev up` → `dev down` parks → `dev up` resumes) is in progress; the unit-tested decision + restore logic gate every code path that doesn't require a live builder boot.

## Next (tracked in `specs/notes/plan-205-builder-snapshot-park-execution.md`)
Slice 3 — the idle keeper (auto-park after N minutes idle) consuming the shipped `decide_builder_residency_action`.

### DIFF
--- a/crates/mvm-backend/src/vz.rs
+++ b/crates/mvm-backend/src/vz.rs
@@ -350,7 +350,8 @@ impl VmBackend for VzBackend {
             {
                 bail!(
                     "supervisor exited before writing PID file (status: {status}). \
-                     Check stderr above for VZ configuration errors. Console log: {}",
+                     Check stderr above for VZ configuration errors.{} Console log: {}",
+                    stale_supervisor_hint(&supervisor_path),
                     console_log.display()
                 );
             }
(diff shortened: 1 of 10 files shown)


auser:
Superseded by #1139 (parallel session shipped the park/restore primitives + explicit `dev park` + restore-on-`dev up` + doctor). Re-doing just the non-redundant delta — auto-park on `dev down` (residency-gated) — as a small PR on top of #1139's primitives.

PR #1150, opened 2026-06-20
feat(plan-175): VMGenID delivery on PostRestore (Task 1)

auser:
## Plan 175 — Task 1: VMGenID delivery on resume

The smallest end-to-end increment of Plan 175 (Firecracker live-memory warm-start): make a snapshot restore actually rotate the guest CSPRNG, so two clones of one snapshot don't generate identical secrets. The host already had the token mint (Plan 122 D) and the PostRestore sender (#734); the guest already had the reseeder (122 D). **Only the wire payload was missing — this wires it.**

### What changed
- `GuestRequest::PostRestore` is now a struct variant `{ token: [u8; GENID_BYTES] }`. `#[serde(default)]` → an all-zero token means "no rotation" (template-restore / no-fresh-generation callers), keeping them honest without a second request type.
- `GuestResponse::PostRestoreAck` gains `reseeded: bool` (`#[serde(default)]`).
- The guest agent feeds the delivered token to a **process-resident `GenIdReseeder`** (held in a `static Mutex`, seeded with the all-zero baseline). Its state rides the VM memory snapshot, so two clones restored from one snapshot both diverge from the captured value when the host hands each a distinct fresh token. New zero-aware dispatch wrapper `GenIdReseeder::on_post_restore_token`.
- Both host senders mint a fresh token per resume: `VsockPostRestoreSignal { token }` (the `mvmctl resume` path) and `post_restore_at(path, token)` (the backend restore path). `mvmctl resume` surfaces `VMGenID rotated (CSPRNG reseeded)` when the guest reseeds.
- `GenIdState::new` / `GenIdReseeder::new` are now `const fn` (enables the plain `static`).
- Regenerated protocol stubs (`schema/protocol-v0.json` + Python/TS `_protocol`) via `cargo xtask gen-stubs`; `check-stubs` is clean.

### Tests (host + unit, no KVM)
- `genid::tests::post_restore_zero_token_is_no_rotation` — zero = no-op, a later real token still rotates, idempotent re-send.
- `genid::tests::two_clones_of_one_snapshot_rotate_to_distinct_state` — both clones reseed from the shared snapshot state.
- `vsock::tests::post_restore_token_roundtrips_and_defaults_to_zero` — non-zero round-trip, `{"PostRestore":{}}` defaults to zero, `deny_unknown_fields` rejects extras.
- `vsock::tests::post_restore_ack_carries_reseeded_flag` — `reseeded` round-trips; a pre-rotation ack defaults to `false`.
- `instance_snapshot::tests::signal_post_restore_*` updated for `reseeded`.

`cargo fmt --all --check`, `cargo clippy -D warnings`, and the changed-crate unit suites are green on macOS.

### Out of scope / still gated
- **Step 3 (live-KVM)**: snapshot→restore round-trip asserting `/dev/urandom` divergence across two restores + `reseeded == true`. It rides **Task 4**'s Firecracker restore driver (a real mvm guest-agent snapshot/restore over vsock) and remains gated.
- **Tasks 2–4** (UFFD/NBD/hugepages substrate; SIGUSR1 primed barrier; `FirecrackerBackend::warm_start` + verb + `agent_ping` e2e) remain.

Tracking updated in `specs/plans/175-*.md` and `specs/REFACTOR-STATUS.md` in the same change.


auser:
**Linux verification (live /dev/kvm box, native cargo).** Built and ran the changed-crate unit suites natively on x86_64 Debian — all green, including `mvm-backend` (which is codesign-SIGKILL'd under macOS amfid locally):

```
mvm-core    : 209 passed
mvm-guest   : 368 passed
mvm         : 1311 passed
mvm-backend : 307 passed
EXIT=0  (2195 tests, 0 failures)
```

Covers the cfg(linux) guest-agent + Firecracker backend paths the Mac can't exercise.

PR #1206, opened 2026-06-21
feat(plan-203): TranscriptObserver capture engine + Firecracker-bridge tap

auser:
Extends forensic transcript capture (Plan 203) to the **Firecracker** workload
backend. Slice 2 (#1182/#1190) landed the byte-capture sink + tap on the
libkrun `gateway_bridge`; this adds the matching tap on the Firecracker bridge —
the other claim-10 workload backend — so an armed capture records egress/ingress
bytes on both paths.

## What

- `TranscriptObserver` (`mvm-hostd/src/supervisor/transcript_capture.rs`): a
  host-initiated, opt-in `Observer` (Plan 141 packet-observer pipeline) added to
  a VM's bridge observer list only when a capture is armed for that VM. `on_packet`
  always returns `Verdict::Forward` — capture never alters, delays, or drops
  traffic. Encryption + disk writes run on a dedicated worker thread fed by a
  bounded channel; frames are dropped + counted rather than back-pressuring the
  guest's network when the disk falls behind. The worker owns the
  `TranscriptWriter` and is joined on drop, sealing the manifest at bridge teardown.
- `mvm-core::transcript` additions to support the observer-side writer surface.
- Wires the tap into `mvm-firecracker-bridge` + `firecracker_bridge::parse`.

## Tests

3 unit tests in `transcript_capture.rs` (capture round-trip, drop-counting under
a stalled worker, seal-on-drop). No new runtime deps in mvm-core.

## Follow-up

- Sync `specs/plans/203-forensic-network-transcript-capture.md` (currently states
  "implementation-complete on-host" against the libkrun tap only) + REFACTOR-STATUS
  to record the Firecracker-bridge tap as a landed slice.
- Branch is a few commits behind main; the merge queue will require it current.


auser:
Closing as superseded. The transcript capture sink + gateway-bridge tap landed via #1190 (parallel implementation) while this branch independently built an equivalent `TranscriptObserver`. My contribution from the live-validation effort is instead #1214 (now merged): three real FC gateway-bridge bugs that made #1190's capture non-functional on Firecracker — plan.json envelope decode, plus a seccomp profile that had never run with seccomp on (`sigaltstack` + ~9 more syscalls). Found via a live FC e2e on a KVM box (the existing relay tests can't catch sidecar-spawn bugs).

PR #1217, opened 2026-06-21
feat(plan-204): scope typed BuildGuestImage route to override-free builds (WS-D)

auser:
## Summary

Follow-up to #1215 (typed `BuildGuestImage` `dev_build` route, opt-in). Closes the override-input artifact-equality gap by **scoping**, so the typed route stays default-off-safe and becomes a valid candidate for the eventual artifact-equality proof + default flip.

#1215 wired `dev_build_via_builder_vm_uncached` to the typed daemon route behind `MVM_BUILDERD_TYPED`, asserting the legacy persistent dispatch is override-free. That is **not** true for a source-checkout user flake: `stage_flake_dispatch_job` stages the workspace and `render_flake_cmd_sh` bakes `--override-input mvm path:/work/nix --override-input mvm/mvm-workspace …` into `cmd.sh`, building the workload against the in-repo `mvm` checkout. The typed `BuildGuestImage` op carries no overrides field, so a typed build of such a flake resolves `mvm` differently and is **not artifact-equal**.

This PR scopes the typed route to genuinely override-free builds:

- `try_typed_persistent_build` now takes `flake_ref` and bails to the legacy dispatch via `typed_route_blocked_by_local_override` whenever the legacy override would apply — keyed on the same `local_mvm_workspace` (source-checkout + GitHub-pin) signal the legacy path uses.
- Corrects the false "ignores the source-checkout staging" claim in the code comment, the Plan 204 doc, `REFACTOR-STATUS.md` (glance + detail + a dated header block), and `SPRINT.md`.

Default-off, so **zero behaviour change**. This is the precondition for a live override-free artifact-equality proof (shell vs typed) before flipping `resolve_route`'s build default.

## Tests

- `typed_route_blocked_for_source_checkout_github_pinned_flake`, `typed_route_allowed_for_override_free_flake`.
- `dev_build` / `persistent_builder` / `builder_route` suites green (41 / 16 / 10).
- Gates green: `cargo fmt --all -- --check`, `cargo clippy -p mvm-build --features builder-vm --all-targets`, `check-no-spec-refs-in-comments`, `check-builder-shell-job-sites` (4/allowlisted, no new site), `check-no-host-nix`.

## Remaining (next, separate)

Builder-image rebake (the running daemon predates #1213's `output_dir`), a live override-free artifact-equality proof per backend, then flip `resolve_route`'s build default, then the raw-shell debug gate (gated on the flip).


auser:
Closing — the guard this PR adds rests on a false premise I introduced.

The typed route (`try_typed_persistent_build`) is reached **only** from the
live-persistent-session branch, so its sole competitor is the **persistent**
legacy dispatch. That path (`PersistentBuilderVm::run_build` →
`persistent_builder::stage_flake_dispatch_job`) builds `nix build /work#<attr>`
with **no `--override-input`** — it computes `mounts.staged_user_flake` upstream
but drops it; only `mounts.artifact_out` is used. The source-checkout override
lives in the *single-shot* path (`builder_vm_runtime::stage_flake_dispatch_job`),
which the typed route never competes with. The typed daemon builds the same
`nix build /work#<attr>` override-free.

So typed-persistent and shell-persistent are artifact-equal by construction, and
#1215's original "ignores the source-checkout staging" claim was correct. This
guard is unnecessary, and worse: it would block the typed route whenever the user
flake pins `github:tinylabscom/mvm`, sabotaging the live artifact-equality proof
on a source checkout.

Proceeding instead to the live proof (rebake + typed-vs-shell on the persistent
`/work` build), then the `resolve_route` default-flip, then the raw-shell gate.

PR #1230, opened 2026-06-21
chore(deps): patch rand + time security advisories

auser:
## What

`cargo audit` flagged two actionable advisories. Both are in-major patch bumps.

| Crate | Advisory | Change |
|---|---|---|
| `rand` | RUSTSEC-2026-0097 (unsound — aliased mutable ref in `ThreadRng` reseed; via `quinn-proto`/`reqwest`) | 0.9.2 → 0.9.4 |
| `time` | RUSTSEC-2026-0009 (DoS — parser stack exhaustion; via `sigstore`/`rcgen`) | 0.3.45 → 0.3.47 |

`rand` was a **new, unaccepted** finding — it was not in `deny.toml`. `time` was previously accepted there; 0.3.47 actually fixes it, so the stale ignore entry is removed.

## MSRV stays at 1.85

`time 0.3.47` declares Rust 1.88, but only needs it to *build* — the CI/dev toolchain is 1.95 and there is no MSRV gate, so `--locked` builds are unaffected. An initial bump to `rust-version = 1.88` was reverted: at MSRV 1.88 clippy begins enforcing let-chain `collapsible_if` workspace-wide (let-chains stabilized in 1.88), which fails the Lint job on pre-existing code. Promoting the floor to 1.88 + the let-chain modernization is a deliberate separate change, not security-patch scope.

## `rsa` (RUSTSEC-2023-0071) — accepted, no clean fix

Traced in full and documented in `deny.toml`:
- Reaches the **release** binary (`manifest-verify`) via `openidconnect → sigstore`.
- **Not exploitable here:** Marvin targets RSA *private-key* ops (decrypt/sign); mvm only *verifies* cosign bundles with public keys (`image_verify.rs`), and bundles are ECDSA P-256 — the rsa path isn't exercised.
- **Cannot be trimmed:** `rsa` is non-optional in sigstore 0.13. A sigstore 0.14 experiment confirmed 0.14 drops the *direct* edge but `rsa` still arrives via `verify → fulcio → oauth → openidconnect`. Real fix is upstream `rsa` 0.10 (pre-release) adopted by `openidconnect`. Tracked in `specs/plans/36`.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` → clean
- `cargo deny check advisories` → advisories ok
- `cargo fmt --all -- --check` → clean
- Remaining `cargo audit` warnings are unmaintained-crate notices already documented in `deny.toml`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3678,9 +3678,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "521739c6d2bac4aa25192232afe6841231376b2b26d4d9fae5ecf8ca5772e441"
 
 [[package]]
 name = "num-integer"
@@ -4562,7 +4562,7 @@ dependencies = [
  "bytes",
  "getrandom 0.3.4",
  "lru-slab",
- "rand 0.9.2",
+ "rand 0.9.4",
  "ring",
  "rustc-hash",
  "rustls",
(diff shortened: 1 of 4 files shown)


auser:
Superseded by #1234, which eliminates all 10 advisories outright (including rsa, via the sigstore-verify migration) instead of fixing a subset and accepting rsa. #1234 carries these same time/rand fixes.

PR #1239, opened 2026-06-22
feat(build): extend libkrun→qemu auto-fallback to the dev_build + Stage 0 paths

auser:
Stacked on #1237. Base retargets to `main` once #1237 merges.

## Why

#1237 added the libkrun→qemu auto-fallback for the **OCI-materialize** path (`machine run --image`). The other build entry points — `mvmctl build` / `up --flake` / `template build` — go through `dev_build`'s single-shot `run_build`, which wraps the `BuilderVmError` into an **anyhow** chain, so #1237's `run_with_builder_fallback` (keyed on a raw `BuilderVmError`) couldn't be applied at that call site. This finishes the job.

## What

- **Preserve the builder error** in the anyhow chain — `anyhow::Error::new(e).context("builder VM")` instead of `anyhow!("builder VM: {e}")` (which stringified it) — so it stays downcastable.
- **`run_with_builder_fallback_anyhow`** — same policy as the typed helper, but fires only when the anyhow chain carries a VMM-level `BuilderVmError` (`anyhow_has_builder_vm_level_failure` walks `Error::chain()` + `downcast_ref`). A genuine build error (or a stringified one) surfaces unchanged, no retry.
- Wrap the `dev_build` single-shot call site in it. The **persistent-builder path already falls through to single-shot on failure**, so it's covered too. Explicit `--builder`/`MVM_BUILDER_BACKEND` opts out.

## Verification

`cargo fmt --all --check`, `cargo clippy -p mvm-build --features builder-vm --tests -D warnings`, `mvm-cli` build: clean. New unit tests: the anyhow-chain detection (wrapped VMM failure → true; wrapped build error → false; **stringified → false**, proving why the preservation change matters), the loop retrying qemu on a wrapped `SupervisorExited`, and a real build error surfacing without retry.

The fallback mechanism itself was **live-proven in #1237** (libkrun rc -22 → qemu → alpine booted); this PR reuses the same infra + qemu builder via the anyhow-wrapped path.

### DIFF
--- a/crates/mvm-build/src/builder_backend_select.rs
+++ b/crates/mvm-build/src/builder_backend_select.rs
@@ -185,7 +185,19 @@ pub fn resolve_builder_backend_with_override(
 /// `verbose` streams the libkrun console; the QEMU path always logs to
 /// `console.log`.
 pub fn resolve_stage0_backend(verbose: bool) -> Box<dyn BuilderVm> {
-    match resolve_choice() {
+    resolve_stage0_backend_for_choice(resolve_choice(), verbose)
+}
+
+/// Stage 0 driver for an explicit `choice` — used by the auto-fallback loop to
+/// construct the next backend to try. QEMU when chosen; libkrun for everything
+/// else (Vz Stage 0 is a gap, and the "Stage 0 is libkrun even on Vz-default
+/// hosts" invariant holds — and the Linux fallback order only ever yields
+/// libkrun→qemu, never Vz).
+pub fn resolve_stage0_backend_for_choice(
+    choice: BuilderBackendChoice,
+    verbose: bool,
+) -> Box<dyn BuilderVm> {
+    match choice {
         BuilderBackendChoice::Qemu => Box::new(QemuBuilderVm::new()),
         _ => Box::new(LibkrunBuilderVm::default().with_verbose(verbose)),
     }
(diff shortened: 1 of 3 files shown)


auser:
Extended to also cover **Stage 0** (the from-scratch builder-image/kernel bootstrap): `run_stage0` is a `BuilderVm` trait method returning `BuilderVmError` directly, so the typed `run_with_builder_fallback` applies cleanly. Added `resolve_stage0_backend_for_choice` and wrapped all three Stage 0 call sites (root-dir build, external-kernel rootfs, kernel compile).

So the libkrun→qemu auto-fallback now covers **every** local builder entry point: OCI materialize (#1237), dev_build flake, and Stage 0. 33 mvm-build tests green; clippy/fmt clean across mvm-build + mvm-cli.

auser:
Final coverage: also routed the CLI's dev-image (`build_image_via_libkrun`) and default-microvm (`build_default_microvm_via_libkrun`) loops through the **shared** `builder_attempt_order` (they previously had a local Vz→libkrun-only order). So **every** builder entry point now shares one fallback policy: OCI materialize, `dev_build` flake (`up --flake`/`build image`/`template build`), Stage 0, the dev-image loop, and the default-microvm loop. `up --flake` was already covered via `dev_build`; this closes `up`/`dev up`/`up --dev`. Tests green, clippy/fmt clean.

PR #1265, opened 2026-06-22
fix(builder): page-align the libkrun builder kernel (Linux KVM rc -22)

auser:
## What

libkrun maps the external builder kernel into guest memory with the `vmlinux` file's **exact size**. Linux KVM rejects a `KVM_SET_USER_MEMORY_REGION` whose size isn't a multiple of the host page size, so an unaligned `vmlinux` fails VM creation with `EINVAL` (`rc -22`).

`page_aligned_kernel` (in `mvm_build::libkrun_builder`) zero-pads the kernel up to a page boundary in a cached `vmlinux.page-aligned` sibling before `krun_set_kernel`. No-op when already aligned or off Linux. The padded copy is made writable before `set_len` so a non-root builder doesn't hit `EACCES` on the read-only Nix-store kernel.

## Root cause (corrects ADR-093)

`strace -f` of the failing supervisor on a bare-metal Linux KVM box:

```
ioctl(KVM_SET_USER_MEMORY_REGION, {slot=1, guest_phys_addr=0x80000000,
      memory_size=8963072, ...}) = -1 EINVAL
```

`8963072` is exactly the builder `vmlinux` file size, and `8963072 % 4096 = 1024` — not page-aligned. The failing region is the **kernel**, independent of guest RAM. ADR-093 framed this as a libkrun ">4 GiB" defect; it's an mvm-addressable alignment bug. macOS HVF has no such requirement, which is why the same unaligned kernel boots libkrun on macOS. ADR-093 carries a dated correction.

## Verification

- Unit tests: `align_up`, `write_zero_padded_copy` (pads + preserves prefix + zero tail), read-only-source regression (the `EACCES` case), `aligned_copy_is_current`, `page_aligned_kernel` pass-through + Linux-gated padding. All green; `cargo fmt`/`clippy -D warnings` clean for `mvm-build`.
- **Live on the box:** with the unaligned kernel on disk, the code auto-creates `vmlinux.page-aligned`, every `KVM_SET_USER_MEMORY_REGION` succeeds, and `KVM_RUN` runs the vCPUs — `rc -22` is gone.

## Scope

This removes the unaligned-kernel EINVAL as a trigger of the ADR-093 libkrun→qemu fallback; the fallback stays for other VM-creation failures. A *separate* later issue on at least one host (guest userspace not reaching `cmd.sh` under libkrun — empty console, no nix output) is tracked independently and is not addressed here.

### DIFF
--- a/crates/mvm-build/src/libkrun_builder.rs
+++ b/crates/mvm-build/src/libkrun_builder.rs
@@ -1264,10 +1264,23 @@ impl BuilderVmImage {
 /// pre-populated with the variant's kernel cmdline (where applicable).
 /// `RootDir` images carry no cmdline — libkrun handles `set_root`
 /// mode without one.
-/// Page size KVM's `KVM_SET_USER_MEMORY_REGION` aligns to (4 KiB on x86_64 /
-/// aarch64 Linux, where the bug below manifests; larger guest pages are still
-/// multiples of it).
-const KVM_REGION_ALIGN: u64 = 4096;
+/// The alignment KVM's `KVM_SET_USER_MEMORY_REGION` enforces: the **host** page
+/// size. 4 KiB on x86_64 and 4K-page aarch64, but 16 KiB / 64 KiB aarch64 hosts
+/// exist (RHEL/CentOS 64K-page kernels), where a kernel aligned only to 4 KiB
+/// still trips `EINVAL`. Query `sysconf(_SC_PAGESIZE)` so the padding is correct
+/// on every host instead of assuming 4 KiB; fall back to 4 KiB only if the query
+/// is unavailable.
+fn host_page_size() -> u64 {
+    #[cfg(unix)]
+    {
+        // SAFETY: `sysconf` is a side-effect-free query with no preconditions.
+        let v = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+        if v > 0 {
+            return v as u64;
+        }
+    }
+    4096
+}
 
 /// Round `n` up to the next multiple of `align`.
 fn align_up(n: u64, align: u64) -> u64 {
@@ -1286,6 +1299,10 @@ fn align_up(n: u64, align: u64) -> u64 {
 /// `vmlinux.page-aligned`) created next to it and reused across builds. Trailing
 /// zeros are safe — libkrun loads the kernel from its own headers; the padded
 /// tail is unused guest RAM.
+///
+/// Scope: only the external-`vmlinux` (`Rootfs`) path runs this. `RootDir` /
+/// Stage 0 boots use libkrun's bundled libkrunfw kernel via `set_kernel`, which
+/// libkrun maps itself — they never reach here and need no padding.
 fn page_aligned_kernel(kernel_path: &Path) -> Result<PathBuf, BuilderVmError> {
     if !cfg!(target_os = "linux") {
         return Ok(kernel_path.to_path_buf());
@@ -1298,10 +1315,11 @@ fn page_aligned_kernel(kernel_path: &Path) -> Result<PathBuf, BuilderVmError> {
             ))
         })?
         .len();
-    if size.is_multiple_of(KVM_REGION_ALIGN) {
+    let align = host_page_size();
+    if size.is_multiple_of(align) {
         return Ok(kernel_path.to_path_buf());
     }
-    let aligned_size = align_up(size, KVM_REGION_ALIGN);
+    let aligned_size = align_up(size, align);
     let aligned_path = kernel_path.with_extension("page-aligned");
     if !aligned_copy_is_current(&aligned_path, aligned_size, kernel_path) {
         write_zero_padded_copy(kernel_path, &aligned_path, aligned_size)?;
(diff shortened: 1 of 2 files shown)


auser (commented):
Reviewed — strong PR. The `strace`-backed root cause is conclusive and correctly supersedes the ADR-093 ">4 GiB" framing (the rejected ioctl is the kernel region, size == vmlinux file size, `% 4096 == 1024`). Fix is centralized at `krun_context_for_image` (all 5 libkrun boot sites route through it), uses an atomic temp+rename, and handles the read-only Nix-store source with a regression test. A few robustness notes inline — the page-size one (#1) is the only one I'd want addressed before this is relied on across host types; the rest are optional. No conflict with the #1256 health cache / ADR-093 fallback — those stay as defense-in-depth for other VM-creation failures.

crates/mvm-build/src/libkrun_builder.rs
```diff
@@ -1264,6 +1264,108 @@ impl BuilderVmImage {
 /// pre-populated with the variant's kernel cmdline (where applicable).
 /// `RootDir` images carry no cmdline — libkrun handles `set_root`
 /// mode without one.
+/// Page size KVM's `KVM_SET_USER_MEMORY_REGION` aligns to (4 KiB on x86_64 /
+/// aarch64 Linux, where the bug below manifests; larger guest pages are still
+/// multiples of it).
+const KVM_REGION_ALIGN: u64 = 4096;
```
auser: **(Medium) Hardcoded 4096 → query the host page size.** KVM aligns `KVM_SET_USER_MEMORY_REGION` to the *host* `PAGE_SIZE`. On x86_64 and 4K-page aarch64 this is correct and the fix works. But aarch64 Linux hosts built with **16K/64K pages** would still get `EINVAL` — a 4K-aligned size isn't 16K-aligned. Suggest `libc::sysconf(_SC_PAGESIZE)` and align to that; over-aligning to a larger page is always safe. Cheap, removes a latent portability gap.

crates/mvm-build/src/libkrun_builder.rs
```diff
@@ -1264,6 +1264,108 @@ impl BuilderVmImage {
...
+/// Copy `source` to `dest` and zero-extend it to `padded_size` (`set_len`
+/// zero-fills the tail). Written to a temp sibling then renamed so a concurrent
+/// reader never sees a half-written kernel.
+fn write_zero_padded_copy(
+    source: &Path,
+    dest: &Path,
+    padded_size: u64,
+) -> Result<(), BuilderVmError> {
+    let mut tmp = dest.as_os_str().to_owned();
+    tmp.push(".tmp");
```
auser: **(Low-Med) Concurrent-writer race on the fixed `.tmp` name.** Two builds sharing a cache dir both write `vmlinux.page-aligned.tmp`; `fs::copy` + `set_len` to the same path can interleave before the rename. The atomic rename protects *readers*, not concurrent *writers*. A unique suffix (PID or random) closes it. Low-probability (builds usually serial per host), but worktree/parallel-test setups in this repo do share cache dirs.

crates/mvm-build/src/libkrun_builder.rs
```diff
@@ -1264,6 +1264,108 @@ impl BuilderVmImage {
...
+/// creation fail with `EINVAL` (`rc -22`) — distinct from a `nix build` error,
+/// and previously masked by the ADR-093 libkrun→QEMU fallback. macOS HVF has no
+/// such requirement, so this is a no-op off Linux.
+///
+/// Return a kernel path whose file size is page-aligned: the original when it
+/// already is, otherwise a zero-padded sibling (`vmlinux` →
+/// `vmlinux.page-aligned`) created next to it and reused across builds. Trailing
+/// zeros are safe — libkrun loads the kernel from its own headers; the padded
+/// tail is unused guest RAM.
+fn page_aligned_kernel(kernel_path: &Path) -> Result<PathBuf, BuilderVmError> {
```
auser, line 1306: **(Low) Scope confirm — Stage 0 / `RootDir`.** This guards the external-`vmlinux` (`Rootfs`) path. `RootDir`/Stage 0 boots use the *bundled* libkrunfw kernel via `libkrun-sys::set_kernel`, outside this function. Almost certainly already page-aligned (libkrun release artifact), but worth a one-line confirmation — if it ever isn't, a fresh-host Stage 0 boot on Linux would hit the same EINVAL here, uncaught.

specs/adrs/093-linux-builder-libkrun-fallback.md
```diff
@@ -111,3 +111,37 @@ fallback).
...
+ioctl(KVM_SET_USER_MEMORY_REGION, {slot=1, guest_phys_addr=0x80000000,
+      memory_size=8963072, ...}) = -1 EINVAL
+```
+
+`memory_size=8963072` is exactly the builder `vmlinux` file size, and
+`8963072 % 4096 = 1024` — **not page-aligned**. Linux KVM requires
+`KVM_SET_USER_MEMORY_REGION` sizes to be a multiple of the host page size; mvm
+passed the kernel to libkrun (`krun_set_kernel`), which maps it verbatim, so an
+unaligned `vmlinux` fails VM creation regardless of guest RAM (the earlier
+"2 GiB boots past it" reading was a coincidence of layout). macOS HVF imposes no
```
auser: **(Low) The "2 GiB booted past it" anomaly isn't fully explained.** If the *kernel* region fails on this host, the 2-GiB-RAM case should fail identically at the same ioctl. "Coincidence of layout" is plausible but unverified — a confirming `strace` of the 2-GiB case would close the loop (otherwise a second factor may be lurking). Not a blocker; the failing-case strace is authoritative. Nit: the `Context` section above still reads ">4 GiB" until this correction — a forward-pointer up top + ticking follow-up #1 resolved would help future readers.

auser:
Addressed all four review points in 413fce9c:

- **Medium (hardcoded 4096):** now queries `sysconf(_SC_PAGESIZE)` for the host page size (fallback 4 KiB), so 16K/64K-page aarch64 hosts get correctly-aligned kernels.
- **Low-Med (tmp race):** temp sibling is now `.tmp.<pid>`; concurrent builders no longer share a `.tmp` path — each renames its byte-identical copy atomically.
- **Low (scope):** doc note added — RootDir/Stage 0 use the bundled libkrunfw kernel via `set_kernel` and never hit this external-kernel padding path.
- **Low (ADR-093 anomaly):** retracted the 'coincidence of layout' claim — the kernel region size is RAM-independent, so a smaller-RAM run fails at the same ioctl; recorded as an unexplained prior observation rather than a mechanism.

Tests now use the host page size (correct on any host) + added `host_page_size` coverage.

PR #1274, opened 2026-06-22
fix(sdk-ts): migrate live-mode Sandbox ops to `machine` verbs

auser:
**Plan 208 — TypeScript sibling of #1266** (the Python SDK migration).

## Why
The live-mode TS SDK (`sdks/typescript/src/_sandbox.ts`) shells each `Sandbox` op to `mvmctl`. #1249 folded `proc`/`fs`/`cp`/`forward`/`wait` under `machine` and renamed `down`→`machine stop`, so those bare top-level verbs no longer parse — the TS live-mode SDK is broken on `main`, exactly like Python was.

## What
- `_sandbox.ts`: `proc start`/`proc wait`/`fs write`/`cp`/`forward` → `machine …`; `down`→`machine stop`. Comments + error strings updated.
- `sandbox_live.test.ts`: fake-`mvmctl` fixture unwraps the `machine` noun (+ `down`→`stop`); all 16 verb assertions (`startsWith`/`toContain`/`toMatch`) updated. **114 vitest tests pass; `tsc --noEmit` clean.**

## Deliberately NOT changed
- **`up`** stays on its own verb (the SDK parses its `--up-json` envelope; `machine run` doesn't emit a compatible one yet). That migrates in Plan 208 sub-step 2, same as the Python PR.

Independent of #1266 (different files); both un-break their respective SDKs.

### DIFF
--- a/sdks/typescript/src/_sandbox.ts
+++ b/sdks/typescript/src/_sandbox.ts
@@ -20,7 +20,7 @@
  *   produces a Workload.
  * - `MVM_SDK_MODE=live` (Plan 73 Followup H-live): every
  *   `Sandbox` call shells to `$MVM_CLI_BIN` (`mvmctl up`,
- *   `mvmctl proc start`, `mvmctl fs write`, `mvmctl down`)
+ *   `mvmctl machine proc start`, `mvmctl machine fs write`, `mvmctl machine stop`)
  *   against a real microVM. The shell is dispatched by
  *   {@link LiveTransport} below.
  *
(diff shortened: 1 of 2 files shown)


auser:
Closing as a duplicate — the TS SDK migration to `machine` verbs already landed on `main` via #1266 (a parallel commit, `fix(sdk-typescript): migrate live-mode Sandbox ops to machine verbs`, was bundled into that PR). main already has the identical functional change + test fixture, so this PR is redundant.

PR #1282, opened 2026-06-22
test(cli): guard the CLI surface the live-mode SDKs depend on

auser:
A regression guard for the bug class that started the Plan 208 SDK thread.

## Why
The Python and TypeScript live-mode SDKs (`sdks/{python,typescript}/.../_sandbox`) shell each `Sandbox` operation to `mvmctl` as a fixed argv string (`machine proc/fs/cp/forward/stop`, `machine proc wait`, + the `up` boot call). There is **no compile-time link** between those strings and the CLI — so #1249 *silently* broke both SDKs at runtime when it folded the per-op verbs under `machine`. We only caught it via CI failures on unrelated PRs.

## What
A single test (`sdk_live_mode_shelled_commands_keep_parsing` in `crates/mvm-cli/src/commands/tests.rs`) that parses each SDK-shelled argv shape against the real `Cli`. If a future rename/removal breaks a verb the SDK depends on, **CI fails here** with a message pointing at the SDK coupling — instead of shipping a broken live mode.

Mirrors the SDKs' exact argv (incl. `machine forward … --port`, `proc start -e … --`, `up --up-json …`). **Passes on `main` today.** When `machine run --up-json` ships and the SDK swaps its `up` call (see the contract note in #1278), the `up` entry here updates alongside.

### DIFF
--- a/crates/mvm-cli/src/commands/tests.rs
+++ b/crates/mvm-cli/src/commands/tests.rs
@@ -4000,3 +4000,51 @@ fn machine_console_refused_on_sealed_image() {
         "claim-15 gate must fire through machine console path: {err}"
     );
 }
+
+/// Contract guard: the live-mode SDKs (`sdks/python/mvm/_sandbox.py`,
+/// `sdks/typescript/src/_sandbox.ts`) shell each `Sandbox` operation to
+/// `mvmctl` as a fixed argv shape. Those commands MUST keep parsing, or the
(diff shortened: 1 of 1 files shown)


auser:
Superseded by #1292 — the SDK live-mode CLI-contract guard is folded in there with the corrected `machine run -d --up-json` boot entry (the standalone `up --up-json` entry here would fail now that #1258 removed `up`). #1292 fixes the underlying breakage *and* carries the guard.

PR #1286, opened 2026-06-22
Harden 'build kernel build --boot-check' (force libkrun + builder-image precondition)

auser:
Follow-up to #1275 (merged). Live testing the new `--boot-check` surfaced two implicit requirements its name doesn't imply — this makes it deterministic and gives a fixable error instead of a deep failure.

## What live testing showed
`--boot-check` runs `up --flake examples/sleeper`, which:
1. routed through the host's **default** workload backend — Vz on macOS 26, needing its own supervisor — so the check wasn't deterministic; and
2. needs the **full builder VM image** to build the sleeper, not just the kernel that `build kernel build` produced — so on a kernel-only cache it failed 20 min in with a deep `up` error (`builder VM image not found`).

## Changes
- **Force `--hypervisor libkrun`** for the boot. libkrun drives Stage 0, so it's available wherever `--source compile` ran; the kernel is backend-agnostic, so a libkrun boot proves it boots — without depending on the host's default backend or its supervisor.
- **Builder-image precondition** — check `<cache>/builder-vm/<arch>/rootfs.ext4` up front and bail with `run \`mvmctl dev up\` once` instead of letting `up` fail deep.

## Verification
- clippy/fmt clean; mvm-cli compiles with + without `builder-vm`.
- Live end-to-end against a warm cache (builder image present) — boot-check reaches the agent over vsock. (Running; will note the result on the PR.)

Tooling only; no runtime/CI change.

### DIFF
--- a/crates/mvm-cli/src/commands/build/kernel.rs
+++ b/crates/mvm-cli/src/commands/build/kernel.rs
@@ -141,7 +141,7 @@ fn run_build(args: BuildArgs, verbose: bool) -> Result<()> {
     }
 
     if args.boot_check {
-        run_boot_check(&variants)?;
+        run_boot_check(&variants, &arch)?;
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


auser:
Live-validated on macOS 26 / libkrun (warm cache with the builder VM image present): `build kernel build --which workload --source compile --boot-check` runs the hardened path end-to-end — kernel compiles, the **builder-image precondition passes**, and the boot routes through the forced **`--hypervisor libkrun`** (`backend 'libkrun'` in the log) rather than the host default. **`EXIT=0`**, which by `run_boot_check`'s contract means both `up` and `machine wait` succeeded — i.e. the new kernel booted and the guest agent answered over vsock, then the VM was torn down. (The earlier hung run was a stale pre-hardening binary from a failed zig cross-compile, not a logic issue.)

PR #1292, opened 2026-06-22
fix(cli): `machine run --up-json`/`--ttl` + migrate SDK boot off retired `up` (un-break live mode)

auser:
## Regression
#1258 removed the `up` verb, but the live-mode SDKs still shell to `up --up-json` to boot — so **SDK live-mode boot is broken on `main`** (`mvmctl up` → unrecognized subcommand). This is the regression the #1278 contract warned about; CI missed it (the SDK tests use a *mock* mvmctl), but the new guard (below) catches it.

## Fix
- **CLI:** add `--up-json` + `--ttl` to `machine run`. `--up-json` boots persistent (boot-and-return), reserves stdout, and emits exactly one envelope — `{"schema_version":1,"vm_id":"<name>","build_mode":"dev"|"prod"}` (ports the logic #1258 deleted from `up`). `build_mode` resolves from the manifest revision's build mode, else the booted image's `accessible` flag (fail-closed to `prod`). `--ttl` sets the registry `expires_at` via the same helper as `machine set-ttl`.
- **SDK:** `_sandbox.py` + `_sandbox.ts` boot call → `machine run -d --up-json …`; fixtures + assertions + stale `up`/`down` strings updated.
- **Guard:** `sdk_live_mode_shelled_commands_keep_parsing` parses every SDK-shelled command against the real `Cli` — the next rename fails CI instead of shipping a broken SDK. **(Supersedes #1282** — folded in here with the corrected `machine run` boot entry.)

Green: `mvm-cli` 1174/1174, `sandbox_live` 37 (py) + 114 (ts), fmt + clippy clean.

## ⚠️ Separate breakage found, NOT in this PR
#1258 also dropped `invoke --no-vm`, but the SDK's **remote no-vm dispatch** (`mvm/_remote.py`) still shells to `mvmctl invoke --no-vm` → `tests/test_remote_no_vm_e2e.py` fails locally (CI-gated; needs a built binary). That's a *feature drop*, not a rename, so it needs its own decision (restore the host shortcut, or rework the SDK's remote path) — flagged for follow-up.

### DIFF
--- a/crates/mvm-cli/src/commands/cmd_audit.rs
+++ b/crates/mvm-cli/src/commands/cmd_audit.rs
@@ -152,8 +152,9 @@ impl Commands {
                 // Folded advanced ops delegate to their own check.
                 machine::MachineAction::Vm(cmd) => cmd.emits_machine_readable_stdout(),
                 // `machine run --json` streams a structured MachineRunSummary;
+                // `machine run --up-json` emits the SDK boot envelope;
                 // reserve stdout so reconcile chrome can't interleave.
-                machine::MachineAction::Run(r) => r.json,
+                machine::MachineAction::Run(r) => r.json || r.up_json,
                 _ => false,
             },
             _ => false,
--- a/crates/mvm-cli/src/commands/machine/mod.rs
+++ b/crates/mvm-cli/src/commands/machine/mod.rs
@@ -198,6 +198,17 @@ pub(in crate::commands) struct MachineRunArgs {
     /// when `--name` is absent (the chosen name is printed). Implies persistence.
     #[arg(short = 'd', long)]
     pub detach: bool,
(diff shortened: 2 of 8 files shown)


auser:
Heads-up for follow-up (not fixed here): `crates/mvm-cli/tests/core_demo_e2e.rs` still shells `mvmctl up --flake …` (retired by #1258). It's **gated behind `MVM_E2E_SMOKE=1`** so it doesn't break CI, but `MVM_E2E_SMOKE=1` runs are broken until it migrates to `machine run --flake`. Left out of this PR because it's a live-boot E2E I can't verify here, and the `up --flake`→`machine run --flake` wait/output semantics need a real run to confirm. Same #1258 stale-verb family as #1293.

PR #1294, opened 2026-06-22
fix(builder-vm): mkGuest symlinks multi-output package binaries (fixes /sbin/mkfs.ext4 — OCI materialize regression)

auser:
## Symptom

`machine run --image <ref>` fails OCI rootfs materialization:

```
builder VM ext4 materialization failed: guest shell job exited 127 — stderr tail:
   /job/cmd.sh: line 8: /sbin/mkfs.ext4: not found
```

## Root cause

nixpkgs **e2fsprogs is multi-output** — `mkfs.ext4` ships in its **`bin`** output (`e2fsprogs-<v>-bin`), not the default output (libs/headers). mkGuest's binary-symlink loop iterated `${pkg}/{bin,sbin}` on the **default** output, so for any split package it found no executables and never created `/sbin/mkfs.ext4`. The materialize script (`crates/mvm-build/src/rootfs.rs`) calls `/sbin/mkfs.ext4` by absolute path → `127`.

This was masked while the OCI rootfs was cache-hit; it surfaced once that cache emptied. Both the libkrun and Vz builder paths hit it (it's in the shared mkGuest rootfs, not backend-specific). The e2fsprogs reference is old, so this is a nixpkgs-pin multi-output split surfacing in a long-standing assumption.

## Fix

Resolve each package's binary output with `lib.getBin` before symlinking:

```nix
(pkg: let binOut = lib.getBin pkg; in ''
  for srcdir in bin sbin; do
    if [ -d "${binOut}/$srcdir" ]; then ...
```

- Single-output packages → default output (**unchanged**).
- Split packages (e2fsprogs, util-linux, …) → correctly pick `.bin`.

Strictly more correct; no regression for single-output packages.

## Verification

The host can't mount the guest ext4 to inspect `/sbin` directly, and a builder-VM rebuild (flake-fingerprint change triggers it on the next `machine run --image`) is needed to observe the fix end-to-end. Confirmed up to the regression point: materialize reaches `/sbin/mkfs.ext4` and fails only because the *cached* builder rootfs lacks it; this fix makes the next rebuilt rootfs contain it.

### DIFF
--- a/crates/mvm-build/src/rootfs.rs
+++ b/crates/mvm-build/src/rootfs.rs
@@ -231,17 +231,15 @@ fn materialize_ext4_in_builder_vm(
 
 #[cfg(feature = "builder-vm")]
 fn ext4_materializer_choice() -> crate::builder_backend_select::BuilderBackendChoice {
-    use crate::builder_backend_select::{BuilderBackendChoice, resolve_env_override};
-
-    // Preserve the historical default: OCI materialization has always used
-    // libkrun unless an operator explicitly asks for QEMU. Do not use
-    // resolve_choice(), because macOS 26+ auto-detects Vz and Vz has no
-    // shell-job materializer yet.
-    match resolve_env_override() {
-        Some(BuilderBackendChoice::Qemu) => BuilderBackendChoice::Qemu,
-        Some(BuilderBackendChoice::Vz) => BuilderBackendChoice::Vz,
-        Some(BuilderBackendChoice::Libkrun) | None => BuilderBackendChoice::Libkrun,
-    }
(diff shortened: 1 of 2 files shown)


auser:
Note for reviewers/upgraders: this fix lives in `nix/lib/mk-guest.nix`, but the builder-VM cache fingerprint does **not** include that file (#1295), so an existing install needs a builder-VM cache clear (`rm -rf ~/.cache/mvm/builder-vm/<arch>`) for the rebuilt image to pick up the fix. CI / fresh environments build the image fresh and are unaffected. #1295 fixes the auto-invalidation.

PR #1379, opened 2026-07-01
feat(backend): HvfBackend opt-in relay egress path (Plan 214 P2.3)

auser:
Routes `HvfBackend::start` onto the relay egress path so the host-side gating endpoint becomes the sole path off the box — claim-10 enforced at the endpoint, the run loop a pure relay — replacing the legacy in-loop supervisor gate. Stacks on P2.1 (`WorkloadRunner<D,S>`); base branch `plan-214-p21-base` is a throwaway ref at that commit so this PR shows only its own delta (retarget to `feat/plan-214-hvf-spike` once that's on origin).

## What changed

- **`HvfBackend::start` opt-in relay routing** (`MVM_HVF_EGRESS_RELAY`, default off): routes through `WorkloadRunner<InHouseDriver, RealEndpointSpawner>`. Off by default — the relay's WireRequest (secret-bearing) path is not yet live-proven, so legacy stays the default until that lands and the default flips (P2.4). Mirrors the P1 additive-migration discipline and the note's invariant: *legacy stays until both raw and wire relay paths are live-green*.
- **`resolve_relay_egress_inputs`**: pure mapping of an admitted `VmStartConfig` + its persisted plan → `{tenant, secrets, redaction, network_policy}`. Secret-free plans take the raw egress branch; secret-bearing plans the wire substitution branch; the endpoint gates against the config's `network_policy` either way.
- **Fail-closed reap** at the call site: if the runner errors after spawning the endpoint, reap it so a decrypted-secret process can't outlive a failed launch (covers the current `RealEndpointSpawner`-returns-`PathBuf`-no-guard shape).
- **`workload_blocks` fix**: an empty `rootfs_path` no longer synthesizes a bogus empty-source virtio-blk (the legacy backend filtered this; the spec path didn't). Enables initramfs-only guests through the WorkloadRunner path.
- **New `hvf-backend-relay-egress` example**: live-proof regression through the production backend.

Lifecycle is unchanged: the in-house driver writes the same `hvf.pid` / `console.log` / `workload.exit` that `stop`/`status`/`wait`/`list` read.

## Verification

TDD'd (RED→GREEN watched) for the pure helpers + the `workload_blocks` fix.

Local gates (all green): `cargo fmt --all --check`, `cargo clippy -p mvm-backend --all-targets -D warnings`, `cargo test -p mvm-backend --lib` → **482 passed, 0 failed**, no spec-refs in comments, `--all-targets` build clean.

**Live-proven on HVF (macOS / Apple silicon)** through the production `HvfBackend::start` relay path:

```
admitted destination reachable: true
non-admitted destination reachable: false
```

Admit VM console logged `>>> egress reply over vsock: ping`; deny VM had no reply line — the endpoint gate discriminates by policy from its new home, no guest NIC, no in-loop gate.

## Not in scope

- **Default stays legacy** — the Wire relay path live proof is P2.2; P2.4 flips the default and deletes the in-loop gate afterward.
- Does not touch the shared plan/note (concurrent session owns it); status recorded in commit messages.

### DIFF
--- a/crates/mvm-backend/examples/hvf-backend-relay-egress.rs
+++ b/crates/mvm-backend/examples/hvf-backend-relay-egress.rs
@@ -1,0 +1,204 @@
+//! Live proof that claim-10 egress is enforced through the *production*
+//! `HvfBackend` when it is routed onto the relay path (`MVM_HVF_EGRESS_RELAY=1`):
+//! the backend spawns the per-VM `mvm-substitution-endpoint` itself, threads its
+//! UDS into the guest's `EGRESS_PORT`, and boots via `InHouseDriver`. The run
+//! loop is a pure relay; the endpoint gates.
+//!
+//! Sibling to `hvf-relay-egress`, but entered through `HvfBackend::start` (not
+//! the driver directly) so it exercises the real backend routing + the
+//! `RealEndpointSpawner`, not a hand-wired harness. Runs twice against a
+//! discovered LAN echo server: once with a policy that admits it (reply expected)
+//! and once with a policy that admits only a different port (reply refused). Both
+//! verdicts must hold.
+//!
+//! macOS / Apple silicon. Build the per-VM aux bins and echo guest first:
+//! ```sh
+//! cargo build -p mvm-vm-host --bin mvm-hvf-supervisor
+//! cargo build -p mvm-hostd --bin mvm-substitution-endpoint
+//! OUT=/tmp/hvf-egress-guest bash crates/mvm-backend/examples/hvf-egress-guest/build.sh
+//! MVM_HVF_SUPERVISOR_PATH=target/debug/mvm-hvf-supervisor \
+//!   MVM_SUBSTITUTION_ENDPOINT_PATH=target/debug/mvm-substitution-endpoint \
+//!   MVM_HVF_KERNEL=/tmp/mvm-hvf-kernel/Image-builder \
+//!   MVM_HVF_INITRD=/tmp/hvf-egress-guest/initramfs.cpio \
+//!   cargo run -p mvm-backend --example hvf-backend-relay-egress
+//! ```
+
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+fn discover_lan_ipv4() -> Option<std::net::Ipv4Addr> {
+    use std::net::Ipv4Addr;
+    // SAFETY: standard getifaddrs walk; every pointer is null-checked and the list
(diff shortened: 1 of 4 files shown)


auser:
## Dependency & landing note (read before merging)

This PR **cannot merge to `main` yet**, and its base is deliberately synthetic. Recording the ordering so it isn't merged into the wrong lineage.

**What it is:** Plan 214 **P2.3** — routes `HvfBackend::start` onto the relay egress path (opt-in `MVM_HVF_EGRESS_RELAY`, default legacy) via `WorkloadRunner<InHouseDriver, RealEndpointSpawner>`, plus a `workload_blocks` fix and a live-proof example. Green CI.

**Why the base is `plan-214-p21-base`:** that's a throwaway ref at `f27c77e1` (the WorkloadRunner / P2.1 commit) pushed only so this PR's diff shows its own 3 commits instead of the whole spike stack.

**The real blocker — prerequisites aren't on a landing branch yet.** `WorkloadRunner`, `InHouseDriver`, `RealEndpointSpawner`, `mod workload_runner`, and the raw-egress endpoint serve mode exist **only on `feat/plan-214-hvf-spike`**. They are absent from `main` and from every clean landing slice (A/B/C1/C2/D/E/F). The A→D stack is *behind* the spike. So there is currently no base that contains P2.3's dependencies.

**Landing path:**
1. The driver seam (S0) + `WorkloadRunner` (P2.1) + endpoint relay mode (P1) get carved from the spike into landing slices (upstream of this PR).
2. This PR re-targets onto the real `WorkloadRunner` slice — a mechanical rebase — and becomes a proper mergeable slice.
3. `plan-214-p21-base` is then deleted.

**Note for the gateway/(D) owners:** commit `3d38a674` here fixes a Linux-only `-D warnings` failure — an unused `Read`/`Write` import in `mvm-hostd/src/supervisor/raw_egress.rs:182` (`handle_raw_conn_blocking`, spike commit `b7b88701`). It only surfaces on the Linux target, so macOS-local checks miss it; it will fail CI wherever the raw-egress endpoint lands. Please fold that fix into whichever slice carries `raw_egress.rs`.

Holding this PR until the WorkloadRunner slice lands; will re-target then.

auser:
Superseded by #1382 (the canonical in-house-VMM PR, which contains this as an ancestor). These egress slices are stacked on feature branches with no independent path to main after the spine-collapse; their content lands via #1382. Closing to avoid duplicate-merge conflicts.

PR #1382, opened 2026-07-01
feat(backend): in-house VMM driver seam + claim-10 gate relocation + WorkloadRunner (Plan 214)

auser:
## Plan 214 — in-house VMM driver seam, claim-10 gate relocation, and `WorkloadRunner`

This branch builds the first-party VMM (HVF on macOS, KVM on Linux) and the
ADR-102 `VmmDriver` seam, relocates the claim-10 egress gate out of the guest run
loop into the host-side endpoint, and introduces `WorkloadRunner` as the driver-
generic workload backend. The security-bearing path is **live-verified on HVF
(macOS-26 Apple silicon)** at each step.

### What lands

**The seam (ADR-102).** `VmmDriver` (pure VMM mechanics) + `RunningVm`, with
`MockDriver` (hypervisor-free) and `InHouseDriver` (HVF/KVM). `VmmSpec` is the
no-NIC physical recipe; a guest's only path off the box is a reserved vsock egress
port. `VmmDriver::attach` reconstructs a disk-backed handle by `VmId` for stateless
lifecycle.

**G1 — claim-10 gate relocation (endpoint is the unified bridge).**
- The `mvm-substitution-endpoint` gains the claim-10 gate (`network_policy`) and a
  raw-TCP egress mode, so it is the single host-side egress bridge for both the
  raw and the WireRequest (substitution) protocols — it refuses an unadmitted
  destination before connecting/substituting.
- The in-house VMM run loop becomes a **pure relay** (`SubstitutionBridge`
  relay-only): it pipes the guest's `EGRESS_PORT` stream to the endpoint with no
  in-loop gate. The relay has no upstream path except through the gating endpoint,
  so there is no window where guest egress reaches the network ungated.

**`WorkloadRunner` — the sole workload role over the seam.** Maps
`VmStartConfig → VmmSpec`, spawns the gating endpoint with the resolved policy,
boots via the driver, and implements `VmBackend` + `WorkloadBackend`. Selectable
via `--hypervisor inhouse` (additive; `hvf` and every existing selector are
byte-identical). Admission + audit stay in the existing CLI funnel.

### Verification

- **claim-10 live on HVF, three independent ways** — `InHouseDriver` (admit+deny),
  the production `HvfBackend` relay path, and the full `WorkloadRunner` `VmBackend`
  surface (`start→status→logs→stop`): admitted destination reachable, non-admitted
  refused, gate now in the endpoint, run loop a pure relay.
- **Wire (secret-bearing) path** — a real-bin-over-UDS integration test proves the
  endpoint's claim-10 gate refuses a *bound* destination the network policy doesn't
  admit, before the claim-12 bind-check (the exact frame the relay delivers).
- Address-agnostic proofs (LAN discovered via `getifaddrs`; guest reads the target
  from `/proc/cmdline`). Reproducible guest initramfs build under
  `examples/hvf-egress-guest/`.
- Gates green throughout: `cargo fmt --all`, per-crate `nextest`, `clippy
  --all-targets -D warnings`, `xtask check-vsock-only-egress`, `xtask
  check-claim-catalog` (16 claims / 39 witnesses).

### Not yet flipped / deliberately deferred

- The `hvf` default still uses `HvfBackend`'s in-loop gate; the relay is opt-in
  (`--hypervisor inhouse` or `MVM_HVF_EGRESS_RELAY=1`) until a full
  `machine run --hypervisor inhouse` proof and the default flip.
- The legacy in-loop gate (`EgressProxy` / `set_egress_gate`) is shared with the
  KVM boot path (`kvm/vm.rs::boot_with_egress`), which has no relay wiring and
  can't be live-verified on macOS — so its deletion is gated on KVM relay wiring +
  a Linux/KVM live proof.
- libkrun/vz/Firecracker still use their own backends (their `VmmDriver` migration
  and the deletion of the five old backend types are the remaining ADR-102 slices).

Also includes a hardening fix: the builder vsock file-transfer now strips
setuid/setgid/sticky bits from untrusted (guest→host) modes.


auser:
## CI fix: the only failure is a Linux-only unused import

All three red jobs (Lint, Test, MCP roundtrip) cascade from a single error — nothing else is failing:

```
error: unused imports: `Read` and `Write`
  --> crates/mvm-hostd/src/supervisor/raw_egress.rs:182:19
```

It's inside `#[cfg(target_os = "linux")] fn handle_raw_conn_blocking`, which pumps via `std::io::copy` (a free function — no `Read`/`Write` trait import needed in scope) and delegates line-reading to `read_target_line_blocking` (its own `R: Read` bound). So the import is genuinely dead. It only surfaces on the **Linux target**, so macOS-local `clippy` misses it — which is why it slipped through.

**Fix** (drop the import):

```diff
 ) -> std::io::Result<()> {
-    use std::io::{Read, Write};
     use std::os::fd::{FromRawFd, OwnedFd};
```

Verified on the Linux target: `RUSTFLAGS="-D warnings" cargo zigbuild --target x86_64-unknown-linux-gnu --workspace --lib` is clean with this applied (and no other Linux-only lib warnings behind it). It's already committed as `3d38a674` on `feat/plan-214-hvf-workload-runner` if you'd rather cherry-pick.

`raw_egress.rs` is **not** among the files conflicting with `main`, so this applies cleanly — please fold it into the conflict-resolution rebase (the driver-seam `add/add` conflicts against `main`'s landed S0 slice are the separate, bigger part).

PR #1389, opened 2026-07-01
feat(client): GatewayBackend — complete remote MvmClient over mvmd-gateway (Plan 216 S3+S5)

auser:
Stacked on #1387 (S0). Begins slice **S3** — the remote `MvmClient` that makes the facade usable from `mvm-studio` (and `mvmctl --remote`, later). Behind the off-by-default **`remote`** feature, so a local-only build stays light.

## What's here (the security/transport core)
- **Fail-closed endpoint guard** — TLS required off loopback; `http` only to a local sidecar (127.0.0.1 / ::1 / localhost). Enforced at construction *and* per request (so a redirect/misjoin can't downgrade to cleartext). ADR-104 client rule.
- **`GatewayBackend`** — TLS-validating reqwest client, bearer auth **endpoint-bound** to the configured base, **redacting `Debug`** so the token never lands in a log.
- **`stop_machine` / `machine_logs`** — implemented against the gateway sandbox routes (`POST /api/v1/sandboxes/{id}/stop`, `GET …/logs`) with HTTP-status→error mapping.
- **`list_machines` / `run_machine`** — return an honest `pending typed DTO contract` error rather than guessing the gateway's untyped JSON shape; they land with **S5** (gateway adopts the shared typed DTOs). Not silent stubs.

## Verification
17 tests: guard matrix (incl. `[::1]`), guarded construction refuses cleartext-remote, endpoint join preserves the guard, **token-redaction in Debug**, status→error mapping. `clippy -D warnings` clean with the feature **on and off**; default build unchanged and light (reqwest only under `remote`).

## Gate reminder
Per ADR-104 §6, enabling the production **remote-fleet** path (default-reachable `mvmctl --remote <fleet>`) stays gated on ADR-104 Accepted + Plan 57 CT-1 green. This slice is the transport core + the local-sidecar case (the allowed loopback exception); it does not flip that gate on.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2577,6 +2577,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "mvm-client"
+version = "0.16.1"
+dependencies = [
+ "async-trait",
+ "reqwest 0.12.28",
+ "serde",
+ "serde_json",
+ "thiserror 1.0.69",
+ "tokio",
+]
+
+[[package]]
 name = "mvm-core"
 version = "0.16.1"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2581,6 +2581,8 @@ name = "mvm-client"
 version = "0.16.1"
 dependencies = [
  "async-trait",
+ "mvm-backend",
+ "mvm-core",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


auser:
Updated: GatewayBackend is now **complete** — all four `MvmClient` methods (list/run/stop/logs) implemented against the real gateway contract (typed `SandboxRecord` + `ApiResponse`/`PaginatedResponse` envelopes), mapped client-side (no mvmd imports). 22 tests. Studio can drive the full remote path against its sidecar. `run_machine` fails closed on env (the create-sandbox endpoint has no env field). Production remote-fleet enablement remains gated on ADR-104 + Plan 57 CT-1 per ADR-104 §6.

Issue #1388, opened 2026-07-01
Plan 214 machine library: expose public boot + list/persistence seams for mvm-client LocalBackend (Plan 216 S1)


auser:
`mvm-client`'s `LocalBackend` (Plan 216 S1) needs to drive local machine lifecycle in-process. The `mvm::machine` library (`crates/mvm/src/machine/`) already exposes the **construction** seam — `MachineBuilder` → `select_backend()` → `to_start_config() → VmStartConfig` — but not the pieces `LocalBackend` needs to implement `run` / `list`:

**Requested public seams (so LocalBackend consumes the library instead of forking a boot path or duplicating CLI-private types):**

1. **Boot entrypoint.** A public function to launch a `Machine` / `VmStartConfig` into a running VM and return a typed result (id/name/status) — the piece `feat/plan-214-machine-run` is building. Today boot is embedded in `crates/mvm-cli/src/commands/vm/up.rs::start_persistent_oci_machine`, coupled to admission/signing and not exported.
2. **List / persistence.** A public way to enumerate persisted machines with status (typed state), or expose the persisted machine-spec type + a `list` function. Today the persisted `MachineSpec` is `pub(in crate::commands)` in `crates/mvm-cli/src/commands/machine/mod.rs` and enumeration is CLI-internal (`list_machine_specs()` over `mvm_core::config::machine_state_root()`).
3. **Confirmation** that `mvm::machine::{Machine, MachineSpec}` are the canonical types a non-CLI caller (LocalBackend, mvmd, SDKs) should build from.

**Why now:** Plan 216 S1 is scoped to build on this library rather than race it (see `specs/plans/216-mvm-client-facade.md` §S1). We deliberately did NOT fork a boot path. S1b (wiring `run_machine`) unblocks the moment the boot + list seams are public.

Refs: Plan 216 (mvm-client facade), Plan 214 (machine library / machine-run).

auser:
**Update — this is now substantially unblocked; scope narrowed to one seam.**

`LocalBackend` has been built (Plan 216 S1a, PR #1389) against the **already-public** seams on main:
- **list** — `AnyBackend::list() -> Vec<VmInfo>` maps cleanly to the facade's `MachineState` (VmInfo carries id/name/status). Running-VM listing needs no new seam.
- **stop / logs** — `AnyBackend::{stop,logs}` directly.

So requests #2 (list/persistence) and #3 (canonical types) from the original ask are effectively satisfied for a working LocalBackend.

[...]

PR #1394, opened 2026-07-02
feat: converge SDK on the MvmClient facade — extract mvm-client-local + SDK subprocess impl (Plan 218 P0/P1)

auser:
Implements Plan 218 P0+P1 (ADR-105, #1393). The SDK now drives machine lifecycle through the `MvmClient` trait instead of its own bespoke driver.

## P0 — extract `mvm-client-local` (break the dependency cycle)
`mvm-client` declared `mvm-backend` (for the `local` feature). **Cargo detects cycles on the manifest graph, optional deps included** — so `mvm-sdk` (which sits below `mvm-backend` via `mvm-build → mvm-network → mvm-sdk`) could not depend on `mvm-client`:
```
mvm-sdk → mvm-client → mvm-backend → mvm-build → mvm-network → mvm-sdk   (cargo refuses)
```
Fix is structural: **`LocalBackend` moves to a new `mvm-client-local` crate**. `mvm-client` keeps the trait + DTOs + `GatewayBackend` + `connect` + mock and now carries **zero `mvm-*` deps**. `connect(Target::Local)` directs callers to construct `mvm_client_local::LocalBackend` directly. **Studio (remote) is unaffected.**

## P1 — `SubprocessBackend` in the SDK
`mvm-sdk` depends on `mvm-client` and adds `SubprocessBackend`: an `MvmClient` impl that shells `mvmctl machine` (sync `std::process`, mirroring the sibling `machine.rs`). `stop`/`logs` implemented; `list` parses `machine ls --json`; `run` refuses until the admitted-boot seam (#1388) — the SDK never gets a second, admission-skipping boot path. A **`cargo tree` cycle-guard test** fails if `mvm-sdk` ever links `mvm-backend`.

## Verification
Cycle gone (`cargo tree` clean); `cargo check --workspace` clean (1m13s). Tests: mvm-client 26, mvm-client-local 3, mvm-sdk 259 lib + 3 facade + cycle-guard — all green. clippy `-D warnings` + fmt clean.

## Known limitation (documented)
Subprocess `list` maps `machine ls --json` (persisted specs), which carries no live runtime status, so status is reported `Stopped`. A status-aware listing closes it (follow-up).

Depends on ADR-105 / Plan 218 (#1393).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,12 +2606,21 @@ name = "mvm-client"
 version = "0.16.1"
 dependencies = [
  "async-trait",
- "mvm-backend",
- "mvm-core",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
+ "tokio",
+]
+
+[[package]]
+name = "mvm-client-local"
+version = "0.16.1"
+dependencies = [
+ "async-trait",
+ "mvm-backend",
+ "mvm-client",
+ "mvm-core",
  "tokio",
 ]
 
@@ -2793,11 +2802,13 @@ dependencies = [
 name = "mvm-sdk"
 version = "0.16.1"
 dependencies = [
+ "async-trait",
  "base64",
  "flate2",
  "globset",
  "hex",
  "hmac",
+ "mvm-client",
  "mvm-core",
  "mvm-guest",
  "schemars",
@@ -2807,6 +2818,7 @@ dependencies = [
  "tar",
  "tempfile",
  "thiserror 1.0.69",
+ "tokio",
  "toml",
  "tree-sitter",
  "tree-sitter-javascript",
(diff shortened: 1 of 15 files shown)


auser:
Extended: **added `exec_machine` to the `MvmClient` trait** (non-interactive `command -> ExecResult`), so the facade covers what the SDK's `machine.rs` does (enabling P3's retire). **`shell` is deliberately NOT a trait method** — interactive shells need a duplex PTY the request/response trait can't model, and it matches claim-15's dev-only interactive posture; it stays a CLI concern. Impls: SubprocessBackend (real, `mvmctl machine exec`), MockBackend (canned), LocalBackend + GatewayBackend (honest errors — no AnyBackend exec seam; gateway exec is streaming). All green.

PR #1416, opened 2026-07-04
feat(build): materialize_ext4_pure — in-process rootfs materialize (Plan 221 B)

auser:
Connects the `mvm-ext4` writer (#1414) to the real pipeline: `mvm_build::rootfs::materialize_ext4_pure` walks an OCI-unpacked tree and builds the rootfs ext4 **in-process, no builder VM, no `mkfs`, no subprocess**.

## What

- Symlink-aware walk (never follows) → `mvm_ext4::Node` list with guest-absolute paths + unix permission bits. Files/dirs/symlinks captured; other inode types skipped (an OCI rootfs mounts devtmpfs).
- Delegates to `mvm_ext4::build_image` — deterministic + `#![forbid(unsafe_code)]`.
- New `RootfsError` variants for the walk/build/write steps.

## Scope / honesty

Reads the tree + assembles the image in memory — fine for small/medium rootfs; **large images want streaming + multi-block-group** (Plan 221 follow-ups; the writer currently caps at a single 128 MiB block group). The builder-VM `materialize_ext4` stays for Nix builds and as the >128 MiB fallback (ADR-093-style), so nothing regresses.

## Tests

- Materialize a temp tree (`/etc/hosts`, `/hello`, `/etc/localhost` symlink) → valid ext4 (superblock magic `0xEF53`), and **byte-identical across two runs** (determinism).
- Non-directory input is rejected.
- clippy + fmt clean.

## Plan 221 progress

| step | what | status |
|---|---|---|
| B2 | `mvm-ext4` memory-safe writer | ✅ #1414 |
| B4a | `materialize_ext4_pure` (dir → ext4) | **this PR** |
| next | pure-Rust dm-verity roothash; multi-block-group; wire `LocalBackend.run` | — |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,6 +169,11 @@ jobs:
           cargo test -p mvm-build --features builder-vm --lib \
             builder_backend_select::tests::auto_detect_default_for_everything_else_picks_libkrun
 
+      - name: In-process pure-mkfs materialize tests
+        # `pure-mkfs` is off by default (keeps mvm-ext4 out of the mvmctl default
+        # closure), so the default workspace run skips these. Cover them here.
+        run: cargo test -p mvm-build --features pure-mkfs --lib rootfs::tests::pure
+
   test:
     name: Test
     runs-on: ubuntu-latest
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,6 +2571,7 @@ dependencies = [
  "libkrun-sys",
  "lzma-rs",
  "mvm-core",
+ "mvm-ext4",
  "mvm-guest",
  "mvm-network",
  "mvm-sdk",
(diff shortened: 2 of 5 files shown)


auser:
Informational, from Plan 221 **B3** (#1429): the `mvm-ext4` writer now returns three new `Err` variants on hostile trees — `NotADirectory` (a file used as a parent dir), `DuplicatePath` (repeated path / node re-claiming `/`), and a tightened `BadPath` (empty/`.`/`..`/NUL path segments). These were silent `Ok`-with-broken-image bugs before.

No change needed in `materialize_ext4_pure` — the OCI-dir walk will now surface these instead of emitting an unreachable inode, which is the behavior you want. Just flagging so a hostile-tree `Err` from `build_image` isn't a surprise.

PR #1427, opened 2026-07-04
feat(mvm-ext4): multi-block-group images (rootfs past 128 MiB)

auser:
## What

The pure-Rust ext4 writer no longer caps at a single block group's **128 MiB**. It now:

- **Resolves geometry** (`Layout::plan`) from the inode + data-block counts: how many groups, inodes-per-group, GDT/inode-table sizing, and the final (possibly partial) group.
- Lays out a **uniform per-group metadata prefix** — superblock/GDT backup, block bitmap, inode bitmap, inode table — then data.
- **Distributes inodes** across groups (`locate_inode`) with correct per-group inode bitmaps + free counts.
- Allocates file data as **extents** from each group's data region (`RegionAllocator`), so a file crossing a group boundary becomes multiple extents.
- Writes **SB+GDT backups** into every group (no `sparse_super`) for offline-tool consistency; kernel RO mounts trust the primary.

A single file that would fragment past the **four inline extents** an inode holds is rejected with a clear `FileTooFragmented` error (extent-tree interior nodes are a future addition). The common rootfs — many modest files — needs one or a few extents each, so this unlocks arbitrarily large *total* images.

## Validation

- **Oracle test** (`multi_group_image_round_trips_through_real_reader`): a > 128 MiB image with a 130 MiB file mounts in the independent `am-fs-ext4` reader and reads back **byte-exact across the group boundary**.
- **CI** (`ext4-real-mount`): a new `write_multigroup` example + a mount step loop-mounts a multi-group image on the **real Linux kernel** and checks the big file's size + a byte well past the first group.
- Single-group images are unchanged (the layout reduces to the prior fixed one at `groups=1`); all existing oracle + determinism tests stay green. `#![forbid(unsafe_code)]` intact; clippy clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,3 +336,22 @@ jobs:
             exit 1
           fi
           echo "pure-Rust dm-verity root hash + hash tree == veritysetup"
+      - name: Write + loop-mount a MULTI-block-group image (> 128 MiB)
+        run: |
+          set -euo pipefail
+          cargo run -p mvm-ext4 --example write_multigroup -- /tmp/multigroup.ext4
+          # Sanity: bigger than one group's 128 MiB.
+          [ "$(stat -c '%s' /tmp/multigroup.ext4)" -gt 134217728 ]
+          mnt=$(mktemp -d)
+          sudo mount -o ro,loop /tmp/multigroup.ext4 "$mnt"
+          trap 'sudo umount "$mnt" || true; rmdir "$mnt" || true' EXIT
+          [ "$(cat "$mnt/etc/marker")" = "multi-group marker" ]
+          # The big file spans two block groups as two extents — size + a byte
+          # past the group boundary must read back correctly.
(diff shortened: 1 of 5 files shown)


auser:
Heads-up from the Plan 221 **B3** adversarial-test work (#1429, just opened):

#1429 adds `crates/mvm-ext4/tests/adversarial.rs::file_over_one_block_group_is_rejected`, which asserts `build_image` returns `Err(TooLarge)` for a file one block past the 128 MiB single-group cap — correct against `main` today.

**When this PR lands, that assertion will flip** (a 128 MiB+ file should now build `Ok` across multiple groups). Please update that test as part of #1427 — either re-point it at the new multi-group ceiling or convert it to assert successful multi-group layout. It's the only test coupling between the two PRs; I flagged it in an in-test comment referencing #1427.

Merge order doesn't matter as long as whichever lands second reconciles that one assertion.

PR #1429, opened 2026-07-04
test(mvm-ext4): adversarial-tree regression suite + 3 writer fixes (Plan 221 B3)

auser:
Completes the **adversarial-tree** half of Plan 221 **B3** / ADR-106, sibling to the fuzz harness in #1428.

## Why this is not covered by #1428

The merged `fuzz_build_image` target proves *no panic on well-shaped node sets*, but it restricts its inputs by design:

- path segments from a safe alphabet — **never** `/`, NUL, `.`, `..`, or empty
- files capped at **64 KiB** (never near the 128 MiB block-group boundary)
- depth capped at **4**
- symlink targets are single clean tokens (**never loops**)

So the four shapes ADR-106 B3 explicitly names — **deep / huge / symlink-loop / malformed** — all land in code paths the fuzz never reaches. This PR adds a deterministic regression suite (`tests/adversarial.rs`, 7 tests) over exactly those shapes, mounting output through the independent `am-fs-ext4` reader.

## 3 real writer bugs found (red → green)

Writing the tests first surfaced three contract violations — the writer's own doc promises *"a bug is at worst a returned error, never corruption"*, so a silently-broken image is a bug:

| Input | Before | After |
|---|---|---|
| `/a` (file) + `/a/b` | `Ok` with orphaned, unreachable inode | `Err(NotADirectory)` |
| two nodes same path; node at `/` | 2 inodes + duplicate dirents; root inode silently remapped | `Err(DuplicatePath)` |
| `//a`, `/a/./b`, `/a/../b`, NUL segment | `//a` aliased `/a`; dot/NUL passed through | `Err(BadPath)` — `normalize` rejects empty/dot/dotdot/NUL segments |

New `Ext4Error` variants are **additive** (nothing consumes `mvm-ext4` yet). `deep` / `symlink-loop` / `DirTooLarge` / `TooLarge` pass as characterization tests, pinning the contract.

## Verification

`cargo fmt` + `clippy --all-targets` clean; 7 adversarial + 3 oracle + doctests green locally.

## Coordination

- **#1416 (`materialize_ext4_pure` walker)** — the walker will now correctly surface `NotADirectory` / `DuplicatePath` / `BadPath` on a hostile OCI tree instead of emitting a broken image. No change needed there; this hardens its foundation.
- **#1427 (multi-block-group)** — `file_over_one_block_group_is_rejected` asserts `TooLarge` against `main`'s single-group cap. When #1427 lands, that boundary moves; **update that test's expectation as part of #1427** (flagged in an in-test comment). This is the only ordering coupling.

Closes the B3 adversarial-tests line item; B0 (ADR-106) is still a separate draft.

### DIFF
--- a/crates/mvm-ext4/src/lib.rs
+++ b/crates/mvm-ext4/src/lib.rs
@@ -82,6 +82,12 @@ pub enum Ext4Error {
     BadPath(String),
     /// A directory's entries did not fit the single-block assumption.
     DirTooLarge(String),
+    /// A node's parent path resolves to a non-directory (e.g. a regular file
+    /// or symlink), so the node could never be reached on a mounted image.
+    NotADirectory(String),
+    /// Two nodes claim the same path (including a node re-claiming the implicit
+    /// root `/`), which would emit duplicate directory entries.
+    DuplicatePath(String),
 }
 
 impl std::fmt::Display for Ext4Error {
@@ -105,6 +111,10 @@ impl std::fmt::Display for Ext4Error {
             Ext4Error::DirTooLarge(p) => {
                 write!(f, "directory {p} has too many entries for one block")
             }
+            Ext4Error::NotADirectory(p) => {
+                write!(f, "parent of {p} is not a directory")
+            }
+            Ext4Error::DuplicatePath(p) => write!(f, "duplicate path {p}"),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


auser:
Rebased onto `main` after the multi-block-group change merged. Removed the now-obsolete single-group `TooLarge` test (that boundary no longer exists; positive multi-group coverage already lives in `oracle::multi_group_image_round_trips_through_real_reader`). The writer hardening (`NotADirectory`/`DuplicatePath`/`normalize`) re-applied cleanly onto the rewritten `build_image`; its three tests pass unchanged. This was the exact coupling flagged earlier — now reconciled.

PR #1472, opened 2026-07-05
feat(hvf): wire the virtio-fs root device into the in-house boot path (Plan 223 A1b-ii)

auser:
Connects the merged virtio-fs device (FuseServer #1465 + VirtioFs transport #1469) to a real HVF boot. **Dormant until a virtiofs root is set** — every existing config site passes `None`, so zero behavior change on the block-rootfs path.

## What
- `kernel_boot`: `HostChannels.virtiofs_root` / `RunInputs.virtiofs_root`; when set, attach a `VirtioFs` at a dedicated MMIO window (above the disk band + vsock), add its FDT node so the guest probes it, and default the cmdline to `rootfstype=virtiofs root=mvmroot rw init=/init` (no block rootfs).
- `run`: `impl RunDevice for VirtioFs`.
- `HvfSupervisorConfig`: `virtiofs_root: Option<PathBuf>` (serde default); supervisor threads it into `HostChannels`, plus a `MVM_HVF_VIRTIOFS_ROOT` dev hook for live-mount iteration.
- `HvfBackend` capabilities: `virtiofs_root: true` — the A4 tier gate selects it only for non-prod, non-sealed dev workloads.

## Verification
Full chain compiles; 16 hvf + 19 virtio unit tests pass; clippy `-D` / fmt / spec-refs clean. The device path is correct-by-construction (mirrors the live-proven `VirtioBlk` virtqueue/FDT/cmdline handling).

## Next
Live boot on `--hypervisor inhouse` / `MVM_HVF_VIRTIOFS_ROOT` and tune the FUSE wire layout against the guest mount (the one thing unit tests can't witness); then the run-path gate wiring (A4→backend carrying the unpacked tree), A2 `/init`, A3 integrity, A5, + a FUSE-parser fuzz target.

PR #1474, opened 2026-07-06
feat(run): virtiofs-root selection wired into run --image on HVF (Plan 223 A4→run path)

auser:
Stacked on #1472. Threads the virtiofs-root selection from the run path to the in-house backend.

- `VmStartConfig.virtiofs_root: Option<String>` (defaults `None`; every site spreads `Default`, zero behavior change).
- `HvfBackend::start`: when set, attach **no block rootfs** and pass the directory to the supervisor's `virtiofs_root` (served read-only over virtio-fs).

Backend-side plumbing only — any `VmStartConfig` producer can now request a virtiofs-root boot; the mechanism is the one already **live-validated on HVF** (#1472). The run-path tier gate that *sets* the field (`select_root_strategy` → inject-only tree → `VmStartConfig`) is the next slice.

clippy `--all-targets -D` / fmt / spec-refs / core-runtime-free clean.

### DIFF
--- a/crates/mvm-backend/src/hvf_backend.rs
+++ b/crates/mvm-backend/src/hvf_backend.rs
@@ -218,18 +218,26 @@ impl VmBackend for HvfBackend {
             .map(PathBuf::from)
             .unwrap_or_else(|| state_dir.join("hvf-agent.sock"));
 
+        // virtiofs-root dev boot (Plan-223 tier gate): serve the unpacked+injected
+        // tree read-only over virtio-fs; no block rootfs is attached.
+        let virtiofs_root = config.virtiofs_root.clone().map(PathBuf::from);
         // Single workload rootfs → one ephemeral (RAM-backed) writable disk: the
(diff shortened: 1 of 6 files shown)


auser:
Expanded to the **full run-path wiring** (two commits): the `VmStartConfig.virtiofs_root` plumbing **plus** the CLI tier gate.

`mvmctl run --image` on a virtiofs-capable backend now boots a **dev-tier** OCI image straight from the unpacked tree over virtio-fs (no ext4), while **prod, sealed, and block backends stay on Option B** (claim 3). `resolve_virtiofs_root` wraps the exhaustively-tested `select_root_strategy` as the single authority; a safety test pins that prod / sealed / non-virtiofs-backend / non-OCI-source all fall back to the block rootfs — "prod on virtiofs" is unrepresentable end to end.

Combined with the boot path (#1472), this makes the whole Plan-223 Option A mechanism reachable from a real `mvmctl run --image --hypervisor inhouse` — the mechanism already **live-booted on HVF**.

Issue #1478, opened 2026-07-06
checkpoint/fork is Vz-coupled — untracked gap in the Vz deletion; re-home onto HVF + Firecracker


auser:
## (a) The gap

`mvmctl machine checkpoint create/fork` (checkpoint-lineage / sandbox-fork) currently **only works on the Vz backend**, which is being deleted (HVF is the macOS backend; we should never depend on Vz). Confirmed live on the Linux/KVM/Firecracker box:
- `checkpoint create --class vm-full` → `Error: vm save requires Vz save/restore support, but backend 'vz' reports snapshot tier 'unsupported'`.
- `checkpoint create --class fs-quick` (after `machine pause`) → `Error: fs_quick checkpoint is not supported for this VM's backend`.

**This is an untracked gap in the Vz-deletion / HVF-migration effort:**
- **ADR-098** (raw-HVF as the macOS snapshot/warm-pool backend) covers **warm-pool / warm-restore** (`vm resume --warm`, Plan 175/206, FC-native `/snapshot/load`) — a *different* feature. It has **zero** mention of `machine checkpoint`/`fork`, `CheckpointClass::VmFull`/`FsQuick`, or fork lineage.
- The 2026-07-02 Vz-deprecation design note names deletion targets (`vz.rs`, `vz_control.rs`, `mvm-vz-supervisor`, the `HvfBackend` duplicate) but **does not account for checkpoint's Vz-dependent code**.
- No plan owns re-homing checkpoint/fork onto HVF or Firecracker.

**Consequence:** when Vz is deleted, `machine checkpoint`/`fork` loses its backend entirely. And the verb-grant restore-reconciliation (#1452 Task 5 — mint-at-fork, `PostRestore { grant_envelope }`, `re_pin_verb_grant`, `capture_vm_full` sidecar propagation) rides this transport, so its live-validation (#1462) is blocked on this too. The reconciliation code is correct at the grant/plan layer (unit-tested + reviewed); only the underlying capture/fork transport is Vz-bound.

## Exact coupling points
[...]

Issue #1462, opened 2026-07-05
verb-grant: live-validate fork/restore grant delivery (fs_quick + vm_full + pause/resume)


auser:
Follow-up from ADR-108 / #1381 item 3 restore reconciliation (#1452 Task 5). The code is unit-tested + opus whole-branch-reviewed (which caught + fixed the `capture_vm_full` sidecar-propagation no-op), but the end-to-end **fork re-pin has not been live-booted**.

**What to verify on Firecracker/KVM:**
1. **fs_quick fork:** fork a grant-bearing VM → the child's `<vm_state_dir>/verb-grant.json` is minted at fork admission and delivered on the child cmdline; the restarting child agent re-pins it (child enforces the grant).
2. **vm_full fork:** fork a grant-bearing VM → child gets a fresh grant over the extended `PostRestore { grant_envelope }` frame; the surviving agent re-pins (the `capture_vm_full` sidecar fix means `image_is_sealed`/mint now fires). Confirm the child grant matches the child's own plan nonce (not the stale parent grant).
3. **pause/resume + warm-restore:** the pinned grant survives in guest RAM (no re-pin needed).

**Setup notes (established live 2026-07-05):**
- A grant-bearing VM without a heavy sealed build: `mvmctl machine run --image alpine --name base --agent-verb ping -d` mints a real signed grant (verified: `~/.mvm/vms/base/verb-grant.json` written with `session_id=base`). `machine pause base` works.
- **Blocker to resolve first:** the checkpoint/fork CLI verb path — `mvmctl vm checkpoint create|fork` is rejected by clap on current main (the Plan-208 machine consolidation moved/renamed these; find the real verb, likely under `machine` or a hidden alias), and confirm fs_quick/vm_full forks are supported on the Firecracker backend (the code bails `fs_quick checkpoint is not supported for this VM's backend` for some backends).

Non-blocking for the merged work (restore code is unit+review covered); this closes the "must live-verify" item from the #1452 whole-branch review.

Refs: `crates/mvm-cli/src/commands/vm/checkpoint.rs`, #1452, ADR-108 §3.

auser:
**Progress (2026-07-05): CLI blocker resolved; live vm_full fork cycle hit box flakiness — leaving a clean repro.**

**Resolved — the checkpoint CLI verb path** (the original blocker): the Plan-208 machine-consolidation hid `vm` under `machine`. The real verbs are:
[...]

PR #1481, opened 2026-07-06
feat(checkpoint): de-Vz fs_quick — resolve rootfs backend-neutrally (works on Firecracker/HVF) (#1478 step 1)

auser:
Step 1 of #1478 — decouple the `fs_quick` checkpoint path from Vz, since **Vz is being deprecated/removed** (HVF is the macOS backend; nothing should depend on Vz).

## Problem
`machine checkpoint create <vm> --class fs-quick` only worked on Vz: `resolve_quiesced_vm_rootfs` resolved the VM's rootfs **only** from the persisted Vz supervisor config, so on Firecracker (and HVF) it errored `fs_quick checkpoint is not supported for this VM's backend`. fs_quick is a CoW-rootfs clone with no memory snapshot — it should be backend-agnostic.

## Change
Record the rootfs path in a **backend-neutral** place every backend already writes:
- `VmRuntimeMeta` (`mode.json`) gains `rootfs_path: Option<String>` (`#[serde(default)]`, back-compat — old files read as `None`).
- `record_from_rootfs` (the cross-backend start helper, which already receives the rootfs) persists it. Firecracker already calls this at start, so FC now records its rootfs.
- `resolve_quiesced_vm_rootfs` reads `mode.json`'s `rootfs_path` (backend-neutral) **before** the Vz supervisor-config fallback.
- Removed the Vz-family-only guard in `boot_forked_child` (the instance-rootfs adoption seam is backend-agnostic); added `fc.pid` to the liveness probe so Firecracker VMs are correctly detected as running vs. quiesced.

## ⚠️ Vz-removal note
This keeps `vz_rootfs_from_supervisor_config` as a **transitional fallback** for VMs booted before the `rootfs_path` field existed. It is **dead-code-in-waiting** and should be **deleted together with `vz.rs`** when the Vz removal lands — it is not new Vz coupling. Flagged on #1478 so it's removed as part of the Vz deletion.

## Testing
Unit: `VmRuntimeMeta` serde round-trip incl. `rootfs_path` + legacy-shape `None` default; `record_from_rootfs` persists it; `resolve_quiesced_vm_rootfs` resolves from `mode.json`. `cargo check --workspace`, `nextest -p mvm-backend -p mvm-cli`, `fmt --all`, `clippy -D warnings` clean.

**Live-validated on Firecracker/KVM:** boot a VM, `machine pause`, `machine checkpoint create --class fs-quick` → **succeeds** (`ckpt id=...`; was "not supported" before), `machine checkpoint fork --boot` → **child boots**. `mode.json` recorded `rootfs_path` on FC. So checkpoint/fork now runs with **zero Vz dependency**.

Scope: this PR is the fs_quick rootfs-resolution de-Vz only. The verb-grant re-pin on a *sealed* fork (#1452 Task 5) is unit-tested; live-exercising it needs a sealed-image fixture (the FC test used an unsealed OCI image, whose fork child correctly gets no default grant). vm_full de-Vz (a `VmBackend` capture trait) + HVF snapshot remain per #1478 steps 2–3.

Refs: #1478, #1462.

### DIFF
--- a/crates/mvm-backend/src/base/runtime_meta.rs
+++ b/crates/mvm-backend/src/base/runtime_meta.rs
@@ -52,6 +52,16 @@ pub struct VmRuntimeMeta {
     /// (VMs predating the flag were all accessible).
     #[serde(default = "default_accessible")]
     pub accessible: bool,
+
+    /// Absolute path to the rootfs image the VM was started from.
+    ///
+    /// Written at start time by every backend that calls
+    /// `record_from_rootfs`. Absent in older mode.json files (reads
+    /// as `None`). Used by the fs_quick checkpoint path to resolve the
+    /// rootfs backend-neutrally without requiring a backend-specific
+    /// supervisor config.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rootfs_path: Option<String>,
 }
 
 fn default_accessible() -> bool {
@@ -152,6 +162,7 @@ pub fn dev_attached(mode: StartMode) -> VmRuntimeMeta {
     VmRuntimeMeta {
         mode: mode.into(),
         accessible: true,
+        rootfs_path: None,
     }
 }
 
@@ -168,6 +179,7 @@ pub fn from_sidecar(mode: StartMode, rootfs_dir: &std::path::Path) -> Result<VmR
     Ok(VmRuntimeMeta {
         mode: mode.into(),
         accessible,
+        rootfs_path: None,
     })
 }
 
@@ -185,7 +197,8 @@ pub fn from_sidecar(mode: StartMode, rootfs_dir: &std::path::Path) -> Result<VmR
 /// step is best-effort and only logs warnings.
 pub fn record_from_rootfs(name: &str, mode: StartMode, rootfs: &std::path::Path) -> Result<()> {
     let dir = rootfs.parent().unwrap_or_else(|| std::path::Path::new("."));
-    let meta = from_sidecar(mode, dir)?;
+    let mut meta = from_sidecar(mode, dir)?;
+    meta.rootfs_path = Some(rootfs.to_string_lossy().into_owned());
     write(name, &meta)
 }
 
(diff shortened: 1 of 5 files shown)


auser:
**Reviewed (opus whole-diff) + two fixes applied, all live-validated on Firecracker.**

The review confirmed the load-bearing parts are sound (the `boot_forked_child` Vz-guard removal is provably safe — the no-clobber seam is path-equality-based and backend-agnostic; back-compat + resolution ordering OK) but found one **Important** correctness bug, now fixed in two steps:

1. **`fc.pid` resolved at the wrong dir** (`85caf32e`): the liveness probe read `fc.pid` from `vm_state_dir`, but Firecracker writes it to `~/microvm/vms/<name>/fc.pid` — so a **live** FC VM read as quiesced and could be fs_quick-checkpointed mid-write. Now resolves via `mvm_backend::microvm::fc_pid_path`.
2. **`fc.paused` marker for paused-vs-running** (`8c694556`): FC `machine pause` snapshot-seals but **leaves the fc process alive**, so pid-liveness alone can't tell paused from running (the first fix would have refused *paused* VMs too). Mirrors Vz's proven `vz.paused` scheme: `pause` stamps the live fc pid into `vm_state_dir/fc.paused`, `resume` (plain + `--warm`) removes it, and `vm_is_quiesced` treats "marker matches live fc pid" as quiesced. The removal-on-resume is essential — FC keeps the same pid across pause/resume (unlike Vz's self-invalidating supervisor pid).

**Live-validated (Firecracker), all green:**
- Running FC VM → `fs_quick checkpoint` **refused** (`stop or pause…`).
- `machine pause` → `fc.paused` **written**; paused FC VM → checkpoint **succeeds** (`ckpt-…`) + `fork --boot` → **child boots**.
- `machine resume` → `fc.paused` **removed** → checkpoint **refused again** (no stale-marker bug).

Gates: `cargo check --workspace`, `nextest`, `fmt --all`, `clippy -D warnings` clean. Ready.

PR #1423, opened 2026-07-04
feat(backend): default macOS-26 workloads to the in-house HVF VMM (was Vz)

auser:
> **Draft — NOT for merge yet.** Blocked on the in-house workload-path fixes (below). Opened to stage + review the default flip. Fully built/tested/linted.

## Summary

Flips the macOS-26 Apple-Silicon **workload** VMM default from Apple Virtualization.framework (Vz) → mvm's **in-house HVF VMM**, now that the in-house workload runner is live-proven:
- host↔guest agent reachability (#1406), and
- a `WorkloadRunner<InHouseDriver>` egress proof this session: `start → status → logs → stop` with **claim-10 enforced** (`WorkloadRunner: admitted destination reachable: true`).

`--hypervisor vz` / `MVM_HYPERVISOR=vz` remains the explicit escape hatch.

## The default lives in several coupled sites (all flipped together)

- **`AnyBackend::auto_select`** (exec/build/doctor/mvm-client-local) → refactored to a pure `auto_select_from(native, macos_tier, has_libkrun)` (unit-testable without a real platform); macOS-26 → `Hvf`.
- **`resolve_effective_hypervisor`** (the CLI run/pool default — the actual `machine run` lever) → refactored to a pure `platform_default_hypervisor(...)`; macOS-26 → `"hvf"`.
- **`down` / `ps`** markerless-VM platform fallback → in-house (correctness already holds via the registered `hvf.pid` marker; this is the display edge).

## Why `Hvf`/`"hvf"`, not the `InHouse` runner selector

`"hvf"` is already wired into **every** hypervisor-name string classifier — critically `persists_plan_before_start` (up.rs), which gates persisting `plan.json` that the **claim-10 substitution endpoint** reads. `"inhouse"` is *not* in those sets, so routing the default through the runner selector would silently skip plan-persist and break claim-10. Same in-house VMM either way (`kind() == Hvf`); the `hvf → InHouse` runner migration (wiring `"inhouse"` into all classifiers) is a separate slice.

## Bonus fix the flip forced

`HvfBackend` had no `security_profile()` — it fell to the `Unknown`-tier default, which `doctor` now reports as the default backend. Added one: **Tier 2**, matching its Vz sibling's macOS HVF posture (claims 1–7, claim 3 verified-boot partial), with in-house/vsock-only-egress notes.

## Validation

- ✅ In-house **builder** builds live (nix build → artifacts).
[...]

PR #1491, opened 2026-07-06
fix(vmm): reset guest side of host-closed agent streams — HVF machine run works end-to-end (Plan 228 WS-1)

auser:
**Clears the one functional gate for the 0.17.0 release** (Plan 228 WS-1 / Plan 227 WS-A). Stacked on #1490 (the inhouse→hvf rename); base auto-retargets to main once that merges. Only 2 files.

## The bug
On the macOS-26 `hvf` default, a non-interactive `machine run --image alpine -- <cmd>` hung: the guest agent received the host's `ProtocolHello` (host→guest delivery was instrumented and proven earlier) but **never replied**, so `wait_for_agent` blocked. Firecracker worked with the same agent, isolating it to the in-house vsock device.

## Root cause
`agent_bridge::close()` removed a host-dropped stream from its own map but **never told the guest**. The `for_vm` reachability poll connects-then-drops a probe stream every iteration, so each one left a half-open connection on the guest side. They accumulated and wedged the guest agent's ability to reply.

## Fix
`drain_host` now records host-closed conn ids; `drain_agent` sends each an **OP_RST** to the guest so it frees its half. Unit test (`host_close_is_surfaced_for_guest_reset`) asserts the id is surfaced exactly once and the conn is removed.

## Verified live on macOS 26 (hvf default, no flag)
- `machine run --image alpine -- sh -c 'echo ok'` → `ok`, rc=0, across 3 repeated runs.
- `-it` interactive shell → works.

Resolves the in-house agent-reachability blocker that's been open since the #1423 era.


auser:
Superseded by #1489, which landed the same fix (reset the guest side of a host-dropped agent stream) with a cleaner `drain_host -> (ready, closed)` shape. The macOS-26 hvf agent-reachability fix is on main — verified the same root cause (un-RST'd host-dropped `for_vm` probe streams wedging the guest agent). Closing this duplicate.

PR #1513, opened 2026-07-07
fix(mvm-core): MachineSpec doc test import

auser:
Fix doc test import in MachineSpec - was using mvm_client::MachineSpec which does not work because mvm-core can not depend on mvm-client. Changed to super::MachineSpec.


auser:
Superseded by #1507, which now includes the correct mvm-core doctest import fix and has a green CI run. Closing this stacked fix PR to avoid keeping a red duplicate open.

PR #1524, opened 2026-07-07
fix(machine): auto-build per-VM helper bins from a source checkout + ship hvf supervisor

auser:
## Problem

`cargo run -- machine run --image alpine -- ps aux` fails on a fresh checkout:

```
Caused by:
    mvm-hvf-supervisor binary not found (looked at $MVM_HVF_SUPERVISOR_PATH and alongside the current exe)
```

The per-VM helper binaries mvmctl spawns at runtime (`mvm-hvf-supervisor`, `mvm-substitution-endpoint`, `mvm-libkrun-supervisor`, …) are **separate bin targets in other crates**. `cargo run` builds only `mvmctl`, never them, so each backend's `env → sibling → target/` resolver ladder hard-fails on the first missing helper — and fixing one just exposes the next.

Two distinct bugs:
1. **Dev:** the command can't run from a source checkout without manually building each helper.
2. **Prod:** `release.yml` never shipped `mvm-hvf-supervisor` at all — so a **downloaded `mvmctl` on macOS-26 Apple Silicon, where HVF is the auto-detect default, can't boot any machine.**

## Fix

- Add a shared `AuxBin { name, package, extra_build_args }` with `build_from_checkout()` in `supervisor_stale`: when a helper is missing **and** mvmctl runs from a source checkout (`exe` under `<root>/target/<profile>/` with a `Cargo.toml` at the root), materialize it with the same `cargo build` a stale-hint would name. Gated to a checkout, so an installed mvmctl **never** shells out to cargo. Safe to call from the running mvmctl — cargo releases the workspace build lock before the `run` phase, so the nested build can't deadlock the outer `cargo run` (verified empirically).
- Route the **hvf**, **libkrun**, and **substitution-endpoint** resolvers through it (each: one const + one `if let` before their existing bail). **vz** already had an equivalent (`ensure_source_checkout_vz_helpers`) and is untouched.
- Ship `mvm-hvf-supervisor` in `release.yml` (build + package alongside mvmctl).

## Verification

- `machine run --image alpine -- ps aux` boots a guest and runs the workload in **both debug and release** from a fresh checkout, auto-building the helper on first launch.
- Unit test covers source-checkout detection (under `target/` with a manifest → build; installed layout / no manifest → `None`, so no cargo shell-out).
- `cargo fmt` (nightly), `clippy -p mvm-backend -p mvm-cli --all-targets -D warnings`, resolver + `supervisor_stale` unit tests, and `check-no-spec-refs-in-comments` all green.

## Follow-up (not in this PR)

`AuxBin` handles the **missing** case; vz's mechanism also rebuilds a **stale** helper (input-mtime based). Unifying the two — and giving hvf/libkrun stale-rebuild too — is a clean follow-up, left out here to avoid churning vz's working, un-live-testable path.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,6 +123,9 @@ jobs:
       #   - mvm-bridge: cross-platform external-VMM gateway/audit sidecar
       #     (Firecracker passt + vz VzIngest). Builds everywhere, no extra deps.
       #   - mvm-vz-supervisor (macOS only): the vz VMM supervisor.
+      #   - mvm-hvf-supervisor (macOS only): the HVF VMM supervisor. HVF is the
+      #     auto-detect default on macOS-26 Apple Silicon, so without this a
+      #     downloaded install on that tier can't boot any machine.
       #   - mvm-libkrun-supervisor (macOS only, and INTENTIONALLY so — not a
       #     follow-up): it has `required-features = ["libkrun-sys"]`, and the
       #     released Linux mvmctl deliberately does NOT link libkrun — the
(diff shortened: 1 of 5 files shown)


auser:
Superseded — `main` already carries the same fix via `crates/mvm-backend/src/aux_bin.rs` (shared `AuxBin` + `resolve_or_build`: env → sibling → target → build-from-checkout), with the hvf and substitution-endpoint resolvers already wired to it, and `release.yml` already shipping `mvm-hvf-supervisor` (plus `mvm-substitution-endpoint`). Closing as redundant.

PR #1539, opened 2026-07-08
ci(release): new Sigstore bundle format for image-manifest signatures (dev-image verify was broken)

auser:
Follow-up from the investigation prompted by #1537. The cosign round-trip smoke revealed that `cosign sign-blob --bundle` emits the legacy bundle format, which the Rust `sigstore-verify` v0.9.0 stack rejects (`missing field verificationMaterial`). That stack is **not** only used by the new attested packs — it also backs the **dev-image download signature check**.

## Confirmed broken

- `mvmctl` dev-image download calls `image_verify::verify_manifest` (`crates/mvm-cli/src/commands/env/dev_vz.rs:1894`) as a **hard error** — the only bypass is `MVM_SKIP_COSIGN_VERIFY=1`.
- `verify_manifest` → `verify_cosign_bundle` uses `sigstore-verify` **0.9.0** (confirmed in `Cargo.lock`), which requires the new Sigstore bundle format.
- The release signs image manifests with **legacy** `cosign sign-blob --bundle` (the `Sign image manifests` step).

So a downloaded dev image's manifest bundle cannot be parsed by the verifier → the signature check hard-fails (short of the emergency skip). Same root cause and same fix as the builder pack (#1537): `--new-bundle-format`.

## Scope — only the Rust-verified path

- **Fixed:** the image-manifest signing step (`--new-bundle-format`).
- **Left as-is (correct):** the release **tarball/SBOM** signatures. Those are verified by the **`cosign verify-blob` CLI** in `mvmctl update` (`crates/mvm-cli/src/update.rs`), which accepts the legacy format and is non-fatal — changing them to the new format could break that CLI verify.

## Validation

Validated **transitively** by the #1537 smoke: `image_verify::verify_manifest`'s signature step is the *same* `verify_cosign_bundle` primitive the smoke exercised, and that smoke went green with `--new-bundle-format` against `sigstore-verify` 0.9.0. The bundle format is independent of the signed payload, so a green pack round-trip proves the image-manifest bundle will parse too. (Happy to extend the smoke to round-trip a `SignedManifest` directly if you want belt-and-suspenders.)

## Also flagged (future gotcha, not a current break)

The signed **revocation list** would be fetched + verified through the same Rust stack (`try_fetch_revocation_list`, `dev_vz.rs`). There is **no** revocation-signing automation in CI today (no `revocations.yml` workflow, no signing script), so nothing is broken now — but whenever a revocation list is signed and published, it must be signed with `--new-bundle-format` or the consumer-side verify will reject it for the same reason.

### DIFF
--- a/.github/workflows/pack-signing-smoke.yml
+++ b/.github/workflows/pack-signing-smoke.yml
@@ -75,3 +75,35 @@ jobs:
           cargo run --release -p mvm-core --example verify-pack-signature --features manifest-verify -- \
             --manifest smoke/pack-manifest.json --bundle smoke/pack-manifest.json.bundle \
             --identity "${IDENTITY}" --issuer "https://token.actions.githubusercontent.com"
+
+      - name: Round-trip an image manifest through the dev-image verifier
+        # The same Rust stack backs the dev-image download signature check
+        # (image_verify::verify_manifest), which hard-fails on a bundle it
+        # cannot parse. Sign a real image manifest and verify it through that
+        # exact entrypoint, so this smoke also guards the dev-image trust path.
+        run: |
+          set -euo pipefail
+          cat > smoke/image-manifest.json <<'JSON'
+          {
+            "schema_version": 1,
+            "version": "0.0.0-smoke",
+            "arch": "x86_64",
+            "variant": "dev",
+            "rootfs_format": "ext4",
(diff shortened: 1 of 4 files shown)


auser:
Folding into #1537 — the image-manifest fix and the extended smoke that validates it are one logical change (cosign --new-bundle-format for every Rust-verified bundle), and the smoke can only go green with both signing fixes present. #1537 now carries the pack + image release fixes and a smoke that round-trips both.

PR #1554, opened 2026-07-08
feat(release): publish + verify a signed pack revocation list (Plan 213 §I producer)

auser:
Fourth slice of the revocation workstream (ADR-097 §I): the **producer** — publish a signed pack revocation list, so the fetch/apply consumer (#1552) has something to consume. Completes the revocation vertical.

## What

- **`packaging/release/pack-revocations.source.json`** — the human-curated source (starts empty). Maintainers add `RevokedPack` entries here.
- **`.github/workflows/revocations.yml`** — triggered by force-updating the `revocations` tag: stamps a validity window (`issued_at` / `not_after`, `VALID_DAYS=30`), keyless-signs the list with `cosign --new-bundle-format` under the **dedicated `revocations.yml@refs/tags/revocations` identity** (so a leaked release-signing cert can't fabricate a permissive list — matching `release_trust::revocation_keyless_trust()`), **self-verifies** the signed list through the Rust entrypoint before publishing (refuses to ship a list a released binary couldn't verify), and uploads `pack-revocations.json` + `.bundle` to the `revocations` release.
- **`verify-pack-revocation-list`** mvm-core example — calls `pack_revocation::verify_pack_revocation_list` directly (the fetch-path entrypoint).
- **`pack-signing-smoke.yml`** — now also round-trips a real signed revocation list, so the smoke guards all three keyless channels (pack, image manifest, revocation list).

## Validation

Local: the example builds; the jq-stamped source is a valid `PackRevocationList`; a garbage bundle is rejected (`exit=1`); all workflows + the source JSON parse; clippy + actionlint clean. **Live:** I dispatched `pack-signing-smoke.yml` on this branch to prove the revocation round-trip end-to-end with a real cosign bundle (result linked in a comment).

## The vertical is now complete

Core (#1545) → signer binding (#1543) → fetch/apply (#1552) → producer (this). Publishing a revocation = edit the source list, force-update the `revocations` tag. The consumer fetches, keyless-verifies, offline-caches, and unions it with the on-disk config.

### DIFF
--- a/.github/workflows/pack-signing-smoke.yml
+++ b/.github/workflows/pack-signing-smoke.yml
@@ -111,3 +111,23 @@ jobs:
           cargo run --release -p mvm-core --example verify-signed-manifest --features manifest-verify -- \
             --manifest smoke/image-manifest.json --bundle smoke/image-manifest.json.bundle \
             --identity "${IDENTITY}" --issuer "https://token.actions.githubusercontent.com"
+
+      - name: Round-trip a pack revocation list through its verifier
+        # The revocation list is signed + verified through the same Rust stack.
+        # Sign a real list and verify it through the fetch-path entrypoint
+        # (pack_revocation::verify_pack_revocation_list), so this smoke also
+        # guards the revocation channel's signing format.
+        run: |
+          set -euo pipefail
+          now="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+          cat > smoke/pack-revocations.json <<JSON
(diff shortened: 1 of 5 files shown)


auser:
Live proof: dispatched `pack-signing-smoke.yml` on this branch — **green** (run 28923180303). The smoke now round-trips all three keyless channels through the real Rust verifier with real OIDC cosign bundles:

```
pack signature verified
image manifest signature verified
pack revocation list signature verified
```

The revocation-list round-trip (produce → `cosign sign-blob --new-bundle-format` → `pack_revocation::verify_pack_revocation_list`) is the new one this PR adds.

PR #1590, opened 2026-07-09
fix: unblock in-process HVF boots through the MvmClient facade

aneyzberg:
## Summary

Three fixes that make `LocalBackend::run_machine` boot a real HVF microVM in-process. Found by driving the facade end-to-end from **mvm-studio** (its ADR-0003 real-boot smoke test); each was hit, fixed, and re-verified empirically. With these, a UI click boots a live Linux 6.12.87 microVM in ~200 ms.

### 1. musl guest cross-compile broken (`mvm-oci-init`)
`libc::ioctl` request flags were cast to `c_ulong`, but libc's Ioctl type is `c_int` on musl targets — every source-checkout guest-agent build failed with E0308 inside cargo-zigbuild. `as _` defers to whichever type the target's libc declares.

### 2. HVF boots had no kernel (`LocalBackend`)
`run_machine` passed `kernel_path: None` (correct for libkrun/Vz/mock, which carry their own kernel), but HVF — the macOS-26 default backend — requires an explicit arm64 kernel Image. Resolve it from the per-arch workload-kernel cache the CLI populates (`~/.cache/mvm/builder-vm/<arch>/kernels/workload/vmlinux`), cache-hit-or-error with an actionable message. The heavier build/fetch flows stay CLI concerns.

### 3. Error chains flattened (`LocalBackend`)
Two `map_err(backend_err)` sites reduced anyhow chains to their top context — callers saw `"build guest agent binaries from the source checkout"` with no hint of the E0308 underneath, and `"backend start after signed-plan admission"` with no hint of the missing kernel. `{:#}` formatting keeps the chain in `MvmError::Backend`.

## Test plan

- `cargo check -p mvm-client` green on this branch.
- Empirical, on an Apple-Silicon macOS-26 host: `mvmctl machine run --image public.ecr.aws/docker/library/alpine:latest -- echo ok` succeeds (guest binaries cross-compile after fix 1); mvm-studio's facade smoke then boots/lists/logs/stops/removes a live HVF microVM (fixes 2–3), and the same flow works interactively from the studio UI.

## Related gaps observed (not addressed here, happy to file issues)

- `docker.io` refs fail in mvm-oci: the registry host is resolved verbatim and `https://docker.io/v2/...` 302s to a marketing page — needs the standard `docker.io → registry-1.docker.io` alias.
- The facade's registry pull doesn't reuse the OCI cache (repeat runs re-pull; anonymous rate limits bite).
- `resolve_or_build_guest_binaries` inherits `RUSTUP_TOOLCHAIN`/`CARGO*` when the caller was itself launched by cargo (e.g. `cargo test`/`cargo tauri dev`), pinning the nested build to a toolchain that may lack the musl target — the child env should be scrubbed.

### DIFF
--- a/crates/mvm-client/src/local.rs
+++ b/crates/mvm-client/src/local.rs
@@ -143,8 +143,10 @@ fn materialize_from_dir(dir: &Path, name: &str) -> Result<PathBuf> {
     let cache_root = PathBuf::from(mvm_core::config::mvm_cache_dir());
     // The library carries no embedded guest binaries, so it resolves them from
     // the cache or a source checkout.
+    // `{:#}` keeps the anyhow context chain — the top context alone
+    // ("build guest agent binaries…") hides the actionable root cause.
     mvm_build::run_image::inject_and_materialize(&cache_root, dir, &output, name, None, None)
-        .map_err(backend_err)?;
+        .map_err(|e| backend_err(format!("{e:#}")))?;
     Ok(output)
 }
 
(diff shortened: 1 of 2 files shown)


aneyzberg:
Pushed a 4th fix found while booting an unpacked OCI dir through the facade: `LocalBackend::materialize_from_dir` passed the bare cache dir as `cache_root`, but the guest-agent cache lives under the CLI's `oci_cache_root()` (`<cache>/oci/guest-agent/…`) — so the facade always missed a CLI-seeded cache and fell into the source cross-compile. Aligned the segment; the facade now injects from cache (verified: 673MB Debian tree boots in ~36s).

PR #1618, opened 2026-07-10
Compile native per-VM host helpers at build time, not run time

auser:
## Problem

On a source checkout, `cargo run -- machine run …` builds only `mvmctl` — cargo never builds sibling `[[bin]]`s in other crates. The per-VM host helpers (`mvm-hvf-supervisor`, `mvm-libkrun-supervisor`, `mvm-substitution-endpoint`) were therefore compiled lazily by `aux_bin.rs` shelling out to `cargo build` from *inside* the already-running `mvmctl`, mid-`machine run`:

```
Running target/debug/mvmctl machine run …
Error: Stage 0 … (retry)
   Compiling mvm-vm-host …      # compiling a binary while the command that needs it runs
```

Compiling during command execution is the antipattern. All compilation must finish before the command runs.

## Approach

`mvm-cli`'s build script already cross-compiles + embeds the musl host-vm bins via a nested `cargo` aimed at a dedicated target dir under `OUT_DIR` (to dodge cargo's build-lock deadlock). This reuses that proven machinery for the native helpers:

- **build.rs** compiles the host-relevant helpers into `OUT_DIR/aux-helper-target/<profile>/` and exports the dir via `cargo:rustc-env=MVM_AUX_BIN_DIR`. Because build.rs runs in the compile phase, the helpers finish before `mvmctl` executes. Host-conditional and fail-open: `mvm-libkrun-supervisor` only where libkrun is installed, `mvm-hvf-supervisor` only on macOS/aarch64, and a helper the host can't build never fails the outer compile. `MVM_SKIP_EMBED_BINARIES=1` skips it for the fast test loop.
- **mvm-cli** bridges the baked `MVM_AUX_BIN_DIR` into the process env at startup.
- **aux_bin.rs** becomes a pure resolver — no run-time `cargo`. Precedence: `$<ENV_VAR>` override → `$MVM_AUX_BIN_DIR` → exe-sibling → `target/{release,debug}`.

Scope is the native host helpers only; the guest and embedded-host-vm cross-compile paths are untouched.

## Security

Neutral-to-better. Supervisors are host-side code (host is trusted), so no trust boundary moves and no claim witness changes. It deletes a run-time `Command::new(cargo)` from the workload-launch path, so `machine run` no longer needs `cargo` on `PATH` and can't be steered into a compile by a hijacked `PATH`/`CARGO`. `MVM_AUX_BIN_DIR` precedes the exe-sibling and build.rs watches the supervisors' security-relevant sources, so the running `mvm-libkrun-supervisor` (claim-10 egress enforcement + audit-chain sidecar) can't silently lag its source.

## DX tradeoff

Every cold `cargo build`/`cargo run` now pays the host-relevant helper compile (build.rs can't know the subcommand), not just the first `machine run`. Mitigated by `MVM_SKIP_EMBED_BINARIES=1`, `rerun-if-changed` (warm builds ~free), and host-conditional selection. The compile is silent under a normal `cargo run` (cargo hides build-script output).

## Testing

- New unit tests: helper selection (host-conditional, skip, fail-open), resolver precedence (override → aux → sibling → target, absent-skip, missing-helper error), and the env bridge.
- Live-verified on macOS/aarch64: helper compiles land **before** `Running …/mvmctl`, never mid-command; a warm re-run compiles nothing.
- Full gate green (`fmt`, `clippy --workspace -D warnings`, `build --all-targets`, doctests; `nextest --workspace` = 7019 pass / 9 skip, one unrelated network test excepted).

Design + plan: `specs/notes/buildtime-aux-helper-compilation-design.md`, `specs/plans/242-buildtime-aux-helper-compilation.md`.


auser:
Reviewer note — DX tradeoff to weigh: every *cold* `cargo build`/`cargo run` now pays the host-relevant helper compile (the build script can't know the subcommand), not just the first `machine run`. And `rerun-if-changed` on `crates/mvm-core/src` / `crates/deps/libkrun-sys/src` means editing those retriggers a supervisor rebuild on the next cargo command. This is the deliberate "correct-by-construction over lazy" trade (a silently-stale `mvm-libkrun-supervisor` is a claim-10 egress/audit hazard), amortized by `rerun-if-changed` on warm builds and `MVM_SKIP_EMBED_BINARIES=1` for the fast test loop — but if the cold-build tax feels heavy in practice, narrowing the watched source closure is the first knob to turn. End-user (downloaded `mvmctl`) builds nothing: the helpers resolve from the release tarball next to `mvmctl`.

PR #1620, opened 2026-07-10
ci: finish removing the deleted mvm-vz-supervisor references

auser:
Finishes the vz-backend deletion cleanup: removes the dead `mvm-vz-supervisor` references that survived after the vz backend (`vz.rs`, `vz_builder.rs`, `apple_container.rs`, and the `mvm-vz-supervisor` `[[bin]]`) was removed.

These would hard-fail the moment they were exercised — `cargo build --bin mvm-vz-supervisor` errors with "no bin target".

## Changes

- **`.github/workflows/ci-full.yml`** — delete the two wholly-vz lanes `vz-macos` and `vz-macos-26`. Both ran `cargo build … --bin mvm-vz-supervisor` and path-filtered on deleted files; neither is referenced by any gate's `needs:` (only `needs: test` exists), so removal is safe. Also refresh two comments that named the removed bin / the deleted vz lane.
- **`scripts/capture-plan-205-live-gates.sh`** — removed. It was a vz-only, opt-in live-gate capture runner (`cargo build --bin mvm-vz-supervisor`, `MVM_BUILDER_BACKEND=vz`, `.vzsave` snapshots, macOS-26-Vz host guard); with the backend gone it can't run. Its role as historical evidence for the completed Plan 205 acceptance lane is preserved in the plan/spec records.
- **`crates/mvm-cli/src/commands/env/dev_vz/vm_helpers.rs`** / **`nix/lib/workspace-filter.nix`** — refresh doc comments that named `mvm-vz-supervisor`.

## Scope

The identical dead reference in `Justfile` and `.github/workflows/release.yml` (the latter a macOS **release-blocker** — `cargo build --bin mvm-vz-supervisor` breaks the downloadable tarball build) is fixed in #1618. This PR and #1618 touch disjoint files, so they merge in either order.

No Rust logic changes (comment-only), so the normal build/test gates are unaffected; YAML validated.


auser:
Added commit `c673bda3d`: removes the leftover `PackBackend::Vz` variant (+ its `Display` arm and a test fixture) from `mvm-core` — the sole remaining Vz *backend* reference in the workspace, so no pack can declare or resolve a vz backend. Verified `cargo check --workspace --all-targets` clean (no exhaustive-match breaks) and `cargo nextest run -p mvm-core` 1596/1596. Branch also merged latest `main` (picks up #1618 + #1625). This makes #1620 the complete "vz backend is gone everywhere" cleanup.

PR #1631, opened 2026-07-10
Plan 213 SP1: versioned pack cache + lifecycle facade + mvmctl pack

auser:
Plan 213 **SP1** toward instant first-use: give the content-addressed pack cache a deterministic **active version** per pack class, plus lifecycle operations exposed through the `mvmctl::core::pack_cache` facade and an `mvmctl pack` CLI. Design: `specs/notes/instant-first-use-pack-design.md`; plan: `specs/plans/243-versioned-pack-cache-lifecycle.md`.

## Why
Today the cache stores one dir per `pack_hash` (multiple versions already coexist), but `resolve_pack` returns whichever `read_dir` hits first — non-deterministic, and there's no way to update/rollback. This adds a versioned index + active pointer so a new release is fetched+verified *alongside* the current one, promotion/rollback just move a pointer, and nothing is destructively overwritten.

## What
- **Versioned index + active pointer** (`mvm-core/src/pack_cache/index.rs`, `pack_cache.rs`): `<cache>/packs/index.json` records each promoted pack's provenance (`pack_hash`, kind/arch/backend, channel, release version, promoted-at) + an active pointer per `(kind, arch, backend)`. Version labels come from the **download context**, not a (breaking) manifest change.
- **`resolve_pack` prefers the active pointer**, re-verifying it through the existing `PackVerifyCtx`, falling back to the scan if there's no index or the active pack is missing/corrupt. Verification is never bypassed.
- **Lifecycle facade** (reachable by CLI + SDK + mvmd): `promote_and_record`, `set_active_version`, `list_versions`, version-aware `prune_versions` (never removes the active version). The facade derives `mvm_cache_dir()` internally so callers can't pass the wrong root.
- **`mvmctl pack`** — `list` / `rollback` / `prune` / `download` / `update`. `download`/`update` are wired for the **builder pack** (reusing the existing release-fetch + real `PackVerifyCtx`); `runtime`/`dev-image` return an explicit "not yet wired" error (no silent stub).
- **Dev-up path** now records the builder pack's version on download; `mvmctl cache prune` gained version-aware retention (keep active + newest N).

## Scope
OUT (later sub-projects): runtime/dev-image download wiring; the `mvmctl prepare` phase + install.sh (SP2); the seed Nix closure (SP3); the benchmark harness (SP4).

## Testing
Full workspace gate green: `fmt --all`, `clippy --workspace --all-targets -D warnings`, `build --all-targets`, `nextest --workspace`, doctests. New coverage: index primitive (6), persist + active-pointer resolve incl. corrupt-fallback (3), facade incl. an end-to-end guard (`promote_and_record` → `resolve_pack` returns active), CLI smoke tests, dev-up version-recording, prune-report helper.

### DIFF
--- a/crates/mvm-cli/src/commands/cmd_audit.rs
+++ b/crates/mvm-cli/src/commands/cmd_audit.rs
@@ -215,6 +215,7 @@ impl Commands {
             Commands::Prepare(_) => "prepare",
             Commands::Manifest(_) => "manifest",
             Commands::Image(_) => "image",
+            Commands::Pack(_) => "pack",
             // `machine <sub>`: folded advanced ops (pause/snapshot/set-ttl/…)
             // keep their per-op verb; native lifecycle verbs report `machine`.
             Commands::Machine(a) => a.action.verb_name(),
--- a/crates/mvm-cli/src/commands/dispatch.rs
+++ b/crates/mvm-cli/src/commands/dispatch.rs
@@ -25,6 +25,7 @@ impl TopLevelCommand for Commands {
             Commands::Build(a) => build::group::run(cli, a, cfg),
             Commands::Manifest(a) => manifest::run(cli, a, cfg),
             Commands::Image(a) => image::run(cli, a, cfg),
+            Commands::Pack(a) => pack::run(cli, a, cfg),
             Commands::Machine(a) => machine::run(cli, a, cfg),
             Commands::Storage(a) => storage::run(cli, a, cfg),
             Commands::ShellInit(a) => env::shell_init::run(cli, a, cfg),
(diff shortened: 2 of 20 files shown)


auser:
Final whole-branch review (opus): **Ready to merge — Yes.** The security-critical invariant is confirmed and directly tested — an active-pointer-selected pack is always re-verified through the same `verify_candidate_pack_dir` as the scan path, and a corrupt/tampered active pack falls back to a verified one (never errors or serves tampered bytes); the index is a genuine fail-open accelerator that cannot force selection of an unverified pack. Four Minor findings, none blocking. Applied two of them in `777ad5615`: an explicit corrupt-`index.json` fail-open test, and a doc clarification that the first version recorded for a key becomes active. Deferred (non-blocking): an advisory lock around the index read-modify-write (only matters if concurrent `pack`/`cache` invocation becomes a real workflow — revisit in the SP2 prepare flow), and the `rollback`-without-`--to` "second-newest" semantics (help text is accurate).

PR #1745, opened 2026-07-21
feat(cli): surface image/secret/bundle/trust in --help (WS7)

auser:
First WS7 (simple-CLI) slice. Promotes four first-class user-facing nouns from hidden to the visible top-level `--help` surface:

- `image` — inspect/manage the cached OCI images
- `secret` — manage local secret namespaces
- `bundle` — seal/fetch/verify portable VM bundles
- `trust` — manage trusted bundle publishers

Behavior-preserving: these already dispatched identically; only their `#[command(hide = true)]` → `#[command(display_order = N)]` visibility changes. No test asserted they were hidden; clippy + the CLI flag-contract tests (`tests/cli.rs`, 8/8) pass.

Context: the broader WS7 redesign (resolving top-level-vs-`machine`-nested redundancy — `build`/`machine build`, `ls`/`machine ls` — and the canonical noun-verb shape) is a design decision I'll bring separately; this slice is the unambiguous, zero-risk part.

### DIFF
--- a/crates/mvm-cli/src/commands/mod.rs
+++ b/crates/mvm-cli/src/commands/mod.rs
@@ -130,8 +130,8 @@ pub(in crate::commands) enum Commands {
     /// Manage built manifest slots
     #[command(hide = true)]
     Manifest(manifest::Args),
-    /// Inspect cached OCI images
-    #[command(hide = true)]
+    /// Inspect and manage cached OCI images
+    #[command(display_order = 9)]
     Image(image::Args),
     /// Inspect the dm-thin storage pool
     #[command(hide = true)]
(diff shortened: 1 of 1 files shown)


auser:
Closing on reflection. This promote reverses a *deliberate, tested* design decision — `commands::tests::top_level_help_hides_infra` explicitly asserts image/secret/bundle/trust are infrastructure commands hidden from default `--help` (accessible, just not surfaced). A tight visible core with power-commands hidden-but-reachable is arguably already the minimal, legible surface the simplification is aiming for, so surfacing four more nouns works *against* that goal rather than for it. Not overriding an intentional design call by editing its guard test. The visible CLI surface stands as-is.

PR #1748, opened 2026-07-21
feat(net): make Firecracker a vsock-only (Model-B) egress backend

auser:
Firecracker becomes a NIC-less, vsock/UDS-only egress backend, matching libkrun/HVF. This is the cutover that lets the smoltcp-L3 stack be deleted (separate follow-up, gated on live-KVM validation). One atomic change — any partial state is a bypass or breaks egress:

- **Token** — FC emits `mvm.vsock_egress=1` (gated `allows_egress() && !bound_secrets`), so the guest brings up the `mvm-egress-client` vsock proxy.
- **Host endpoint** — spawn the substitution endpoint over the **Uds** transport at `<vsock>_<EGRESS_PORT>`. FC's userspace virtio-vsock muxes a guest dial to `CID_HOST:EGRESS_PORT` onto that per-port host UDS, **not** a host `AF_VSOCK` listener (which an FC guest can never reach) — so the previous `Vsock` transport was unreachable. Self-gates on `secrets OR allows_egress`; `network_policy: Some` makes the host `EgressGate` the sole claim-10 authority; `raw_egress` per secrets; `terminator_listen: None`.
- **No routable NIC** — drop the `/network-interfaces/net1` device, `mvm.ip=`/`mvm.gw=`, the bridge/TAP provision, and the post-boot nft `:80`/`:443` REDIRECT. So **both ingress and egress now flow over vsock/UDS** (ingress via vsock port-forward). Scoped to the FC *workload* path only — the FC *builder* VM and the legacy `mvmctl start` dev path keep their NICs (verified).
- **Caps** — `no_routable_guest_nic`/`host_vsock_proxy` true, `tap_networking` false.
- **Transport declaration** — `egress_substitution_transport()` → `VsockUdsChannel` (was the now-removed `NftTerminator`), identical to HVF: proxy-aware substitution, no transparent terminator.

**Security-property note:** claim-10 enforcement moves from admission-time IP pinning at nftables to the host-side `EgressGate` (re-resolves per connection). Non-proxy-aware (transparent `:80`/`:443`) workloads are no longer accepted on FC — the honest consequence of NIC-less egress, same as HVF.

Verified: `mvm-runtime` clippy `-D warnings`, the FC/egress/network test sweep + `workload_backend` suite (FC now `VsockUdsChannel`, refused by the transparent-terminator gate), the `x86_64-unknown-linux-gnu` cross-build (Linux-only endpoint edit), and `check-vsock-only-egress` / `check-no-spec-refs-in-comments`. The end-to-end live witness (FC guest egressing over vsock; claims 12/13 rejection ladder) is a deferred KVM-box follow-up, same posture as the existing libkrun/HVF Model-B code. smoltcp-L3 stack left in place until that lands.

### DIFF
--- a/crates/mvm-cli/src/doctor/warm_start.rs
+++ b/crates/mvm-cli/src/doctor/warm_start.rs
@@ -283,7 +283,10 @@ mod tests {
 
         let fc = by("firecracker").unwrap();
         assert_eq!(fc.snapshot_tier, "live-memory");
-        assert!(fc.tap_networking, "firecracker uses a host TAP device");
+        assert!(
+            !fc.tap_networking,
+            "firecracker is vsock-only: no routable guest NIC"
+        );
         assert!(fc.vsock);
         assert!(fc.balloon);
         assert!(
--- a/crates/mvm-runtime/src/backend.rs
+++ b/crates/mvm-runtime/src/backend.rs
@@ -214,11 +214,17 @@ impl VmBackend for FirecrackerBackend {
         // `VmStartConfig::mem_initial_mib` is `Some`. Capability is
         // advertised unconditionally so the host-side controller can
         // discover support before deciding to plumb a workload.
+        //
+        // A workload guest is vsock-only: it attaches no routable NIC, so all
+        // ingress/egress rides the host-mediated vsock proxy — matching the
+        // libkrun/HVF posture.
         VmCapabilities {
             pause_resume: true,
             snapshots: true,
             vsock: true,
-            tap_networking: true,
+            tap_networking: false,
+            no_routable_guest_nic: true,
+            host_vsock_proxy: true,
             balloon: true,
             fs_quick_checkpoint: false,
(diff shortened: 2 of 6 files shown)


auser:
Superseding this direct-mirror cutover with the uniform-trait convergence: Firecracker will route through `WorkloadRunner<FcDriver, RealEndpointSpawner>` like HVF, so the vsock egress data-plane (claim-10 gate + claims 12/13 substitution) lives in ONE place every backend is forced through — the point being to make the security claims structurally provable, not per-backend copies that drift. The live witness on this branch validated FC's vsock transport (that carries forward to FcDriver) and surfaced two real defects: Defect A (egress-client path) is already fixed on main by #1747, Defect B (endpoint seccomp lseek) is fixed in #1749. Keeping the branch for reference; the no-NIC/token guest-cmdline changes fold into FcDriver's spec assembly.

PR #1775, opened 2026-07-23
feat(net): add policy-gated in-seam DNS resolver

auser:
## Summary

- add a bounded, `forbid(unsafe)` DNS question/A/AAAA codec with a cargo-fuzz target and security-CI witness
- add a pins-first host DNS handler on the existing vsock egress seam, gated by the workload `NetworkPolicy`
- reject unsafe live answers with a strict SSRF/rebinding classifier unless an address is explicitly pinned
- emit metadata-only, chain-signed per-query DNS audit events and enforce per-workload rate/concurrency limits
- add a guest UDP/TCP loopback DNS stub on `127.0.0.1:53` and seed resolver configuration across OCI, Nix-busybox, and Stage 0 init paths
- add live BDD scenarios for admitted resolution, refusal, and audit-chain visibility; wire the codec/filter/audit tests into claim 10

## Why

NIC-less workloads previously had no resolver, so applications that performed their own DNS lookup failed even when their destination was admitted. DNS also needed to remain inside the same default-deny, auditable egress boundary rather than becoming a covert bypass.

This follows the admitted-connect completion in #1767: one `NetworkPolicy` now gates both name resolution and the subsequent TCP connection.

## Security and user impact

Allow-listed names resolve through the host-mediated vsock seam without adding a routable guest NIC. Non-admitted names are refused before upstream resolution. Returned addresses are filtered against private, loopback, link-local, ULA, and metadata ranges, and every query is rate-limited and chain-audited without recording query payloads or secret data.

## Validation

- `cargo nextest run -p mvm-protocol -p mvm-core -p mvm-hostd`: 2,931 passed, 1 skipped
- `cargo nextest run -p mvm-agentd --features addons`: 535 passed
- serialized workspace test suite: passed before rebase
- workspace/all-target clippy with agent addons: passed before rebase
- Linux cross-builds for OCI init and Stage 0 init: passed
- `cargo run -p xtask -- check-claim-catalog`: clean, 16 claims / 41 witnesses
- formatting, workflow YAML parsing, and `git diff --check`: passed

The `@live` DNS scenarios require the KVM/HVF witness host and are intentionally deferred to that lane.

After rebasing onto current `main`, the full hermetic conformance run has four unrelated failures in the newly merged workload-identity suite: `mvmctl workload address` exits 2 because `workload` is currently an unrecognized subcommand. All DNS/default-deny scenarios executed by that run pass; the live DNS scenarios are parsed and skipped without `MVM_BDD_LIVE=1`.


auser:
## Live-boot finding: `machine run --image <read-only OCI image>` kernel-panics

Ran this branch on HVF (macOS 26, arm64). Every `machine run --image alpine --allow-host example.com -- …` fails identically with **"guest agent did not become reachable within 30s"**. The guest console shows the cause:

```
mvm-oci-init: failed to point resolv.conf at the loopback DNS stub: write /etc/resolv.conf: Read-only file system (os error 30)
```

**Root cause:** `mvm-oci-init` calls `seed_loopback_resolver()` → `seed_resolv_conf_bytes`, which falls through to a direct `write("/etc/resolv.conf")` (alpine's minimal image ships no `/etc/resolv.conf`, so the `exists()` bind-mount branch is skipped), and that write is **EROFS** on the verity-sealed rootfs. `mvm-oci-init` then `std::process::exit(1)` on the failure — killing PID 1 → kernel panic. **The DNS feature fails to boot on any read-only OCI image.**

**Fix proposed in #1778** (`fix(agentd): don't panic init when the resolv.conf seed fails on a read-only rootfs`): make the seed non-fatal (log + continue) — the workload still reaches admitted egress through the loopback proxy, unlike the missing-agent / missing-egress-client cases where fail-closed is correct.

**Deeper gap (not in #1778):** even non-fatal, resolv.conf can't be repointed on a read-only rootfs with no writable `/etc/resolv.conf` — so in-guest `getaddrinfo` via the stub won't work for such images (proxy egress still will). That needs a writable-`/etc` mechanism (e.g. a tmpfs overlay for `/etc/resolv.conf`).

Positive side-signal: the mandatory-deny CIDRs (incl. CGNAT `100.64.0.0/10`) install correctly on every boot.

PR #1778, opened 2026-07-23
fix(dns): reject IPv4-mapped IPv6 SSRF + audit malformed DNS queries

auser:
Two follow-up hardening fixes for the in-seam DNS resolver (stacks on #1775), surfaced by an independent security review of that PR.

## C1 — SSRF via IPv4-mapped IPv6 (the merge-blocker)

`::ffff:169.254.169.254` (the IPv4-mapped form of the cloud-metadata IP) bypassed **both** egress deny filters: `dns_answer_forbidden` and the connect-time `is_mandatory_deny` each classified it as an opaque IPv6 address, and the mandatory-deny ranges are IPv4 `IpNet`s that `contains()` false-matches across address families. Under an unrestricted policy a workload could reach metadata / RFC1918 space via a crafted AAAA answer **or** a direct connect to the mapped address (Linux dual-stack routes it to the embedded IPv4).

- New shared `unmap_v4_mapped` collapses `::ffff:a.b.c.d` → `a.b.c.d` before both checks.
- Adds CGNAT `100.64.0.0/10` to `dns_answer_forbidden` — it was already in the connect-time deny list, so the "stricter" DNS-answer guard was out of sync.
- Witness tests for the mapped metadata/RFC1918/loopback/CGNAT forms on both filters; a mapped **public** address stays allowed.

## C2 — malformed DNS queries bypassed audit and rate-limit

`decode_query` ran before the rate guard, and the malformed branch emitted no audit event — so a guest could flood malformed frames at unlimited rate, invisible to the chain-signed log.

- Every malformed frame now consumes a rate token (throttled like a valid query) and emits a `malformed` audit event, on both the async and blocking transports.
- Unit + end-to-end handler witnesses.

## Verification
`mvm-core` policy tests 39/39, `mvm-hostd` DNS tests 32/32, clippy `-D warnings` clean, nightly rustfmt clean.

## Not included (follow-ups)
- 6to4 / Teredo / IPv4-compatible embedded-v4 forms (lower-confidence than the kernel-routed mapped form).
- Connection-level flood bound (each DNS query opens a fresh vsock connection spawned upstream of `serve_dns`; C2 adds backpressure + audit but does not bound the accept rate).
- Other review findings still open on #1775: plan-unbound DNS audit, `:53` boot-readiness race, `TokenBucket` refill test, OCI-init uid-0, EDNS0 doc note.

### DIFF
--- a/crates/mvm-agentd/src/bin/mvm-oci-init.rs
+++ b/crates/mvm-agentd/src/bin/mvm-oci-init.rs
@@ -33,10 +33,14 @@ mod linux {
         if cmdline_has_flag("mvm.vsock_egress=1") {
             bring_loopback_up();
             if let Err(error) = mvm_agentd::guest_net::seed_loopback_resolver() {
+                // Non-fatal: a read-only image rootfs may have no writable
+                // /etc/resolv.conf to repoint, but the workload still reaches its
+                // admitted egress through the loopback proxy. Log and continue —
+                // panicking PID 1 here would fail an otherwise-bootable VM.
                 eprintln!(
-                    "mvm-oci-init: failed to point resolv.conf at the loopback DNS stub: {error}"
+                    "mvm-oci-init: could not point resolv.conf at the loopback DNS stub \
+                     (continuing; proxy egress is unaffected): {error}"
                 );
-                std::process::exit(1);
             }
             // Egress is required when this flag is set; the client is the only guest
             // path to the admitted network. Resolve overlay-first (respecting the
--- a/crates/mvm-core/src/policy/dns_guard.rs
+++ b/crates/mvm-core/src/policy/dns_guard.rs
@@ -10,12 +10,16 @@ use std::net::IpAddr;
 /// an admitted public hostname from rebinding to an internal service.
 #[must_use]
 pub fn dns_answer_forbidden(ip: IpAddr) -> bool {
-    match ip {
+    // An IPv4-mapped IPv6 answer (`::ffff:a.b.c.d`) reaches the embedded IPv4
+    // destination on a dual-stack connect, so classify it as that IPv4 address
+    // rather than letting it slip past the IPv4 rules as an opaque IPv6 one.
+    match crate::policy::network_policy::unmap_v4_mapped(ip) {
         IpAddr::V4(address) => {
             address.is_private()
                 || address.is_loopback()
                 || address.is_link_local()
                 || address.is_unspecified()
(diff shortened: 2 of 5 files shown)


auser:
## Rebased onto main + live-validated on HVF

#1775 squash-merged the in-seam DNS resolver into `main` **without** these fixes, so `main` currently ships the resolver with three open defects this PR closes. Rebased onto `main` (3 commits, was stacked on the now-merged feature branch).

### Why this matters now
- **C3 — boot regression on `main`.** On a live HVF boot, **every** `mvmctl machine run --image <read-only OCI image>` kernel-panics: `mvm-oci-init` does a direct `write("/etc/resolv.conf")`, which is EROFS on the verity-sealed rootfs, then `exit(1)`s → PID 1 dies → panic. This PR makes the seed non-fatal.
- **C1 — SSRF.** `::ffff:<private-v4>` bypasses both `dns_answer_forbidden` and the connect-time `is_mandatory_deny` (IPv4-only `IpNet`s false-match across families). Normalized in both filters; CGNAT `100.64/10` added to the DNS-answer guard.
- **C2 — audit/rate-limit gap.** Malformed DNS frames escaped both the rate limiter and the chain-signed log; now counted + audited on both transports.

### Live validation (HVF, macOS 26 arm64), post-fix
```
mvm-oci-init: could not point resolv.conf … (continuing; proxy egress is unaffected)   ← no panic
guest DNS stub started listen=127.0.0.1:53
nslookup example.com → 127.0.0.1:53 → 2606:4700:10::6814:179a                          ← in-guest resolution works
```
The VM boots, the agent becomes reachable, and in-guest DNS resolution via the stub works.

### Follow-ups (not in this PR)
1. **OCI-rootfs cache doesn't invalidate on guest-source change** — the assembled `~/.mvm/cache/oci/{prepared-roots,rootfs}` reuse the old injected `mvm-oci-init` after a guest edit; had to be cleared by hand. The rootfs cache key should cover the injected guest-init fingerprint.
2. **`wget https://…` hangs (rc=124)** — the admitted-egress connect leg (ADR-032 Part 1 territory); the stub returned an IPv6-only answer and the host connect stalled (no IPv4 fallover completed). Separate from this PR.
3. **resolv.conf can't be repointed on a read-only rootfs with no writable `/etc/resolv.conf`** — `getaddrinfo`-based resolution won't hit the stub for such images (busybox `nslookup` still works via its `127.0.0.1` default). Needs a writable-`/etc` mechanism.

PR #1790, opened 2026-07-23
fix(egress): fail over to the pinned IPv4 when a literal IPv6 dial stalls

auser:
## Problem

An admitted `mvmctl machine run --image alpine --allow-host example.com -- wget -qO- https://example.com` hung to timeout (rc=124) on HVF when the guest resolved the name in-guest (DNS-over-vsock stub) to an IPv6-preferred answer and dialed the **literal IPv6**.

Both mechanisms one would suspect are already correct:

- The DnsPin carries **both** A and AAAA (`resolve_network_policy_pins` → `to_socket_addrs`), and the hostname path (`decide_hostname_request` → `admitted_ips`) already returns them IPv4-first.
- `raw_egress::connect_first_admitted` already does correct happy-eyeballs with a 3s/IP budget.

The gap is `EgressGate::decide_addr`: for a **numeric literal** target it returned `ips: vec![ip]` — only the single dialed address. So a guest dialing `[2606:…]:443` gave the host a one-element IPv6 admitted set with no IPv4 sibling; happy-eyeballs had nothing to fall over to when the IPv6 was unreachable → stall → rc=124.

## Fix

Re-expand a pinned literal target to its pin's full admitted set, IPv4 first (reusing the existing `admitted_ips`) — the same `EgressVerdict::Allow` contract the hostname path already honors. Adds `DnsPinRegistry::find_by_ip` for the reverse (IP → pin) lookup, deterministic on the lowest `dest` when hosts share an anycast address.

**Not a policy widening (claim-10):** every returned address is still independently gated by `CanonicalEgress::permits` (mandatory-deny + banned-SSH-port enforced first), so expansion can only ever reach addresses/ports an allow-listed host already admits.

## Tests

- `literal_ipv6_dial_expands_to_pinned_ipv4_sibling` — reproduces the hang at the gate altitude (fails against the old `vec![ip]`).
- `literal_ipv4_dial_expands_to_pinned_ipv6_sibling` — the symmetric direction.
- `literal_dial_on_a_shared_ip_expands_to_the_lowest_dest_pin` — documents the deterministic shared-anycast behaviour.
- `literal_dial_without_a_matching_pin_stays_single` — no spurious expansion.
- `registry_find_by_ip_*` (mvm-protocol) — the reverse lookup + shared-IP determinism.

## Verification

`mvm-protocol` lib **348/348**, `mvm-runtime` lib **985/986** (the one failure is the `MVM_SKIP_EMBED_BINARIES` env gate — a missing `mvm-substitution-endpoint` helper, unrelated; passes in CI which builds it). `clippy -D warnings` clean, nightly `cargo fmt --all --check` clean, no spec-ref-lint hits. Independently `rust-reviewer`'d, security angle included.

## Follow-up

Live HVF end-to-end re-validation is pending #1778 landing (a `machine run --image` off `main` still kernel-panics on the C3 boot bug until then); the fix is verified at the gate altitude with a reproduction test.

### DIFF
--- a/crates/mvm-protocol/src/policy/dns_pin.rs
+++ b/crates/mvm-protocol/src/policy/dns_pin.rs
@@ -173,6 +173,16 @@ impl DnsPinRegistry {
         self.pins.get(dest)
     }
 
+    /// The first pin (in `dest` sort order) whose admitted set contains `ip`,
+    /// or `None` — the reverse of [`lookup`], which keys on the destination
+    /// string. Used to recover a host's full dual-family address set when a
+    /// guest dialed one of its addresses as a literal. When several pinned
+    /// hosts share an address (CDN / anycast fronting), the lowest `dest` wins
+    /// deterministically (`pins` iterates in sorted key order).
+    pub fn find_by_ip(&self, ip: &IpAddr) -> Option<&DnsPin> {
+        self.pins.values().find(|pin| pin.matches(ip))
+    }
+
     /// Number of pins currently in the registry.
     pub fn len(&self) -> usize {
         self.pins.len()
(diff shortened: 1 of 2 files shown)


auser:
Superseded by #1787, which contains the same IPv4 happy-eyeballs fallback fix.

PR #1804, opened 2026-07-24
fix(mvm-fs): OCI multi-layer unpack replaces prior-layer paths (#1674)

### LINKED ISSUE #1674: mvm-oci: O_EXCL unpack refuses any path a later layer rewrites (breaks most multi-layer images)
## Summary

`mvm-oci`'s hardened unpacker opens every regular file with `O_CREAT | O_EXCL` ([`unpack.rs` `write_regular_file`](https://github.com/tinylabscom/mvm/blob/main/crates/mvm-oci/src/unpack.rs)). OCI images are unpacked layer-by-layer into one root with overlay semantics, so any path that a **later layer rewrites** (very common: `apt`/`npm`/`useradd` all rewrite `/etc/passwd`, `/etc/group`, `/etc/shadow`, `/etc/gshadow`, `/etc/subuid`, `/etc/subgid` and their `-` backups) hits `EEXIST` on the second layer. `map_resolve_errno` maps `EEXIST` to `RefusalReason::MalformedHeader`, and `mvmctl machine run` treats a non-empty refusal set as fatal — so the boot aborts.

## Impact

Effectively **any real multi-layer Debian/Ubuntu-based image fails to pull**. Confirmed with a `node:20-bookworm-slim`-based image: pull refuses exactly `etc/{passwd,passwd-,group,group-,shadow,gshadow,subuid,subgid}` and aborts. Single-layer images (or images whose layers never rewrite a shared path — e.g. `alpine`, `busybox`) pull fine, which is why this hasn't surfaced.

Two secondary problems:
1. **Misleading diagnostic** — `EEXIST` reported as `MalformedHeader` ("malformed tar header"), which sent me hunting for a corrupt tar. The entries are pristine (type `0`, mode `0644`/`0640`, no PAX).
2. The refusal is silently fatal at the `machine run` layer despite the unpacker's own comment that a single bad entry "shouldn't poison a multi-thousand-entry layer."

## Repro

```
# any image whose apt/npm layer rewrites /etc/passwd over the base layer's copy
mvmctl machine run --image ghcr.io/…/node-based-image:latest -- true
# → layer unpack refused entries: [etc/passwd, etc/group, etc/shadow, …] → abort
```

## Root cause

`unpack.rs`:
```rust
let flags = OFlags::WRONLY | OFlags::CREATE | OFlags::EXCL | OFlags::NOFOLLOW | OFlags::CLOEXEC;
```
`O_EXCL` is correct for TOCTOU safety within one layer, but wrong across layers where the newer layer must overwrite the older file.

## Suggested fix

Overlay-correct per-file replacement that keeps the anti-symlink-swap guarantee, e.g. for regular files write to a fresh temp name in the same dir (O_EXCL) then `renameat` over the target, or `unlinkat` the prior entry (after the same `RESOLVE_IN_ROOT | NO_SYMLINKS` lookup) before the `O_EXCL` create. Whiteouts/opaque dirs need the same treatment. Also: map `EEXIST` to a distinct `RefusalReason` so the diagnostic isn't "malformed header."

## Workaround in use

Flattening the demo image to a single layer (`docker export | docker import`) so each path appears once — see mvm-studio `demo/image/build.sh`. Works, but shouldn't be necessary.

Found while driving the mvm-client facade from mvm-studio (plan 04 Claude-in-a-microVM demo). Related: #1590.

auser:
Closes #1674.

Threads `prior_layer_paths` through `unpack_layer` so later OCI layers can remove paths written by earlier layers before recreating them. This preserves `O_CREAT | O_EXCL` / `O_NOFOLLOW` first-creation safety for genuinely new paths while allowing later layers to override a prior-layer file/directory/symlink/device. Same-layer duplicates remain refused.

- Adds `unpack_layer_with_prior_paths` and `UnpackReport::paths_written`.
- Removes prior-layer leafs before regular file, symlink, hardlink, device node, and directory materialization.
- Accumulates `paths_written` across layers in `mvm-cli image pull/ingest` and `mvm-client local`.
- Adds BDD scenarios in `features/suites/s9_image_materialization/` covering multi-layer override, type-changing replacement, whiteout + recreation, and same-layer duplicate refusal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2760,8 +2760,10 @@ dependencies = [
  "assert_cmd",
  "cucumber",
  "mvm-core",
+ "mvm-fs",
  "mvm-runtime",
  "serde_json",
+ "tar",
  "tempfile",
  "tokio",
 ]
--- a/crates/mvm-cli/src/commands/image/ingest.rs
+++ b/crates/mvm-cli/src/commands/image/ingest.rs
@@ -97,9 +97,16 @@ pub(super) fn ingest_archive_from_reader<R: Read>(
     }
     fs::create_dir_all(&unpacked_root)
         .with_context(|| format!("create {}", unpacked_root.display()))?;
+    let mut prior_layer_paths = std::collections::HashSet::new();
     for layer in &image.layers {
-        unpack_layer_bytes(&layer.descriptor, &layer.bytes, &unpacked_root)
-            .with_context(|| format!("unpack layer {}", layer.descriptor.digest))?;
+        let report = unpack_layer_bytes(
+            &layer.descriptor,
+            &layer.bytes,
+            &unpacked_root,
+            &prior_layer_paths,
+        )
+        .with_context(|| format!("unpack layer {}", layer.descriptor.digest))?;
+        prior_layer_paths.extend(report.paths_written);
     }
 
     let rootfs_rel = format!("rootfs/{manifest_hex}-{}/rootfs.ext4", oci_runtime_tag());
--- a/crates/mvm-cli/src/commands/image/pull_core.rs
+++ b/crates/mvm-cli/src/commands/image/pull_core.rs
@@ -3,9 +3,10 @@
 //! registry pull. This is the run-path entry every OCI source funnels
 //! through once it's been classified as a registry reference.
 
+use std::collections::HashSet;
 use std::fs;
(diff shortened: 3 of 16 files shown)


auser:
Dequeuing temporarily so I can push the rebased branch after resolving the BDD conflict.

PR #1846, opened 2026-07-25
fix(lineage): atomically publish image.created nodes

### LINKED ISSUE #1833: image.created emit-then-save is non-atomic: a save failure leaves a phantom audit entry
Part of #1806 (surfaced in B2 review of #1808). **Deferred, low-severity.**

`record_image_node` (`crates/mvm-cli/src/commands/build/image_lineage.rs`) emits
the chain-signed `image.created` entry **before** `ImageStore::save`. The
ordering is deliberate and correct for the security posture: the store must
never hold a node the audit chain cannot anchor, so emit-before-save is the
right choice (an un-anchorable stored node would fail `verify_image_lineage`).

The gap: the two writes aren't atomic. If `save` fails *after* `emit` succeeds
(disk full on `~/.mvm/images`, permission drift, process kill), the chain
carries an `image.created` entry with no corresponding stored node. A retry's
`head_for` finds nothing → treats it as a fresh genesis → emits a **second**
`image.created` for the same identity. The chain accumulates phantom entries no
lineage walk references.

**Impact:** harmless to `verify_image_lineage` (walks start from stored nodes)
and the audit chain stays cryptographically valid; it only pollutes `mvmctl
trust audit verify` / `explain` output. Mostly self-heals (the artifacts are on
content-addressed cache paths).

**Fix options:** dedup the emit (skip if the signed chain already records this
`node_digest`), or write the node to a temp path → emit → atomic rename. Neither
is free (a chain read or a temp+rename dance), which is why it's deferred rather
than folded into B2.

auser:
Closes #1833

## Summary

- stage serialized image-lineage nodes before emitting `image.created`
- atomically publish the staged node only after audit succeeds
- clean up abandoned stages and cover staging failures with tests

## Verification

- `cargo check --workspace`
- `cargo test --workspace -- --test-threads=1`

### DIFF
--- a/crates/mvm-cli/src/commands/build/image_lineage.rs
+++ b/crates/mvm-cli/src/commands/build/image_lineage.rs
@@ -3,8 +3,8 @@
 //! A build produces a compiled microVM image; this module content-addresses
 //! that image into an [`ImageNode`], hash-links it onto the prior build of the
 //! same slot identity, and binds its creation into the host-signed audit chain
-//! *before* the build reports success. The store never holds a node the audit
-//! chain cannot anchor.
+//! after staging the node, then atomically publishes it before the build reports
+//! success. The store never holds a node the audit chain cannot anchor.
 //!
 //! # Lineage is PROVENANCE, not AUTHORIZATION
 //!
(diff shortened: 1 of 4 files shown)


auser:
Superseded by #1844, which landed the issue fix while this branch was in CI.

PR #1854, opened 2026-07-27
test(mvm-build): fix runtime-overlay build witnesses

auser:
## What

Fixes two stale bugs in the (un-CI'd) `runtime_overlay_build` integration test. The overlay's on-disk verity is correct and unchanged — these are **test** bugs that only surface where host nix is present (a Linux + nix dev box), so CI never exercised them.

1. **Verity geometry** — the test computed the data-block count over `4096`, but the flake builds the overlay with `overlayBlockSize = 1024` (`mkfs.ext4 -b 1024` / veritysetup `--data-block-size=1024`, `--hash-block-size=4096`), so the sidecar was exactly **4×** the modelled size (260 vs 65 hash-tree blocks). Now computed over `1024` with the `4096` hash-block size, both named constants tied to the flake values.

2. **Resolver compatibility** — the test hand-staged the four canonical files but skipped the `checksums-sha256.txt` manifest that `RuntimeOverlayResolver::resolve`'s integrity check requires (written by `install_overlay_into_cache`, not by the build itself). It now populates the cache via the real `install_overlay_into_cache` — exercising the genuine build → install → resolve path end-to-end — instead of manual staging, and drops the now-dead `copy_file` helper (and its orphaned `Path` import).

## Verification

- `cargo zigbuild --target x86_64-unknown-linux-gnu -p mvm-build --tests` with `-D warnings` — **clean**. This is the meaningful compile gate here: the file is `#![cfg(target_os = "linux")]`, so the macOS pre-commit clippy skips it entirely.
- **Behavioral:** `cargo test -p mvm-build --test runtime_overlay_build` against a real Linux + Determinate-Nix host (the actual `nix build`). CI does not run this test, so the box run is the behavioral proof — **in progress; will confirm the three tests green here before this is merge-ready.**

### DIFF
--- a/crates/mvm-build/tests/runtime_overlay_build.rs
+++ b/crates/mvm-build/tests/runtime_overlay_build.rs
@@ -19,10 +19,11 @@
 
 #![cfg(target_os = "linux")]
 
-use mvm_build::runtime_overlay::{OverlayBuildSpec, build_overlay_with_nix};
+use mvm_build::runtime_overlay::{
+    InstallOptions, OverlayBuildSpec, build_overlay_with_nix, install_overlay_into_cache,
+};
 use mvm_core::arch::GuestArch;
 use mvm_fs::overlay::RuntimeOverlayResolver;
-use std::path::Path;
 use tempfile::TempDir;
 
 fn skip_if_no_nix() -> bool {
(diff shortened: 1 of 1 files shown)


auser:
Verified on a real Linux + Determinate-Nix host (the actual `nix build`): **all three `runtime_overlay_build` tests pass** — `3 passed; 0 failed`.

- `build_produces_verity_sidecar_matching_overlay_block_geometry` — now matches (1 KiB data blocks + the veritysetup superblock block).
- `build_produces_resolver_compatible_artifact` — passes via the real `install_overlay_into_cache` path (which writes the checksum manifest).
- `build_is_byte_deterministic_for_same_workspace` — unchanged, still passing.

Compile also clean for `x86_64-unknown-linux-gnu` with `-D warnings`. This is the behavioral proof CI can't provide (the test needs host nix).

PR #1897, opened 2026-07-28
fix(mvm-runtime): make virtqueue used index device-owned across drains

auser:
Makes the virtio used-ring index device-owned for the in-house VMM's blk, fs, and vsock devices.

Previously each device re-read the used index from guest-writable RAM at the start of every drain/completion. Now it is carried in per-queue device state, mirroring how `last_avail` is already handled.

Lifecycle semantics:
- `next_used` is initialized to 0 when queue state is created.
- It resets to 0 when QueueReady transitions 0 -> 1 (driver must zero the used ring before enabling) and when the device is reset (Status = 0).
- It is carried across drains within an enabled queue.

For vsock, `build_split_queue` now seeds the virtio-queue `Queue` with both `next_avail` (from `last_avail`) and `next_used` (from `q.next_used`); the per-drain callers no longer read the guest's used index.

Tests:
- Added QueueReady-reset and device-reset lifecycle tests for each device.
- Added cross-drain witnesses for each device: guest scribbles a bogus used index between drains, and completions continue at the device's own next slot.
- The vsock differential tests remain byte-identical for conformant/zero-seeded cases. New "device-owned wins" tests cover the non-zero pre-seed divergent case.

Note on CI/gates: `cargo nextest run -p mvm-runtime` shows one unrelated failure in `workload_runner::claim::tests::claim_guards_spawn_endpoint_keys_the_socket_on_the_given_vm`; this reproduces on main when `MVM_HOME` is a deep worktree path (the socket-dir helper falls back to a hashed /tmp path). All virtio/vsock tests pass.

### DIFF
--- a/crates/mvm-runtime/src/vmm/mod.rs
+++ b/crates/mvm-runtime/src/vmm/mod.rs
@@ -61,10 +61,10 @@ pub(crate) fn validated_queue_size(num: u32) -> Option<u16> {
 
 /// The guest-programmed split-virtqueue ring state a device hands to
 /// [`build_split_queue`]: the three ring base addresses plus the device-owned
-/// resume point in the available ring.
+/// resume points in both the available and used rings.
 ///
 /// Grouped into one type because every virtio-mmio device in this module keeps
-/// the same four values (block as flat fields, fs and vsock per queue) and they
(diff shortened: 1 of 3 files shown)


auser:
Closing as a duplicate — the same fix landed via #1896.

Issue #1851, opened 2026-07-27
Plan 255: vsock-first snapshot, egress & warm-start adoption


auser:
Tracking issue for **Plan 255 — Vsock-first snapshot, egress, and warm-start adoption**
(`specs/plans/255-vsock-first-snapshot-egress-adoption.md`).

Adopt the externally-proven, refactor-aligned techniques (snapshot-first templates with
O(1) CoW clone + fast restore, a warm pool that forks fresh identity-scrubbed children,
richer auditable egress policy, a one-step OCI-image → template build path) while keeping
the vsock seam the **only** guest↔host and egress boundary.

Standing invariants (non-negotiable):
- Vsock is the sole guest↔host and egress boundary — no default NIC/TAP/bridge, no
  transparent TLS MITM, no host eBPF data plane.
- One guest = one workload; a warm parent is forked into a fresh child, never resumed to
  continue a prior workload.
- The guest never sees secrets — credential substitution stays host-side on the typed
  connector path.
- Fork/restore never bypasses admission: a forked/restored instance re-admits or inherits
  a bound, validity-windowed signed `ExecutionPlan` and keeps its dm-verity binding on
  block+ext4 backends.
- The snapshot graph IS the existing chain-anchored checkpoint lineage — no second graph.

Phases:
- [x] Phase 0 — Boundary record & spec update (ADR-025 second prior-art data point +
  vsock-first adoption boundary note).
- [ ] Phase 1 — Snapshot-first storage in `mvm-fs` (content-addressed `SnapshotStore`,
  reflink + sparse-copy fallback, plugged beneath the checkpoint lineage, fail-closed on
  an un-audited/missing/tampered parent).
- [ ] Phase 2 — Warm pool + fork identity hygiene in `mvm-runtime`.
- [ ] Phase 3 — Egress policy enrichment on the vsock seam.
- [ ] Phase 4 — OCI-image template build path.
- [ ] Phase 5 — CLI/SDK surface through `mvm-client`.
- [ ] Phase 6 — Docs, ADRs, close-out.

Phase 0 + Phase 1 land on branch `feat/plan-255-phase01`. Later phases are out of scope for
that branch and tracked here.

auser:
[...]

PR #1906, opened 2026-07-28
feat(plan-255): live Firecracker warm claim — spawn, capture, fork-restore, and the capability flip

auser:
## What & why

Makes the Firecracker warm pool real end to end: a standby parent is booted, captured whole, and released; a claim forks a child by **restoring that saved memory** rather than cold-booting it. Plan 255 Phase 2, follow-on to the merged claim substrate (#1881); parent plan `specs/plans/255-vsock-first-snapshot-egress-adoption.md`, tracking #1851.

This slice is memory-restore based, not cold-boot: Plan 265's guarded restore landed (#1855/#1858/#1859), so `warm_restore_instance_from_path` is live behind the no-NIC device-model guard and the pool can build on it.

## The pieces

- **`parent_checkpoint` on the persisted standby handle** — a spawned parent records the checkpoint a later claim verifies content and lineage against.
- **Boot a clean parent + expose its capture control** (`FcDriver::spawn_standby_parent`, `vm_full_control`). Fails closed when the booted parent's pid can't be read: `pid == 0` is the `is_saved_state()` sentinel, so defaulting to it would leave a live Firecracker process that pool eviction and the stale reaper both skip. Also records the parent's rootfs metadata, which this driver otherwise lacks because it bypasses the runner path that normally writes it.
- **Capture the parent's full state, release it, and complete the child's post-restore handshake** — the capture is backend-agnostic so it lives in the runner; the parent is released either way, since the checkpoint carries its whole state and a failed capture must not strand a guest. The forked child is delivered its generation token and must acknowledge, reseed, and resync its clock, so two children of one parent are not CSPRNG twins.
- **`FcDriver::fork_standby_child`** — delegates to the fork restorer. Refuses an unmaterialized child dir, and refuses a clone carrying no saved memory image: that would quietly degrade every claim into a cold boot and defeat the pool.
- **CLI wiring** — `warm_to_target` never passed a rootfs (a Firecracker parent cannot boot without one) and spawned through the parameterless backend method, so the recorded handle carried no checkpoint; it now threads the launch rootfs and routes through the runner spawn-and-capture. `claim_or_cold` called the fail-closed trait method instead of the runner-backed claim; it now assembles a live context (checkpoint store, snapshot store, signed-chain anchor, name registry) and resolves the parent checkpoint first, cold-booting with a warning when a parent was never captured rather than cloning content nothing can verify. Adds `mvm_core::config::snapshots_dir()`.
- **Capability flip, last** — Firecracker advertises `standby_pool` only now that spawn, capture, and fork are all real. Flipping it earlier would have turned a configured warm pool into a silent cold boot plus warning noise. `snapshots` / `snapshot_capability` / `fs_quick_checkpoint` stay off deliberately: they gate the user-facing snapshot verbs, which this slice does not wire.

## Notable decisions

- `WarmParams.backend` and `claim_or_cold` take `&AnyBackend` rather than `&dyn VmBackend`, since only the enum can reach the runner-backed halves. Two `VmBackend` test stubs could not follow that narrowing, so the hermetic mock grew the `spawn_standby` it was missing plus a `with_failing_spawn` knob mirroring the existing `with_failing_stop`. That is better than the stubs were: it exercises the same `AnyBackend::Mock` arm the CLI dispatch actually routes to.
- The "refuses an unsupported backend" test now proves itself against qemu, which still has no pool.

## Testing

`cargo nextest run --workspace`: **7937 passed**, 14 skipped.

Three pre-existing failures in `mvm-build` (`libkrun_builder::tests::ensure_builder_vm_image_*`) are unrelated and untouched by this branch — its diff against `crates/mvm-build/` is empty, and the last commits to that file are already on `main`. They assert `ensure_builder_vm_image()` errors without a `cmdline.txt`, but it now synthesizes one; that mismatch predates this work.

## Not in this slice

Live validation on a KVM host and the first committed warm-restore number (Task 7 of the plan) — it needs real hardware and stays open. Every other task in `specs/plans/255-live-fc-warm-claim.md` is ticked and verified against the code.


auser:
Closing in favour of \`feat/plan-255-live-fc-warm-claim\`, which carries this slice.

Two sessions worked Plan 255's live-FC tasks concurrently. That branch is a strict superset of this one: it has the same Task 3-6 surface (\`fork_standby_child\`, \`spawn_standby_captured\`, \`parent_checkpoint_for\`, \`snapshots_dir\`, the mock \`spawn_standby\` + \`with_failing_spawn\` knob, the doctor/recovery-matrix updates), plus the post-restore reseed handshake and an \`fc_warm_pool_live\` harness this branch never had — and it is mid-run on the KVM host producing the Task 7 numbers.

Diffed the two: nothing authored here is missing there. The only content this branch has that the other lacks is \`main\` itself (the Content-Length API framing and virtio work), which that branch picks up on its next merge. Landing this too would mean shipping a second implementation of Tasks 5-6 for no gain.

No work is lost by closing.

PR #1910, opened 2026-07-29
feat(mvm-cli): add machine warm-restore verb for vm_full checkpoints

auser:
Adds the user-facing `mvmctl machine warm-restore <CHECKPOINT_ID>` verb that drives the existing Firecracker vm_full warm fork/restore path.

- New `commands/machine/checkpoint.rs` wraps the audited fork helpers from `commands/vm/checkpoint.rs`.
- The verb auto-generates a child VM name, validates the checkpoint class, admits a fresh claim-8 plan, verifies lineage, restores via `FcForkRestorer`, and delivers the post-restore generation token.
- The lower-level `MVM_FORK_VMFULL_FC_EXPERIMENTAL` guard is bypassed internally because the verb itself is the explicit opt-in.
- Exposed the required vm_full fork helpers as `pub(in crate::commands)` and removed the duplicate wrapper from `vm/checkpoint.rs`.
- Added CLI contract tests for help output, `--json`, and graceful failure on a missing checkpoint.
- Updated `machine/tests.rs` to cover the new `WarmRestore` action.

Validation:
- `cargo check -p mvm-cli`
- `cargo test -p mvm-cli commands::machine`
- `cargo test -p mvm-cli --test cli`
- `cargo run -p xtask -- check-claim-catalog` (clean)
- Pre-commit hook passed.

The live KVM warm-pool timing harness was not run because the test box is currently unreachable.


auser:
Superseded by #1911.

This branch predates the warm-pool substrate that landed in #1908 and carries its own parallel version across 13 files, so it now conflicts and would regress what merged. It also sets `standby_pool: true`, which would arm a pool whose claim half has never executed.

#1911 carries this PR's unique work forward onto current `main`: the `machine warm-restore` verb, the `probe_ready` post-restore plumbing, the live warm-pool harness, the snapshot-frame fuzz target, and the recorded KVM measurements. The live harness was rewritten against the current driver API and asserts the capability stays off. Nothing unique here is lost.

PR #1960, opened 2026-07-31
feat(apple-container): stage 4 — agent container-mode activation + real bring-up

auser:
Stage 4 of plan 271 (`specs/plans/271-apple-container-backend.md`): turns the stage-2 inject-and-tear-down flow into a real boot. HVF only — `check-no-vz` stays clean.

## What changed

- **Agent container-mode activation** (`mvm-agentd`): new `MVM_ACTIVATION_FILE` entry. Mode selection is pure and fully tested: PID 1 ⇒ RPC path always (a stray file never self-activates an initramfs boot); !PID 1 + file ⇒ container mode; !PID 1 + no file ⇒ legacy no-op. Container mode: `unshare(CLONE_NEWNS)` + private propagation FIRST (mount moves never perturb vminitd's namespace), then the same `mount_rootfs`/`mount_runtime_overlay`/`mount_volumes`/`pivot`/uid-901-drop sequence as PID 1, via a shared `pivot_common` (fresh proc/sysfs/devtmpfs only when the source isn't a mountpoint). Built by a pure `container_activation_plan` (tested for plain/sealed/in_place) executed step-for-step — ordering has a single source of truth. Malformed/missing file ⇒ typed error, never a panic. PID-1 path is behaviorally identical.
- **Real bring-up** (`AppleContainerBackend`): resolve artifacts → boot HVF supervisor → vminitd `Mkdir`/`WriteFile` of the agent, the serialized `ActivateEnvironment`, **and the host-signer trust anchor** (without it the agent rejects all control connections — discovered in recon) → `CreateProcess`/`StartProcess` → poll the authenticated Ping on vsock:5252 (120s deadline) → marker file + `Ok(VmId)`. Rollback on any failure, unchanged.
- **Real lifecycle**: `wait` rides vminitd `WaitProcess`; `stop` signals the agent via vminitd then terminates the supervisor; `logs` reads the console capture; `list`/`stop_all` key on the `apple-container.marker` convention.
- **device_only resolved**: verity sidecars attach as plain read-only virtio-blk disks in the fixed slot order (rootfs/verity/overlay/overlay-verity, initfs last); a strengthened test pins order + read-only + file-backed for every disk.

## Validation

- 3547 tests pass (8 new agent tests: entry-mode matrix, plan shapes, typed-error paths; 3 new backend tests) across mvm-agentd/runtime/protocol/core.
- Workspace clippy clean; **all 15 xtask gates clean** incl. `check-no-vz` and `check-closure-budget` (279, unchanged — zero new deps); musl compiles for both crates; the musl agent binary is confirmed.
- No live boot here (no Apple kernel/initfs artifact); the gated e2e now asserts the real success path (`MVM_AC_E2E=1 MVM_AC_E2E_ROOTFS=<ext4>` → start → Running → listed → logs → stop → Stopped).
- Security profile keeps all claims `DoesNotHold` and the tier says "pending e2e validation" — fully wired ≠ validated with real artifacts.


auser:
Superseded by #1968 (vminitd-free Apple Container on the HVF runner). The agent's container-mode activation entry was for the dropped vminitd flow; the backend now uses the standard PID-1 universal-initramfs path.

PR #2018, opened 2026-08-01
ci: shorten lint and harden merge queue validation

auser:
## Summary

- replace the workspace-wide `test-support` rerun with targeted package, integration, and example coverage
- move man-page tests onto the already-warm Test runner
- keep the removed MCP server and smoke lane out of CI
- share `mvm-cli` nested Cargo targets across feature fingerprints
- remove branch-scoped multi-gigabyte Cargo target caches and the disk purge they required
- trigger required workflows explicitly on `merge_group: checks_requested`
- cancel only superseded pull-request runs, never exact-commit merge-group validation
- keep manual dispatches independent and make CI's `contents: read` permission explicit
- add structural regression tests for the optimized and merge-queue-safe workflow shape

## Why

GitHub Actions run `30664687885` spent 36 minutes in `Lint (fmt + clippy + policy)`. The tail repeated 8,610 workspace tests under `test-support`, rebuilt embedded binaries for multiple feature-specific `OUT_DIR` values, ran the man-page feature suite on Lint's separate test graph, and uploaded a 4.44 GB target cache that sibling PR and merge-queue refs cannot restore.

A live audit of 50 recently merged queue entries measured 38m26s p50 and 2h14m03s p95 queue latency. Across 101 merge-group jobs, runner wait was 7m59s p50 and 34m16s p95. Regenerated speculative groups added as much as 2h04m21s while obsolete jobs continued consuming hosted-runner capacity.

## Impact

The targeted feature lane covers 2,718 tests plus the required example check and completed locally in 7m27s, down from 13m38s for the sampled workspace-wide command. The 6m26s man-page feature suite no longer sits on Lint's critical path. The removed MCP server and smoke lane remain absent. Required feature, policy, Rust, Nix, and exact-merge-commit coverage remain in place.

The repository merge-queue ruleset was updated separately from build concurrency 5 to 3, minimum-group wait 5 to 0 minutes, and check timeout 60 to 90 minutes. Squash, ALLGREEN, group sizes, bypass policy, required checks, and deployment requirements were unchanged.

## Validation

- `actionlint -shellcheck= .github/workflows/*.yml`
- `cargo fmt --all -- --check`
- all 10 `xtask` workflow-shape regression tests
- `cargo check --workspace`
- `cargo test --workspace -- --test-threads=1` with isolated `MVM_HOME`
- `cargo test -p xtask --features man`
- `cargo clippy --workspace --all-targets -- -D warnings`
- targeted `test-support` CI command sequence
[...]

PR #2020, opened 2026-08-01
ci: requeue transient merge-queue ejections safely

auser:
## Outcome

Adds a bounded API-only recovery workflow for pull requests ejected from the merge queue for transient reasons. Genuine merge conflicts remain out of the queue and receive an actionable comment.

## Safety

- Runs only for `pull_request_target: dequeued`.
- Uses `contents: read` and `pull-requests: write`.
- Never checks out or executes pull-request code.
- Waits for GitHub mergeability computation before deciding.
- Retries at most twice, persisted through timeline label events.
- Removes and reapplies the retry label so every attempt is counted; the prior implementation could leave the counter stuck at one.

## Validation

- `actionlint -shellcheck= .github/workflows/*.yml`
- `cargo test --test merge_queue_requeue_workflow`
- `cargo check --workspace`

The trigger itself requires a live queue dequeue. Its safe failure mode is unchanged: if the recovery workflow fails, the pull request remains out of the queue for human diagnosis.

### DIFF
--- a/.github/workflows/merge-queue-requeue.yml
+++ b/.github/workflows/merge-queue-requeue.yml
@@ -1,11 +1,9 @@
 name: Merge queue auto-requeue
 
 # A PR ejected from the merge queue stops being worked on and stays that way
-# until a human notices. Measured on this repo: 4 of 8 recent PRs were ejected
-# at least once, and #1997 sat outside the queue for nearly three hours between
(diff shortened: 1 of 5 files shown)


auser:
PR checks are green. This is intentionally being held out of the merge queue until #2018 lands, so its synthetic merge-group commit inherits the corrected always-conclusive MCP gate. Enqueueing it now could let it advance if #2018 were ejected.

PR #2047, opened 2026-08-01
fix(initramfs): stop the cold-cache test paying for a real initramfs build

### LINKED ISSUE #2035: attach_universal_initramfs_if_cached_cold_cache_is_non_fatal does a real 128s initramfs build
## What

`attach_universal_initramfs_if_cached_cold_cache_is_non_fatal` takes **128 seconds** on Linux. It is a unit test in `mvm-cli`, and it is by some distance the slowest test in the workspace.

```
$ cargo nextest run -p mvm-cli -E 'test(attach_universal_initramfs_if_cached_cold_cache_is_non_fatal)'
    SLOW [>120.000s] mvm-cli commands::vm::up::runtime_source::universal_initramfs_attach_tests::attach_universal_initramfs_if_cached_cold_cache_is_non_fatal
    PASS [ 128.132s] (1/1)
```

Measured on `origin/main` at e4dcc2128, on an 8-core Linux host. It is not slow on macOS, which is why it has gone unnoticed.

## Why

The test isolates `MVM_HOME` to an empty tempdir and calls `attach_universal_initramfs_if_cached`, asserting only that it does not error. Its own comment names the mechanism:

```rust
// Must never return an error: a cold cache falls back to legacy boot
// (or is warmed automatically on hosts that can build/fetch it).
```

On a host that *can* build, the cold cache takes the second branch and the test performs a **real universal-initramfs build** — 128 seconds of work to assert that a call returns `Ok`. The assertion it makes would be satisfied in milliseconds by the fallback branch.

## Why it matters beyond the inner loop

`mvm-cli` carries four claim-surface files, and the `mutation-witnesses` lane re-runs the owning package's whole suite **once per mutant**. At roughly 150 mutants across those four files, this single test contributes on the order of five hours to every nightly run.

It also makes `cargo nextest run -p mvm-cli` a two-minute proposition for anyone touching the CLI on Linux.

## Suggested direction

The test's subject is the *non-fatal* contract, not the warm path — `attach_universal_initramfs_if_cached_attaches_from_cache` already covers a warm cache with a seeded fixture. So the cold case wants the automatic warm-build suppressed rather than exercised: either an explicit "do not build" knob on the resolver that the test sets, or seeding the cache-miss marker so the fallback branch is the one taken.

Both keep the assertion the test exists to make while removing the two minutes.

## Related

Same shape as the `pull_core` defect fixed in #2023, where `seed_guest_runtime_cache` seeded the guest-binary cache under a key the resolver never read, so three tests silently cross-compiled the guest agent at 55s each. Different cause — there the seed was wrong, here the cold path is deliberate — but the same lesson: a test that means to avoid expensive work should assert that it avoided it.

auser:
Closes #2035.

## What

`attach_universal_initramfs_if_cached_cold_cache_is_non_fatal` took **128 seconds** on Linux — by a distance the slowest test in the workspace, and invisible on macOS where the same path fails a download fast instead of running a cargo cross-compile.

## Why it was slow, and why that made it wrong

The cause is in the test's own comment: a cold cache "is warmed automatically on hosts that can build/fetch it". On a host that can build, asserting a miss returns `Ok` built the entire universal initramfs first — then asserted `Ok` on the **build** path, never on the miss path the test was written for.

So it was not merely slow. It could not have failed for the reason it existed: a regression that made a genuine cache miss fatal would have been masked by the build succeeding.

## The fix

The resolver is injected, the same shape `resolve_or_pull_run_image_with` already uses in this crate. Production is unchanged — all four boot-path callers (`exec.rs` ×2, `checkpoint.rs`, `oci_persist.rs`) still resolve-or-build, which is correct on a first boot. The test supplies a resolver that always misses and asserts **both** halves of the contract: `Ok`, and nothing attached.

## Effect

| | before | after |
| --- | --- | --- |
| the test | 128s | 0.08s |
| `cargo nextest run -p mvm-cli` | ~145s | ~37s |

This matters past the inner loop. The mutation-witness lane re-runs the owning package's whole suite once per mutant, so this test alone was worth hours of every nightly over mvm-cli's four claim-surface files — it was the single largest term in that lane's runtime.

### DIFF
--- a/crates/mvm-cli/src/commands/vm/up/runtime_source.rs
+++ b/crates/mvm-cli/src/commands/vm/up/runtime_source.rs
@@ -329,6 +329,15 @@ pub(crate) fn universal_initramfs_available() -> bool {
     mvm_build::initramfs::resolve_or_seed_from_default_cache(&cache_root, version, arch).is_ok()
 }
 
+/// How the universal initramfs is obtained for a boot.
+///
+/// Injected so the cold-cache contract is testable. In production this is
+/// always [`mvm_build::initramfs::resolve_or_build_local_initramfs`],
+/// which *builds* on a miss — the right thing on a first boot, and two
+/// minutes of cargo cross-compile on any host that can do it. A test of
+/// "a miss is non-fatal" has no business paying for that, and paying for
+/// it is also what made the miss path indistinguishable from the build
+/// path in the assertion.
 pub(crate) fn attach_universal_initramfs_if_cached(
     start_config: &mut mvm_core::vm_backend::VmStartConfig,
 ) -> Result<()> {
@@ -1460,8 +1469,16 @@ mod universal_initramfs_attach_tests {
         }
     }
 
+    /// A cold cache must be non-fatal *and* must attach nothing.
+    ///
+    /// The resolver is injected rather than left to the real one. With the
+    /// real one this test took 128 seconds on Linux: a cold cache is
+    /// "warmed automatically on hosts that can build", so asserting that a
+    /// miss returns `Ok` was paying for a full initramfs cross-compile —
+    /// and then asserting `Ok` on the *build* path, never the miss path it
+    /// was written for.
     #[test]
-    fn attach_universal_initramfs_if_cached_cold_cache_is_non_fatal() {
+    fn attach_universal_initramfs_cold_cache_is_non_fatal_and_attaches_nothing() {
         let mut env = TestEnv::new();
         let dir = tempfile::tempdir().unwrap();
         // HOME moves with MVM_HOME or the cache under test is not cold: the


auser:
Superseded by #2045, which landed an equivalent fix while this was open — closure-based resolver injection, and a cold-cache test that supplies a failing resolver so no build happens.

Verified rather than assumed: on current main the test runs in **0.047s** (was 128s), and main's version additionally asserts the resolver was *called*, which mine did not. Strictly better, so there is nothing here worth rebasing.

Leaving #2035 for the numbers and the reasoning — the interesting part was never the duration but that the test asserted `Ok` on the build path while claiming to test the miss path, so it could not have failed for its stated reason.

PR #2096, opened 2026-08-03
feat(mvm-client): expose normalized verified audit events for UI consumers

### LINKED ISSUE #2082: Expose normalized verified local audit events for UI consumers
## Outcome

Expose a bounded, typed local audit reader that normalizes mvm's audit sources and returns events only after their authenticity and chain integrity have been verified.

## Current gap

mvm can emit and verify several audit chains, and `mvmctl audit` provides operator commands, but UI-safe normalization and verification-aware reads remain coupled to CLI and hostd internals.

A consumer must not parse JSONL directly or accidentally present an unverified entry as trusted activity.

## Scope

- Add a normalized `VerifiedAuditEvent` DTO covering launch, admission, lifecycle, readiness, volume, secret-reference, policy, and refusal activity.
- Extract chain reading and verification behind a reusable mvm-client service.
- Provide bounded newest-first reads with typed cursor, limit, machine, time, and event-kind filters.
- Return typed source-level verification refusals separately from verified events.
- Never return entries from a source whose signature, chain link, scope, or inclusion proof fails verification.
- Define a stable event identity and ordering rule that works across local audit sources without rewriting the underlying chains.
- Sanitize and cap all free-text detail at the service boundary.
- Exclude secret values, environment values, credentials, keys, sensitive host paths, raw policy payloads, and terminal control sequences.
- Keep live HTTP/SSE transport in mvm-studio; this issue owns the verified read contract and normalization.

## Acceptance criteria

- Tests cover valid chains, tampered signatures, broken links, reordered entries, truncated tails, wrong keys, cross-scope entries, pagination, filters, oversized detail, and control-sequence sanitization.
- Tampered or unverifiable sources produce an explicit refusal and no trusted events from that source.
- Event serialization round trips and unknown-field behavior are tested.
- Existing `mvmctl audit` readers reuse the canonical verification/normalization seam where their semantics overlap.
- Fuzz/property tests cover untrusted audit input parsing.
- Workspace tests, clippy, formatting, and the required plan/sprint/status updates are complete.

## Unblocks in mvm-studio

- [mvm-studio#12 — Add machine logs, verified audit activity, E2E coverage, and complete CI gates](https://github.com/tinylabscom/mvm-studio/issues/12)
- [mvm-studio#19 — Epic: Single-host mvm Studio dashboard](https://github.com/tinylabscom/mvm-studio/issues/19)

aneyzberg:
## Summary

Adds `mvm_client::audit::LocalAuditReader` — the verified read contract for mvm's local audit sources, per #2082. A UI consumer (mvm-studio) gets bounded, typed, newest-first reads of normalized audit events, returned **only** after the source chain passes signature, chain-link, and tenant-scope verification. Live HTTP/SSE transport stays in mvm-studio; this PR owns only the read contract and normalization.

### Design

- **Sources.** The per-tenant lifecycle chains (`<mvm_home>/audit/<tenant>.jsonl`) and the per-VM workload chains (`<tenant>.<vm>.workload.jsonl`), discovered via the `mvm-core::config` naming convention. The unsigned legacy operator log is deliberately **not** a source — it carries no authenticity.
- **One verification seam, reused.** No second verifier was written. Lifecycle chains go through `mvm_hostd::supervisor::verify_audit_chain_entries`; workload chains go through a new `mvm_hostd::audit_signer::verify::verify_workload_chain_entries`, extracted from the existing count-only `verify_workload_chain` (which now delegates to it) — mirroring the supervisor's existing count/entries pair. `mvmctl audit verify` is rewired through the same `LocalAuditReader::verify_sources` seam, so CLI and UI consumers cannot disagree on what verifies.
- **DTOs.** `VerifiedAuditEvent { id, kind, timestamp, tenant, machine, event, plan_id, detail }` with a normalized `AuditEventKind` (admission, launch, lifecycle, readiness, volume, secret_reference, policy, refusal, other). All serde shapes pinned with `deny_unknown_fields`.
- **Identity + ordering.** Event identity is `(source, seq)` — the entry's zero-based position in its verified chain; chains are append-only and only verified prefixes are exposed, so identity is stable with no chain rewrite. Newest-first ordering is a total order over (signed timestamp, source rank, chain position) — a pure function of signed content and source identity, so pagination is stable across reads. The typed `AuditCursor` carries that ordering key; pages are strictly-older continuations. Limits: default 100, hard cap 1000.
- **Refusals.** A source failing any check yields a typed `AuditSourceRefusal` (`io | malformed | chain_broken | signature_invalid | scope_mismatch`, with line and sanitized detail) and **zero events — never a partial prefix**. A verified-but-foreign-tenant entry (validly signed, spliced into another tenant's file) is refused via the scope check.
- **Sanitization at the boundary.** ANSI CSI/OSC sequences and control chars stripped, all fields capped, sensitive label keys (secret/token/password/credential/private/env/key/auth) dropped, path-like values redacted, and workload `fields` payloads (raw policy material) never rendered.

## Test evidence

53 new tests (52 in `mvm-client`, 1 in `mvm-hostd`); full suite `cargo nextest run --workspace`: **9196 passed, 0 failed**.

**Verification refusal matrix** (each asserts zero events from the refused source):

| Scenario | Result |
|---|---|
| Valid lifecycle + workload chains | events from both, no refusals |
| Tampered signed byte | `signature_invalid`, sibling tenant chain unaffected |
| Deleted middle line | `chain_broken` at line 1 |
| Reordered entries | `chain_broken` |
| Truncated tail (half-written last line) | `malformed` — whole source withheld, no valid-prefix leak |
| Wrong verifying key | `signature_invalid` on every source |
| Cross-scope entry (lifecycle + workload) | `scope_mismatch` naming both tenants |
| Garbage bytes | `malformed`, no panic |

**Sanitization:** CSI/OSC/ST stripping, bare control chars, cap-with-mark truncation, sensitive-key detection, absolute/`~`/`.mvm` path redaction, raw workload `fields` never rendered, refusal details sanitized.

**Property tests** (hand-rolled seeded loops with `rand`, matching the repo's existing pattern — noted in lieu of a cargo-fuzz target since fuzz targets aren't built in normal CI): 200×2 sanitizer invariants (no control chars survive, caps hold, idempotence) and 50 random-byte chain files through the full reader (never panics, never yields events, refuses non-empty garbage).

**Contract pins:** serde roundtrips for event/refusal/response/cursor, unknown-field rejection, optional-field omission, limit default/clamp, ordering tie-breaks, unparseable-timestamp demotion, pagination walk (no gaps, no duplicates, equals the unpaged read), machine/tenant/kind/time filters, `verify_sources` summaries, kind classification tables.

Closes #2082

## Review follow-up (cb3c5184b)

Addresses the independent review's findings:

- **Operator-log exclusion:** the unsigned `<audit_dir>/secrets.jsonl` operator log is a known non-chain and is excluded from discovery — neither events nor a permanent `malformed` refusal, even for a tenant literally named `secrets`.
- **Dotted tenant names:** tenant-scoped classification now anchors on the requested tenant via `mvm_core::config::workload_audit_vm_name` (the writer's own helper), restoring the old tenant-anchored verify coverage; unscoped classification matches workload stems against known lifecycle tenants (longest match) before the first-dot fallback.
- **No key mint on empty verify:** `mvmctl audit verify` early-returns through the new key-free `discover_source_ids` before loading the host signer, so a read-only verify on a pristine host no longer creates `~/.mvm/keys/host-signer.ed25519`.

Accepted as fast-follows (out of scope here, per review): deepening machine-label path redaction, and redacting path fragments appearing mid-string in label values.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2804,6 +2804,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "chrono",
+ "ed25519-dalek",
  "ext4-view",
  "flate2",
  "hex",
--- a/crates/mvm-cli/src/commands/ops/audit.rs
+++ b/crates/mvm-cli/src/commands/ops/audit.rs
@@ -6,9 +6,9 @@ use clap::{Args as ClapArgs, Subcommand};
 use crate::ui;
 
 use ed25519_dalek::VerifyingKey;
+use mvm_client::audit::{AuditSourceId, AuditSourceKind, LocalAuditReader};
 use mvm_core::user_config::MvmConfig;
-use mvm_hostd::audit_signer::verify::verify_workload_chain;
-use mvm_hostd::supervisor::{SignedEnvelope, verify_audit_chain};
+use mvm_hostd::supervisor::SignedEnvelope;
 use mvm_protocol::merkle::{InclusionProof, SignedAuditRoot, verify_inclusion, verify_signed_root};
 
 use super::super::vm::audit_chain::{AuditEmitter, audit_path_for_tenant, default_audit_dir};
@@ -774,10 +774,19 @@ fn print_chain_line(line: &str) {
 
 fn audit_verify(tenant: &str) -> Result<()> {
     let dir = default_audit_dir()?;
-    let lifecycle_path = audit_path_for_tenant(&dir, tenant);
-    let workload_paths = workload_chain_paths(&dir, tenant)?;
+    // Both chain kinds are signed under the host key; load it once.
+    let signer =
+        host_signer::load_or_init().context("loading host signer to verify audit chain")?;
 
-    if !lifecycle_path.exists() && workload_paths.is_empty() {
+    // The canonical verification seam: the same source discovery +
+    // verification the verified-audit UI reader uses, so the CLI and UI
+    // consumers can never disagree on what verifies.
+    let reader = LocalAuditReader::new(&dir, signer.verifying);
+    let outcome = reader
+        .verify_sources(Some(tenant))
+        .context("enumerating audit sources")?;
+
+    if outcome.sources.is_empty() && outcome.refusals.is_empty() {
         ui::info(&format!(
             "No audit chain found for tenant '{tenant}' under {}. Nothing to verify.",
             dir.display()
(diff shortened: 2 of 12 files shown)


auser:
Does the audit log use standard `tracing`? If so, can we have a tracing consumer that can be configurable? I think this should be a standard tracing consumer. Otherwise we'd have to create a new stack for handling tracing.

aneyzberg:
No — and that's deliberate. This isn't the diagnostics log; it's the chain-signed evidence log behind security claim 8: Ed25519-signed, hash-linked JSONL under `~/.mvm/audit/`, written by the existing `mvm-hostd` emitters. This PR adds only the read/verify seam on top — the reader refuses any source whose signatures or chain links don't verify. Routing writes through `tracing` subscribers would insert an unsigned, reorderable, user-configurable pipeline into the middle of a tamper-evidence chain, which is exactly the class of thing the verifier exists to reject. So there's no new logging stack to maintain here — the signed write path already existed; this PR is read-only over it.

If the ask is observability — seeing audit activity in standard `tracing` output with configurable subscribers — that's fully compatible as a layer: emit a diagnostic `tracing` event alongside each chain append, with the signed chain staying the source of truth the verified reader consumes. Happy to file that as a follow-up issue.

aneyzberg:
Filed the follow-up: #2107 — a fire-and-forget `tracing` mirror of each chain append (dedicated target, same sanitized fields as the verified reader, chain bytes unchanged and never blocked by a subscriber). The signed chain stays the source of truth.

PR #2099, opened 2026-08-03
Extract a reusable local encrypted-volume lifecycle service

### LINKED ISSUE #2080: Extract a reusable local encrypted-volume lifecycle service
## Outcome

Provide a reusable local volume service for mvm-client consumers while preserving the canonical encrypted volume, admission, attachment, lease, and cleanup behavior already implemented in mvm.

## Current gap

mvm has `VolumeBackend`, local/encrypted storage primitives, managed-volume registration, and admitted attachment paths. User-facing orchestration remains coupled to CLI command modules, so another local presentation surface would have to duplicate policy and cleanup logic.

## Scope

- Add strict volume DTOs and an object-safe client/service contract for:
  - listing managed local volumes;
  - creating an encrypted block volume;
  - preparing a typed attachment;
  - acquiring and releasing an attachment lease;
  - reporting encryption, attachment, and lifecycle metadata.
- Implement `LocalVolumeService` in mvm-client by composing the existing volume backend, encryption, registry, admission, and attachment primitives.
- Model volume identity, local source kind, guest path, access mode, capacity, encryption state, attachment owner, and creation time as typed fields.
- Restrict guest mount paths to `/mnt`, `/data`, or `/work`.
- Default attachments to read-only; refuse read-write unless the admitted dev profile permits it.
- Unlock just in time, acquire an exclusive writable lease, bind the attachment into the signed plan, and relock after the final release.
- Make failed launch, stop, detach, and process-crash recovery idempotent and leak-free.
- Refuse missing, locked, changed, unencrypted, tampered, cross-scope, or unsupported attachments before boot.
- Keep arbitrary host-directory mounts and destructive local delete out of this issue.

## Acceptance criteria

- Contract tests cover create/list, read-only attach, permitted read-write attach, duplicate guest paths, exclusive lease refusal, profile refusal, detach/relock, failed-launch cleanup, orphan recovery, persistent restart, and tampered ciphertext refusal.
- Tests prove keys and plaintext never enter DTOs, manifests, argv, logs, errors, audit fields, or browser-facing metadata.
- Existing CLI volume commands consume the same service or underlying orchestration rather than maintaining a second lifecycle.
- A Linux builder-VM/KVM E2E proves create, attach, guest write/read, restart, release, and relock.
- Workspace tests, clippy, formatting, and the required plan/sprint/status updates are complete.

## Related

This is the local reusable-service follow-up to the end-to-end volume work tracked in #2040.

## Unblocks in mvm-studio

- [mvm-studio#13 — Manage encrypted local volumes and attach them during launch](https://github.com/tinylabscom/mvm-studio/issues/13)
- [mvm-studio#15 — Launch transient and persistent local machines from Studio](https://github.com/tinylabscom/mvm-studio/issues/15)
- [mvm-studio#19 — Epic: Single-host mvm Studio dashboard](https://github.com/tinylabscom/mvm-studio/issues/19)

aneyzberg:
# Extract a reusable local encrypted-volume lifecycle service

Extracts the canonical local encrypted-volume behavior out of `crates/mvm-cli/src/commands/vm/volume.rs` into `mvm_client::volume`, so any local surface (the CLI today, mvm-studio next) consumes one policy and cleanup implementation instead of duplicating it.

## What landed

**Service contract (`crates/mvm-client/src/volume/service.rs`).** An object-safe `VolumeService` trait — `recover`, `list_volumes`, `describe_volume`, `create_block_volume`, `unlock_volume`, `lock_volume`, `prepare_attachment`, `detach`, `list_attachments`, `acquire_launch_lease`, `release_owner_leases` — implemented by `LocalVolumeService`, which composes the existing primitives unchanged: `LocalVolumeCatalog` / `VolumeMountRegistry` (mvm-runtime), the AES-256-GCM `MVSE` seal + versioned local master-KEK wrap + DEK content-hash binding (mvm-core crypto), `validate_mount_path` (the `/mnt`, `/data`, `/work` allow-roots), and the launch-time `resolve_for_launch` admission resolution.

**Strict DTOs (`volume/dto.rs`).** `VolumeRecord`, `AttachmentRecord`, `AccessMode`, `AdmittedProfile`, `VolumeSourceKind`, `EncryptionState`, `UnlockPolicy`, `ReleaseOutcome`, plus builder-validated `CreateBlockVolumeRequest`, `AttachmentRequest`, and `LaunchLeaseRequest`. All are `deny_unknown_fields` and secret-free by construction — no wrapped key, ciphertext path, or key bytes can appear in a DTO, and a test proves the serialized record of a real volume carries none of them while the on-disk catalog does.

**Access policy.** Attachments default to read-only; read-write is refused unless the admitted profile is dev-tier (`dev` / `permissive`, mirroring the launch path's writable-share rule). The gate applies both at `AttachmentRequest::build` and again over the full merged volume list in `acquire_launch_lease`, whose builder defaults are fail-closed (`AdmittedProfile::Sealed`, `UnlockPolicy::RequireUnlocked`).

**Lease design.** Reuses and extends the persisted exclusivity catalog (`<mvm_home>/volumes/attachments.json`, keyed by sha256 of the canonicalized host artifact): a block volume attaches to at most one VM regardless of access mode, exactly as before. New on top:

- `LaunchLease` is the RAII guard: conflicts are checked before any mutation, the guard's uncommitted drop releases everything it took, and `commit()` (called only after backend start succeeds) persists the leases for the stop path.
- **Just-in-time unlock**: `UnlockPolicy::JustInTime` decrypts a locked managed volume under the launch's lifecycle flock, records `relock_volume` on the persisted lease, and the *final* release — stop path, failed-launch rollback, or refused resolution — re-seals it and reports it in `ReleaseOutcome::relocked`. A refusal after the JIT unlock (profile gate, duplicate guest path, probe failure, lease conflict) re-seals before surfacing the error, so plaintext never outlives the attempt.
- Crash/orphan recovery is idempotent: leases survive process death (proven with `mem::forget`), a fresh service instance releases them, and a second release is a no-op. Lease records keep the historical on-disk field names, so existing catalogs parse unchanged.

**Refusal ladder (all before boot).** Missing volume/catalog entry, locked volume (unless JIT is requested), changed host path or artifact kind after registration, unencrypted catalog entry, tampered ciphertext (DEK binding mismatch caught before the master key is even loaded), cross-scope lease conflict (already attached to another VM, or twice in one launch), duplicate guest paths, guest paths outside the allow-roots, non-ext4 block artifacts, and host-backed paths whose encrypted backing can no longer be verified.

**Host-encryption seam.** The platform diagnostics (`diskutil`/`lsblk`) live in the CLI doctor, so the service takes an injected `HostEncryptionProbe`; the CLI injects `require_local_volume_host_path_encrypted`, and the default probe fails closed (managed encrypted volumes only). Arbitrary host-directory mounts and destructive local delete stay out of the service contract — the ad-hoc host path survives only as an inherent CLI-compat method behind the probe.

**CLI is now a shim.** `mvmctl machine volume create/unlock/lock/catalog/snapshot/restore/mount/ls/unmount`, the launch-time `merge_registered_volumes_for_launch`, and the stop-path `release_volume_leases_for_vm` all delegate to the service; the former CLI-private lifecycle (~1,300 lines including the snapshot module) is deleted, not duplicated. The clap surface is unchanged. Two intentional output changes: `volume catalog --json` now emits secret-free `VolumeRecord`s (it previously serialized raw catalog entries **including the wrapped DEK bytes**), and `volume ls --json` emits `AttachmentRecord`s.

## Test evidence

- 41 service-level tests in `mvm-client` (contract coverage: create/list/describe, read-only default, permitted read-write, profile refusal, duplicate guest paths, exclusive-lease refusal + rollback, detach, JIT unlock → relock on final release, failed-launch cleanup, refused-resolution re-seal, orphan recovery, persistent restart, tampered-ciphertext and swapped-DEK refusal, interrupted-seal recovery, probe fail-closed + injection, snapshot/restore round-trip + tamper refusal, key-material absence from DTOs/JSON/debug) plus 8 CLI shim tests and 6 DTO tests.
- Full workspace suite, doctests, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and the xtask gates (`check-no-spec-refs-in-comments`, `check-single-home`, `check-test-home-isolation`, `check-audit-positional`, `check-closure-budget`, `check-conformance`, `check-claim-catalog`, `check-mutation-witnesses`, `check-cli-runtime-surface`, dependency gates) pass locally.
- Hermetic BDD volume scenarios pass through the new shim unchanged.

## Linux builder-VM/KVM E2E

The live E2E follows the repo's established pattern: the `s26_volumes` BDD suite's `@live @firecracker` scenarios (gated on `MVM_BDD_LIVE=1` + `/dev/kvm` + `firecracker` on PATH, run by the `bdd` CI lane's harness and skipped cleanly where KVM is absent). The existing scenarios already prove create → attach → guest write/read → restart persistence and read-only guest refusal through the CLI shim (which now runs the service); this PR extends the writable-restart scenario to also prove **release → relock**: after the final `machine stop`, `machine volume lock work` must succeed, which it can only do once the stop path released the lease. Deferred to that lane (cannot run on this macOS host): the actual KVM boots, guest write/read, restart, and the extended relock step. Everything not requiring KVM runs in the hermetic suites above. The service-level JIT-unlock/relock semantics are fully covered by hermetic Rust tests that run on every PR.

## Notes for review

- The launch shim passes `AdmittedProfile::Dev` because its only caller is the persistent named-machine lane, which is dev-accessible for its lifetime (`machine run -t` / `machine shell`); the untouched call site (`up/oci_persist.rs`) does not thread its profile string. Service consumers must pass their profile explicitly and get the sealed default otherwise.
- Registration of a locked volume still refuses with "is locked" (preserving the CLI contract and the BDD scenario); JIT unlock applies at lease time to volumes that were locked *after* registration.

Closes #2080


auser:
What happens if two microvms need to mount the same volume?

aneyzberg:
Depends on the volume kind:

- **Encrypted block volumes (`BlockImage`) are single-attacher.** The second launch is refused *pre-boot* by the lease-catalog check (`check_lease_conflicts` in `crates/mvm-client/src/volume/lease.rs`) with "block volume … is already attached to VM \"X\"; detach or stop that VM" — and this applies to read-only attachments too, because two guests mounting one ext4 block device concurrently isn't crash-safe (even a ro mount can replay the journal). Stopping/detaching the first VM releases the lease (and re-locks a just-in-time-unlocked volume); crash leftovers are cleared by orphan recovery, so a stale lease can't wedge the volume.

- **Directory-backed volumes (virtio-fs: managed-directory, host-backed, ad-hoc) take no exclusivity lease** and can attach to multiple VMs concurrently; each launch independently re-verifies the encrypted host backing. Duplicate-guest-path refusal is per-launch only.

So for genuine concurrent sharing between two microVMs, use a directory volume; if both need the same *block* volume's data at once, snapshot it and attach the copy to the second VM.

PR #2106, opened 2026-08-03
docs(plan-291): the develop → build → deploy path to an attested image

auser:
Plan only — no implementation. Written to be disagreed with cheaply, before any code exists.

## The goal

A developer starts from an OCI image or a Nix flake, iterates until the workload actually works — including discovering what it depends on — then runs one command that seals, hashes, and records the result. If a remote (`mvmd`) is configured that command ships it; if not, they still end up holding a sealed, recorded artifact.

## The main thing this plan does

**It marks what already exists**, because the real risk here is rebuilding shipped machinery:

- Both OCI and Nix bases are already first class
- Dependencies are already declarable, and **hash-pinned lockfiles are already enforced at compile time** — a `uv.lock`/`yarn.lock` entry without an integrity hash is rejected, not resolved
- The sealed dependency volume (SBOM, CVE scan, hash-chained `meta.json`, admission-time refusal of a tampered volume) already ships as claim 11
- `mvmctl build address` already gives a workload a content address

So "declare opencv and get a sealed, hash-pinned, CVE-scanned volume" **works today**.

Only three things are genuinely missing: `mvmctl deploy`, `mvmctl watch`, and capture-from-sandbox. And `reseal_volume` already exists — that is the capture primitive, so the security-sensitive half of "capture what I installed" is written; only the loop around it is missing.

## Stating what "sealed" means, once

The word currently does three jobs, which is why this keeps tangling:

1. **Image contents** — whether console/exec are compiled in. *This distinction goes away:* one image, same bytes.
2. **Run tier** — whether a run may be interacted with. *Follows from (3)*, not from a build flag.
3. **Attestation** — hashed, recorded, traceable. *This is what deploy produces, and it does not go away.*

Sealing is therefore neither removed nor a build variant. It becomes an attestation over fixed bytes.

## Addressing

BLAKE3 for identity, SHA-256 for interop. `blake3` is already a workspace dependency, so this adds nothing to the tree.

BLAKE3 is a tree hash, so a large rootfs can be verified incrementally rather than read end to end. It **cannot** replace SHA-256 where OCI registry digests, cosign, or in-kernel dm-verity specify it — dm-verity has no BLAKE3 support at all, so verified boot keeps its SHA-256 roothash. Artifacts carry both, and anything that pins says which it means. This is the σ/κ split plan 276 already formalised, not a new concept.

## Docs

README section and a web guide (`guides/develop-to-attested`, registered in the sidebar) document the path **as it works today**. The three unbuilt verbs are explicitly marked as designed-not-built with a pointer here, so the docs do not describe commands that do not exist. All doc gates pass (`check-doc-claims`, `check-no-overclaim`, `check-honesty`, `check-deferrals`, `check-machine-doc-guards`, `check-two-surfaces`).

## Open question for the owner

`CLAUDE.md` states that deploy-to-control-plane commands live in mvmd, not mvmctl. This plan assumes `mvmctl deploy` performs seal + hash + record (artifact construction, which mvm already owns) and delegates control-plane conversation, tenants and placement to mvmd. If `mvmctl deploy` is meant to own the control-plane side directly, that is an architectural change and should be decided before WS1 starts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,35 @@ inside the microVM.
 
 ---
 
+
+### From dev loop to attested image
+
+The three routes above are the *start* of one path: pick a base, iterate until
+the workload actually works, then end with a sealed, hashed, recorded artifact.
+
+What that looks like today:
+
+```bash
+mvmctl build compile app.py --out ./out   # declared deps; lockfiles must be hash-pinned
+mvmctl machine build --flake ./out        # installs deps into a SEALED volume
+mvmctl deps inspect                       # SBOM + CVE + hash-chained meta, no VM needed
+mvmctl machine run --entrypoint --flake ./out
+```
+
+Dependencies land in a sealed volume rather than in the image: hash-locked
+content, an SBOM, a CVE scan, and a hash-chained `meta.json`. The supervisor
+verifies that volume before launch and refuses a tampered one, and `--prod`
+fails closed on high/critical findings or a stub SBOM. A lockfile entry with no
+integrity hash is rejected at compile time.
+
+`mvmctl deploy`, `mvmctl watch`, and capturing a dependency you installed inside
+a dev sandbox are **designed but not yet built** — see
+[plan 291](specs/plans/291-develop-build-deploy-attested.md). Until they land,
+the declared route above is the supported way to get a dependency into an
+attested workload.
+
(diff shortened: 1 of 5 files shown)


auser:
Tracking: #2123 (Plan 291 umbrella). The design/specification remains the source of scope for WS1–WS4; child issues #2124–#2127 track implementation and acceptance.

PR #2108, opened 2026-08-04
Derive agent-verb tier from run shape

auser:
## Summary

- Derive the attenuated agent-verb grant from the run shape: baked-entrypoint, non-dev runs qualify; PTY and ad-hoc argv runs remain DevOnly.
- Route all agent-verb restriction call sites through the shared helper and remove the image sidecar sealed bit from that decision.
- Record the landed prerequisite in Plan 291, `SPRINT.md`, and `REFACTOR-STATUS.md`.

## Why

The previous grant was keyed on the image's sealed bit. As a result, an unsealed OCI image remained interactively reachable and a sealed image could not vary by run. The tier is a property of the admitted run, not the artifact.

## Validation

- `rustup run nightly cargo fmt --all -- --check` — passed.
- `cargo build -p mvm-cli --all-targets` — passed.
- `git diff --check` — passed.
- Workspace nextest, focused test, workspace Clippy, and the xtask doc gate were attempted but could not complete because the host entered a confirmed zero-CPU filesystem/build stall; they are not reported as green.

### DIFF
--- a/crates/mvm-cli/src/commands/vm/agent_verbs.rs
+++ b/crates/mvm-cli/src/commands/vm/agent_verbs.rs
@@ -19,19 +19,19 @@ pub(crate) fn image_is_sealed(rootfs_path: &std::path::Path) -> bool {
         .unwrap_or(false)
 }
 
-/// Whether a run should receive an attenuated default agent-verb grant. Only a
-/// baked-entrypoint run, on a non-dev profile, of a **sealed** image qualifies:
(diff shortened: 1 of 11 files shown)


auser:
Tracking: #2123 (Plan 291 umbrella) and #2127 (WS4 acceptance/conformance closure). The implementation is merged; the issue remains open for deploy-record integration and end-to-end security witnesses.

Issue #2124, opened 2026-08-04
Plan 291 WS1 — mvmctl deploy attestation and local/remote artifact handoff


auser:
Parent: #2123

## Scope

Implement `mvmctl deploy` as the attestation and artifact-construction path. The command must work without a configured remote and must delegate remote shipping only when mvmd is configured.

## Acceptance checklist

- [ ] Resolve and verify the built workload artifact and its dependency-volume binding.
- [ ] Refuse tampered dependency volumes using `verify_sealed_volume`.
- [ ] Produce a durable local deploy record containing:
  - [ ] workload semantic address / `ir_hash`
  - [ ] mvm-native BLAKE3 artifact identity
  - [ ] OCI/cosign/dm-verity SHA-256 interop digest where applicable
  - [ ] environment pin
  - [ ] sealed dependency-volume hash and manifest digest
  - [ ] SBOM/CVE verdicts and provenance
- [ ] Make the local sealed artifact and record discoverable through existing config/cache helpers.
- [ ] If a remote is configured, upload the recorded artifact through the mvmd contract with authentication and audit events.
- [ ] If no remote is configured, finish successfully with a clear local-only result.
- [ ] Add CLI, tamper, no-remote, remote, replay/idempotency, and audit tests.
- [ ] Update plan/sprint/refactor status and land through a PR.

## Dependencies

- Uses the existing build, sealed-volume, address, signing, and environment-pin contracts.
- Blocks #2126 (WS4 tier finalization) and the WS3 deploy-record integration.
- Related implementation worktree: `feat/plan291-ws1-deploy`.

Security note: this command constructs and records artifacts; it must not silently broaden runtime interactive access or introduce host-side package installation.

auser:
WS1 contract review:\n\nThe local deploy/attestation path is present on feat/plan291-ws1-deploy, including BLAKE3 plus SHA-256 artifact digests and sealed dependency-volume verification.\n\nRemote shipping is not yet safe to complete in mvm alone. The current mvmd gateway contract exposes artifact list/get/promote/archive routes, but no authenticated artifact upload or workload-ingest endpoint matching MvmdClient::ship. The branch still contains a logging stub, so the WS1 issue remains open until mvmd publishes the upload contract and its authentication/idempotency/audit semantics.

auser:
[...]

Issue #2123, opened 2026-08-04
Plan 291 — Develop → build → deploy an attested workload image


auser:
## Objective

Drive Plan 291 through production-ready completion: one developer path from an OCI image or Nix flake, through development and dependency discovery, to a sealed, hash-addressed, recorded artifact that can be shipped remotely or retained locally.

Security and UX requirements:

- BLAKE3 for mvm-native artifact identity; SHA-256 where OCI, cosign, or dm-verity require it.
- No live package installation or interactive agent verbs on production runs.
- Dependency volumes are lockfile-pinned, SBOM/CVE-scanned, atomically resealed, and admission-verified.
- Local deploy is a supported outcome when no remote mvmd is configured.
- Every workstream has focused tests, CI gates, status-doc updates, and a reviewable PR.

## Timeline / checklist

- [x] **WS1 — deploy and attestation** — #2124
  - [x] Seal the built/source deploy artifact and write the local deploy record.
  - [x] Record workload address, BLAKE3 identity, SHA-256 interop digest, environment pin, dependency-volume hash, and SBOM/CVE verdicts.
  - [x] Ship to mvmd when configured; retain the sealed artifact locally otherwise.
  - [x] Refuse tampered dependency volumes.
  - Completed by merged PR #2131; documented in merged PR #2138.
  - Blocks: WS4’s final tier signal.

- [x] **WS2 — watch** — #2125
  - [x] Rebuild on local source changes.
  - [x] Skip no-op rebuilds using the workload address and input identity.
  - [x] Surface changed inputs and build cost.
  - Completed by merged PR #2109; required queue checks passed.

- [x] **WS3 — dependency discovery and capture** — #2126
  - [x] Install dependencies during development.
  - [x] Export bounded content and matching sidecars from a running dev sandbox.
  - [x] Reseal atomically, refresh the lockfile index, and emit the canonical declaration.
  - [x] Prove tamper, path, size, posture, lockfile, SBOM, and CVE failure paths.
  - Completed by merged PR #2132; documented in merged PR #2138.

- [~] **WS4 — tier follows attestation** — #2127
  - [x] Derive the agent-verb grant from the admitted run shape rather than the image sidecar.
  - [~] Bind the tier to the exact attested boot artifact — #2141.
  - [~] Retire the interactive feature fork after the explicit production security decision.
  - [x] Add conformance coverage proving a complete ProdSafe grant refuses Exec and ConsoleOpen.
[...]

PR #2131, opened 2026-08-04
feat(plan291): add local workload deploy attestation

auser:
## Summary

- add mvmctl deploy for local sealed artifact construction and recording
- compute BLAKE3 identity and SHA-256 interoperability digest in one streaming pass
- record workload IR hash, optional environment pin, and verified dependency-volume evidence
- retain the local deploy artifact and record when no remote is configured
- ship the exact record plus bundle to an authenticated mvmd endpoint when configured
- fail closed when a configured remote lacks MVM_MVMD_API_KEY or uses an unsafe cleartext endpoint

## Remote contract

- client posts multipart record + bundle to POST /api/v1/deploy-artifacts
- bearer authentication is required
- mvmd binds BLAKE3, SHA-256, and size, and returns a stable revision/status
- replay returns the original revision; digest mismatch and unauthorized requests fail
- server-side implementation is mvmd PR #206, which must merge before this client path is accepted

## Security and behavior

- dependency volumes are verified with verify_sealed_volume before they enter the record
- tampered volume sidecars are refused
- remote shipping cannot report success without authenticated transport
- the bundle remains available locally even when remote shipping is not configured

## Validation

- rustfmt and git diff --check passed
- cargo test -p mvm-sdk deploy --lib --no-fail-fast — 10 passed
- cargo check -p mvm-sdk --lib passed
- cargo clippy -p mvm-sdk --lib -- -D warnings passed
- the commit hook's full workspace check was blocked by the host Cargo build-script stall; CI must provide the definitive workspace result

Tracks #2124 and #2123.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3069,6 +3069,7 @@ version = "0.18.0"
 dependencies = [
  "async-trait",
  "base64",
+ "blake3",
  "flate2",
  "globset",
  "hex",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2640,6 +2640,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "mime"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
+
+[[package]]
+name = "mime_guess"
+version = "2.0.5"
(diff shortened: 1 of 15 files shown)


auser:
CI found the mvmctl default closure at 278/275 because reqwest multipart pulled mime, mime_guess, and unicase. Commit 5db899111 replaces the dependency feature with a bounded hand-built multipart body, removes those lockfile edges, and returns the closure to 275. The host focused test remains blocked by the known proc-macro2 Cargo build-script stall; CI is the source of truth for the rerun.

Issue #2126, opened 2026-08-04
Plan 291 WS3 — sandbox dependency install, bounded capture, and atomic reseal


auser:
Parent: #2123

## Scope

Finish and land dependency discovery from a development sandbox. The declared install route and the live capture route must converge on the same sealed, lockfile-pinned dependency volume contract.

## Acceptance checklist

- [ ] Keep `mvmctl deps install` lockfile-pinned and development-gated.
- [ ] Capture a running development VM only after positively classifying it as dev posture.
- [ ] Export content through bounded agent filesystem reads; reject unsafe paths, symlinks, special files, listing truncation, and byte/file-limit exhaustion.
- [ ] Export matching SBOM, fetch-log, and CVE sidecars and reseal through the existing atomic path.
- [ ] Verify the prior volume before mutation and publish the new hash/index atomically.
- [ ] Verify lockfile SHA-256 and derived lockfile hash before emitting the canonical `Dependencies` declaration.
- [ ] Prove tamper, missing sidecar, lockfile mismatch, production-posture, race, and cleanup failure paths.
- [ ] Run workspace check/test/clippy/doc/supply-chain gates on a healthy builder and record results.
- [ ] Update plan/sprint/refactor status and land through a PR.

## Dependencies

- Can be developed independently.
- Final artifact-record integration depends on #2124.
- Related implementation worktree: `feat/plan291-ws3-declaration`.
- The current branch contains the uncommitted implementation; do not close this issue until it is committed, reviewed, merged, and fully verified.

auser:
WS3 implementation update:

- mvmctl deps install covers the declared lockfile-pinned development install route.
- mvmctl deps capture-live is implemented on feat/plan291-ws3-declaration; it checks the VM is development posture before guest access, exports only bounded regular files through the agent FS RPC, refuses symlinks and special files, and hands the staged artifacts to the existing atomic reseal path.
- Formatting, Cargo metadata, and diff checks pass.
- The focused Cargo compile remains pending: this host is saturated by unrelated Cargo jobs and stalls while compiling ring; no code diagnostic has been produced.

Keeping the issue open until the compile, tests, clippy, and CI gates are green.

auser:
WS3 focused validation is now green on feat/plan291-ws3-declaration (uncommitted):

- nightly rustfmt check, git diff --check, cargo check -p mvm-cli --all-targets
- cargo build -p mvm-cli --all-targets
- cargo clippy -p mvm-cli --lib -- -D warnings
[...]

PR #2132, opened 2026-08-04
feat(plan291): capture dependencies from development sandboxes

auser:
## Summary

- add lockfile-pinned development `mvmctl deps install`
- add bounded `mvmctl deps capture-live` through agent filesystem RPC only
- reject production or non-running VMs, unsafe guest paths, symlinks, special files, listing truncation, races, and byte/file-limit exhaustion
- reseal atomically through the existing sealed-volume contract
- verify lockfile SHA-256 and derived lockfile hash before emitting canonical `Dependencies`
- keep cache/index and declaration publication fail-closed

## Security and behavior

- no guest shell or host share is used for live capture
- guest files are stat/read/stat checked and read with symlink following disabled
- live capture requires positive development posture and `Running` machine status
- declaration output cannot overwrite the lockfile or land inside the dependency cache
- existing volume and sidecars are verified before capture mutation

## Validation

- nightly rustfmt check
- `git diff --check`
- `cargo test -p mvm-cli --lib commands::deps --no-fail-fast` — 33 passed
- `cargo test -p mvm-build app_deps --lib --no-fail-fast` — 21 passed
- `cargo clippy -p mvm-cli --lib -- -D warnings` — passed
- `cargo build -p mvm-cli --all-targets` — passed
- pre-commit workspace clippy — passed

Live capture still requires a running development VM for an end-to-end witness; full workspace/Linux-builder/supply-chain gates remain required before closure.

Tracks #2126 and #2123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,12 @@ What that looks like today:
 ```bash
 mvmctl build compile app.py --out ./out   # declared deps; lockfiles must be hash-pinned
 mvmctl machine build --flake ./out        # installs deps into a SEALED volume
+mvmctl deps install --lockfile uv.lock --language python  # dev install + seal
+mvmctl deps capture-live HASH --vm dev-vm \
+  --guest-content /mvm/deps/content \
+  --guest-sbom /mvm/deps/sbom.cdx.json \
+  --guest-fetch-log /mvm/deps/fetch.log \
+  --guest-cve /mvm/deps/cve.json              # export + reseal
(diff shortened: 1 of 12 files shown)


auser:
CI diagnosis/fix: the first run failed in the no-default-features xtask man-page lane because deps install left the success-path bindings reachable after the fail-closed builder-vm bail. Commit a307f2560 moves installation into cfg-selected helpers and adds the missing-builder regression test. The branch is now rebased onto current main and its fresh Invariant/Lint/Test run is queued/in progress.

PR #2139, opened 2026-08-04
Workload stream plane: live stdout/stderr, and a grant-gated stdin channel

auser:
Streaming stdout/stderr for every workload — live while it runs and readable after it exits — plus a default-deny channel for feeding a running workload's stdin.

Design and sequencing: `specs/plans/283-workload-stream-plane.md`. Posture: `specs/adrs/035-workload-stream-plane.md`.

## What works

`mvmctl logs -f <vm>` streams live on Firecracker, libkrun, HVF and QEMU, and `mvmctl logs` reads after exit. Guest stdout and stderr stay separable over vsock; the console covers boot and post-agent-death, so a guest that panics before its agent starts is still visible. Output is redacted at one seam, hash-chained, and sealed to an RFC-6962 root anchored in the chain-signed audit log. `mvmctl invoke` streams as the workload runs and returns the caller's own bytes unmasked, reporting when the recorded copy differs.

`mvmctl machine run --entrypoint --stdin -` streams stdin into a running workload; EOF closes the workload's stdin. Omit the flag and piped stdin is still auto-detected and sent as one payload, unchanged.

## Two sources, one stream

The vsock path goes dark before the agent starts and after it dies. The console path cannot separate stdout from stderr. Both feed one host-side broker, tagged by source. Ordering within a source is exact; between sources it is best-effort by host arrival stamp, because the two traverse different paths and a total order is not something the transport can deliver.

## Security posture

Claim 15 moves from enforced-by-absence to enforced-by-policy, and ADR-001 says so plainly rather than asserting parity. Before this branch there was no production host-to-guest input path at all; now there is one, gated by a signed grant.

What still holds by construction: input bytes reach a pipe of an already-running process, so they cannot select a program, alter argv or env, or spawn anything. What is policy: whether stdin is delivered at all, refused without `host.stream.v1` on the signed plan.

Claim 17 ships at `Preview`, not `Shipped`, with its limits stated — `InputGate::bind` has no production caller, so the cross-frame secret scan is inert on every real VM. The ledger says which limits closed and which did not.

Also fixed: `AdmittedPlan`'s fields are now private, so an input grant cannot be forged by an in-process struct literal; the `--prod` shell-entrypoint refusal reads a real entrypoint off the image sidecar and fails closed on one it cannot resolve; and a workload cannot forge an agent-authored control record, which would otherwise let it excise its own output from a verified transcript.

## Reconciliation with main

`#2093` fixed `machine logs` on macOS host-side in `observe.rs`. This branch replaced that path with broker → transcript → console. The mechanism here subsumes it, so plan 289 is marked superseded rather than reverted, and main's `mvm_core::config` path-resolution fix is kept. `--lines` semantics are preserved exactly, which one of main's tests caught when they were not.

`--stdin` had been deliberately removed from `machine run` in favour of TTY auto-detection. Auto-detection is unchanged; the flag returns only for streaming, because a signed grant cannot be inferred from "stdin happens to be a pipe" without making default-deny depend on the shape of the caller's shell. That guard now locks the property rather than the flag's absence.

## Verification

12 xtask gates green, including `check-stream-redaction-seam`, both vsock-egress gates, `check-guest-agent-runtime-free`, `check-core-runtime-free`, and `check-claim-catalog` resolving all 55 witnesses across 17 claims. Workspace builds all targets; clippy `-D warnings` and fmt clean.

Two environmental test exclusions, both verified rather than assumed: `check_cmd_rustup_on_host` fails because rustup is not installed on the dev machine, and seven `mvm-runtime` spawn tests carry a 10s handshake budget that concurrent builds blow — all seven pass in under 1.3s each on an idle machine.

## Known limits, stated rather than implied

- The console fallback is unredacted while the transcript is redacted.
- The cross-frame secret scan is inert until `InputGate::bind` has a production caller.
- No VM-to-VM fan-out yet; that is `specs/plans/292-fleet-stream-fan-out.md`, with its four design decisions settled.

## Scope

22 tasks, each reviewed with fix loops to green. Six were added during execution, every one a seam no original task owned — the RPC re-buffering what the pump had unbuffered, chunk batching, nothing constructing the broker, nothing feeding the entrypoint source, transcripts sealing only for foreground runs, and no route between the input gate and the guest sink.


auser:
CI has never run on this branch — the check panel is empty rather than red, and there are zero workflow runs for the head SHA.

Cause is the merge conflict, not a workflow problem. `pull_request` events are tested against `refs/pull/N/merge`, and with 42 conflicting files GitHub cannot compute that ref, so no workflow is dispatched. `ci.yml` has no branch or path filter and Actions is healthy on other PRs.

Resolving the conflicts against main (21 commits of drift) starts CI on its own. Merge plan, invariant checklist, and the open `replay_vectors` question are in #2152.

Issue #2147, opened 2026-08-04
IPv6: measure the CONFIG_IPV6 guest-kernel delta, then decide unconditional vs image variant


auser:
Part of #2111. Child of #2116, which landed the host side.

The host admits and carries IPv6 as of `ed17103c4`. The workload kernel does not build with `CONFIG_IPV6`, and no in-guest v6 address is configured, so **a guest cannot originate v6 traffic**.

## The measurement is the deliverable

ADR-038 requires this delta be measured rather than assumed. Work is in flight to shrink guest kernels to the virtual hardware floor, and enabling `CONFIG_IPV6` silently reverses part of that if landed carelessly.

Build both, and record:
- image-size delta
- boot-time delta

## The decision it feeds

- If the delta is the expected few hundred kilobytes: land it unconditionally, recorded as a known accepted cost rather than an unexplained regression.
- If it is material: ADR-038 says IPv6 becomes a **guest-image variant** rather than an unconditional default.

Either way the number goes in the plan, so the choice is defensible later.

## Also in scope

In-guest v6 address configuration alongside the existing v4 bring-up, so a guest with the option compiled in can actually use it.

## Navigation

**Epic:** #2111
**Parent:** #2116 — host side, landed
**Related:** the guest-kernel floor work this must not silently reverse

auser:
**Located the exact change site**, so this is a measure-and-decide session rather than an investigation.

The workload kernel config is `nix/images/kernel/workload.nix`, built through `base.mkKernel` with an `extraEnables` list:

```nix
base.mkKernel {
  extraEnables = [ "MD" "BLK_DEV_DM" "DM_VERITY" "VIRTIO_FS" "FUSE_FS" "TUN" ]
    ++ pkgs.lib.optionals optimizeForSize [ "CC_OPTIMIZE_FOR_SIZE" ];
```

`CONFIG_IPV6` appears **nowhere** in `nix/images/kernel/` today — confirmed by grep. The change is adding `"IPV6"` to that list.

Note `TUN` is already there and is documented as **workload-only** ("the builder VM has no tunnel"), which is the right precedent: IPv6 belongs in the same workload-only place, not in shared `base.nix`, since the builder VM does not need it.

### Why this cannot be eyeballed

[...]

Issue #2121, opened 2026-08-04
WS9: validate the userspace datapath on WSL2


auser:
> **BLOCKED — needs hardware that does not exist yet.**
> Requires a live Windows/WSL2 host. No code work remains on the mvm side; the
> datapath carries no `target_os` gating and its Linux path is already built and
> tested on a real Linux/KVM host. The three-step test plan below is
> pre-specified so this is a short session once a host exists.

Part of #2111. **Order: 10.**

Validate `l3-vsock` and the userspace socket datapath under WSL2 on a real runner.

## Why it is separate

WSL2 is the one target in this epic with no validation behind it. The userspace datapath should work there by construction — it needs no privilege and no TUN — but "should work by construction" is a hypothesis, and this epic's standard is a live witness rather than an argument.

Currently untestable here: no live Windows host is available. The work is documented and scheduled rather than claimed.

## Scope

- Run the unprivileged end-to-end suite on WSL2 and record the result.
- Confirm backend selection resolves to the userspace datapath, and that the fallback diagnostic names the reason.
- Record what does **not** work, if anything, as a capability the backend declines rather than a defect to hide.

---

## Navigation

**Epic:** #2111
**Depends on:** #2112, and on backend selection landing
**Previous:** #2120 · last in sequence

auser:
**Status: blocked on hardware, not on code. Here is exactly what is and is not established, so this is actionable the moment a host exists.**

### What is already established, without a WSL2 host

WSL2 is Linux, and the userspace datapath carries **no `target_os` gating whatsoever** — it is platform-neutral by construction. `select_host_datapath()` probes for TUN and falls back to `UserspaceSocketDatapath` when the probe fails, which is the same path an unprivileged Linux host takes.

That path is exercised today:
- `cargo zigbuild --target x86_64-unknown-linux-gnu -p mvm-hostd --all-targets` builds it, including the Linux-gated test files.
- The full suite has been run on a real Linux/KVM host (6.8.0 x86_64) against this branch's tree.
- The readiness registration was mutation-tested on **both** kqueue and epoll.
- `epoll_wait` empty-return cost was measured there directly (~480 ns; no macOS-style pathology).

So the claim "the userspace datapath works on Linux userspace with no privilege" is supported. WSL2 inherits that by being Linux.

### What genuinely cannot be established without a live host

[...]

Issue #2120, opened 2026-08-04
WS8: mvmd-facing node-control API — the mvm side


auser:
Part of #2111. **Order: 9.**

The node-control surface a fleet control plane calls to manage networking on a single host.

## Boundary

This issue covers **the mvm side only**: the surface mvm exposes and the types it speaks. Fleet orchestration — tenants, pools, instances, scheduling, and the admission decisions that belong to a control plane — lives in the mvmd repository and is tracked there. The split follows the existing one: image, entrypoint, source and dependencies are mvm's; resources and network policy are mvmd's.

Landing the mvm half without the mvmd half is expected and fine; the surface is useful to a local operator on its own.

## Requirements

- Typed, versioned, and `deny_unknown_fields` on every wire type, so an unexpected field fails closed.
- Authenticated, with a caller bound to the VMs it owns.
- Reports capability honestly: a node that cannot serve ICMP or arbitrary IP says so, and admission refuses rather than degrading silently.

---

## Navigation

**Epic:** #2111
**Depends on:** #2112
**Related:** #2119 · mvmd-side counterpart tracked in the mvmd repository
**Previous:** #2119 · **Next:** #2121

auser:
**Complete for the scope this issue names — the mvm side.**

`mvm_hostd::nodectl` implements the node-control surface: typed `deny_unknown_fields` wire types, bounded tables that drop rather than evict, and ownership as a uid comparison taken from the connection's peer credential rather than any field in the message. A caller is refused a machine it does not own, and a listing carries only its own.

**Mutation-proved:** forcing `CallerIdentity::owns` to `true` reddens five tests. It binds no listener of its own — the server takes an already-accepted stream — so it adds no server to a substrate that has an architectural gate against them.

ADR-041 answers the question this workstream had to settle first: **the cross-node issuer is a fleet concept and belongs to the control plane; the node's job is to verify what the issuer says.** That is why this half-unblocks #2119 rather than unblocking it, and why no issuer was invented here — a key scoped to a node pair would still be a second trust root, just one with a smaller blast radius.

The fleet-orchestration half lives in mvmd and is out of scope for this issue by its own title.

Issue #2119, opened 2026-08-04
WS7: node-to-node transport for cross-host VM traffic


auser:
> **BLOCKED — cannot be completed in this repository.**
> Needs a cross-node issuer both nodes trust. ADR-041 places that with the
> control plane, so the remaining half lives in **mvmd**, not mvm. The mvm-side
> verification seam has landed (#2120). Design is settled in ADR-040; the four
> unblocking conditions are listed there. Nothing further can be built here
> until the issuer exists.

Part of #2111. **Order: 8.**

Carry admitted traffic between VMs on different hosts, so an `l3-vsock` workload can reach a peer that is not local.

## Scope

A transport between nodes that preserves the properties the single-host datapath establishes. The hard part is not moving bytes; it is that the policy decision, the audit record, and the identity binding must remain intact across a hop that leaves the machine.

## Requirements

- The remote end is authenticated, and a node accepts flows only for VMs it actually hosts. A node asking about another node's VM gets a refusal, not an answer.
- Default-deny survives the hop: crossing a node boundary is not an implicit admission, and the destination is authorized against the same policy as a local flow.
- The audit chain still records what happened, on the node that made the decision.
- No workload NIC appears anywhere in this — the transport is host-to-host, and the guest still reaches it only over vsock.

---

## Navigation

**Epic:** #2111
**Depends on:** #2112
**Related:** #2120 (a control plane drives this)
**Previous:** #2118 · **Next:** #2120

auser:
**ADR-040 written; implementation deliberately declined. Blocked on #2120.** Commit `204799538`.

This is the ADR-039 outcome again: a new trust boundary between hosts deserves a design record before code, and three of WS7's four required invariants cannot be preserved with what exists today. **I verified all four blockers myself rather than accepting the reasoning:**

| Blocker | Verified |
|---|---|
| No cross-node trust root | `lease.rs` signs under `key_id: "local-node"`; `verify_lease` returns `LeaseError::WrongNode` before reading the grant |
| Addresses don't name VMs across nodes | `PoolAllocator::allocate(&mut self)` — **zero** occurrences of any node identifier in `alloc.rs`; two nodes issue the same /30 |
| Inbound admission has no source | `IngressTable::admits(&self, protocol: u8, guest_port: u16)` — protocol and port only |
| Audit chain doesn't reach netd | no `AuditEmitter` anywhere under `crates/mvm-hostd/src/netd/` |

[...]

PR #2150, opened 2026-08-04
Unprivileged L3 networking: the userspace socket datapath

### LINKED ISSUE #2116: WS3: IPv6 as a first-class address family
Part of #2111. **Order: 5.** Blocked on #2113.

Support IPv6 with the same default-deny posture, policy projection, audit chain, and capability honesty as IPv4. Design: ADR-038.

## Why it is blocked

smoltcp's `proto-ipv6` is already compiled in and is inert for exactly one reason: the admission guard refuses any non-V4 destination. It is **not** inert because the interface lacks an IPv6 address — `process_hopbyhop` runs before any address check, and the no-address drop is not total (`::1` passes via `is_loopback()`, and `ff02::1` and solicited-node addresses pass `has_multicast_group`, reaching `process_nxt_hdr`, ICMPv6/NDP/MLD, and socket dispatch).

Relaxing the guard therefore hands a hostile guest a live NDP/ICMPv6/hop-by-hop parser. That parser must be fuzzed first. This issue may not land its admission change until #2113 exists and runs, with IPv6 corpus entries.

## The decision that carries the weight

IPv6 gives **four** ways to write an IPv4 address, and every one of them reaches `169.254.169.254`:

| Form | Prefix |
|---|---|
| v4-mapped | `::ffff:0:0/96` |
| v4-compatible (deprecated) | `::/96` |
| NAT64 well-known | `64:ff9b::/96` |
| 6to4 | `2002::/16` |

v4-mapped is the famous one; NAT64 and 6to4 are the ones that get forgotten, and a NAT64 gateway on the host network makes the third genuinely routable.

So the v6 class check does not enumerate v6 hazards and hope the list is complete. It **extracts first** — one `embedded_v4(addr) -> Option<Ipv4Addr>` helper — and any address it yields goes through the **entire existing v4 class check**, unchanged. One function, one place to get right, and an encoding nobody anticipated fails closed because the surrounding posture is deny-by-default.

This must land in the same change as the guard relaxation. The destination-integrity assertion compares addresses in canonical form, which collapses exactly the distinction an embedded-v4 bypass exploits — so the class check is the only line of defence here, not a backstop.

## Native v6 policy mirrors v4

`::1`, `::`, `fe80::/10`, and `ff00::/8` are mandatory denies; `fc00::/7` is deny-unless-admitted, the direct analogue of RFC1918. Link-local is mandatory rather than opt-in because `fe80::/10` is where NDP neighbours and routers live — a guest reaching it reaches the host's own interfaces.

## Capability honesty

`arbitrary_ipv6` stays **false**: a userspace socket gateway cannot put an arbitrary v6 packet on the wire, exactly as it cannot for v4. Add `ipv6_flows: bool` for the genuine distinction — whether the backend carries v6 TCP/UDP flows at all — and set it true. A plan needing arbitrary v6 packets stays refused, with the shortfall naming `arbitrary_ipv6`.

## Guest kernel

Needs `CONFIG_IPV6` plus in-guest v6 address configuration. Work is in flight to shrink guest kernels to the virtual hardware floor, so measure the image-size and boot-time delta and land it as a known accepted cost. If the delta proves material rather than the expected few hundred kilobytes, IPv6 becomes a guest-image variant instead of an unconditional default.

---

## Navigation

**Epic:** #2111
**Blocked by:** #2113 — hard gate; the admission change may not land until the ingress fuzz target exists and runs with IPv6 corpus entries
**Depends on:** #2112
**Previous:** #2115 · **Next:** #2117

auser:
Makes `l3-vsock` work on a host where mvm holds no privileges. Closes the bulk of epic #2111.

## What this is

`l3-vsock` previously worked only on Linux hosts that could open a TUN device, create a netns, and program nftables. Everywhere else `host_datapath()` returned a datapath whose entire behaviour was a refusal.

This adds a datapath that terminates guest TCP/UDP in userspace and re-originates each admitted flow on an ordinary host socket — no privilege required. macOS becomes a first-class host; Linux gets a fallback when `CAP_NET_ADMIN` is absent. `MacosUserspaceGateway`, the placeholder that only refused, is deleted rather than left as a shim.

**No workload gains a NIC.** Egress still leaves the guest only over vsock, and the host still originates every outbound connection — which is what keeps default-deny, secret substitution, and the audit chain mechanically enforceable. `check-vsock-only-egress` and `check-uniform-vsock-egress` stay green.

## The properties that make it trustworthy

- **The guest's `connect()` never lies.** The SYN is held until the host connect resolves, so the guest sees ESTABLISHED only for a destination that actually accepted.
- **Destination integrity.** The host socket connects to the exact `IpAddr` from the admitted packet — never a hostname or anything that could re-enter DNS below the policy seam — and the connected peer is asserted equal to it.
- **No flow a guest can pin forever.** Every state carries a deadline the guest cannot refresh by acting, including the two where no host error can ever arrive. Without this a guest spent its socket budget on flows it never completed and took the datapath out of service for the machine's life.
- **Bounded memory, asserted not described.** 46,476,608 bytes, computed from the component constants by a test whose residual form makes a dropped term name itself. Thirteen mutations, thirteen red.
- **Ingress is declared, never inferred.** A UDP mapping binds exactly the address declared, and binding is not admitting — a synthesized packet still goes back through `admit_inbound`, so withdrawing a declaration stops delivery while the socket is still bound.
- **IPv6 without the embedded-v4 bypass.** One extraction collapses v4-mapped, v4-compatible, NAT64 and 6to4 before the v4 rules judge the result, so `169.254.169.254` stays refused however it is spelled. Stubbing that extraction reddens seven tests.

## Two decisions recorded rather than deferred

**No root component.** A privileged macOS helper would have bought ICMP and raw IP. It was designed, reviewed, and **rejected** (ADR-039): the question was never whether it could be built safely but whether a workload needs raw IP on macOS, and none does. Those capabilities stay refused at admission, honestly and permanently.

**Multi-queue and zero-copy declined on measurement.** Throughput rises 3.2× from 1→16 flows, so a serial pass is not the ceiling; per-byte capacity is ~26 Gb/s on one core, so copying is not the limit either. The benchmark instead found that every service pass made one zero-return `kevent` costing ~12 µs on macOS — fixed, worth 2.9× on guest→host. The numbers live in the plan so the decision is auditable.

## Verification

The full workspace suite passes: 10,289 tests, 0 failures. Workspace clippy `-D warnings`, Linux `zigbuild --all-targets`, nightly fmt, and the xtask gates (`check-vsock-only-egress`, `check-uniform-vsock-egress`, `check-claim-catalog`, `check-doc-claims`, `check-no-overclaim`, `check-no-spec-refs-in-comments`) all clean.

The datapath ingress is fuzzed — `fuzz_datapath_ingress`, 47 seeds, ~441k executions, no crash — and that landed **before** backend selection made the parser guest-reachable, not after. Claim 5's witness ledger records it.

## IPv6 landed end to end after this PR was opened

The first revision admitted and carried IPv6 on the host while the guest kernel could not originate it, and said so. That gap is now closed rather than deferred: `CONFIG_IPV6=y` in the workload kernel, the agent brings the v6 half of the link up beside the v4 one over rtnetlink, the host leases a unique-local `/126` per machine on request, and the privileged Linux packet backend carries the family too instead of refusing what the socket backend already accepted. IPv6 is opt-in per plan and off by default; a plan that does not ask for it is unchanged in every byte. ADR-038 records the design and the measurements.

The kernel cost is one built-in symbol. IPv6 arrives with the v6 IPsec families and the 6-in-4 tunnel driver, and both are dropped: the IPsec options select the XFRM transform framework that stays in the required-disable set, and SIT — which carries no XFRM and so would have slipped past that guard — brings five more built-ins for a tunnel a guest with one point-to-point link cannot use.

## Known limitations, stated in the docs rather than omitted

- **Declared TCP ingress binds nothing** on this backend. Recorded as the remaining over-claim; UDP ingress is honest.
- ICMP, raw IP, and arbitrary IPv4/IPv6 are refused at admission on the userspace socket backend, permanently. The privileged Linux packet backend carries all of them.

## Follow-ups

None outstanding. The epic (#2111) and every workstream issue it tracked are closed: #2116 / #2147 (IPv6) and #2120 (node-control) landed here, #2119 (node-to-node transport) was designed and declined in ADR-036, and #2121 (WSL2) is resolved.


auser:
**This PR is currently red — do not merge yet.** Two jobs fail; `Test` passes.

- `Invariant #1` — the datapath fuzz harness binds a `TcpListener` as its own sink, tripping the no-server-in-mvm grep.
- `Lint` → closure budget — the branch adds `smoltcp`/`mio`/`socket2` to mvmctl's default closure (279 vs a budget of 275).

Both fixes are written but **uncommitted in the worktree**, along with ~1,500 lines of WS8 (#2120) node-control work whose agent died on a spend limit mid-report.

Full state, the reasoning behind each fix, and the verification still owed: **#2153**. Read that before touching the branch.

auser:
**Green.** Invariant, Lint (33m24s), Test (20m25s) and the claim-bearing lanes all pass on `590900980`.

The two failures are fixed in `c47a4e66f`, committed separately from the feature work:
- The architecture gate now skips `**/fuzz/**`, joining tests/benches/examples for the same reason its `#[cfg(test)] mod` filter exists — a listener a harness binds as its own sink is not a server. Verified by before/after `rg` to remove exactly one file and nothing else.
- The closure budget is raised to what actually ships (279), since the datapath's `smoltcp`/`mio`/`socket2` cannot be gated off on macOS where it is the only backend.

`590900980` also lands the node-control API (#2120) with the ownership mutation proof: forcing `CallerIdentity::owns` to `true` reddens five tests, restored green.

Ready for review.

auser:
⚠️ **Do not merge — a live-KVM regression was found in this branch.** See #2161.

The privileged Linux lane (`MVM_L3_PRIVILEGED_TESTS=1`) on a real KVM host:

| Tree | Result |
|---|---|
| `main` (`6d8931ad6`) | **7 passed, 0 failed** |
| this branch (`0749f5364`) | **5 passed, 2 failed** |

Both failures are host TUN **teardown**: `close()` no longer removes the device synchronously. It is reclaimed at process exit, so nothing leaks permanently, but a workload's teardown depends on the synchronous removal these tests assert.

No CI gate catches it: the lane is `cfg(target_os = "linux")` *and* env-gated, so it runs nowhere by default and `zigbuild --all-targets` only compiles it. **This PR is green with the regression in it** — which is the more important finding than the regression itself.

Likely cause is descriptor ownership: the branch added `Watched<S>` (socket + `Registration`, drop-order dependent) and `DatapathHandle::service`. If a TUN fd is now held past `close()`, the device outlives it. Unconfirmed.

auser:
**Correction: the merge block is lifted. The failures I reported were test-isolation, not a regression.**

Same branch, same KVM host, single-threaded:

```
MVM_L3_PRIVILEGED_TESTS=1 cargo test -q -p mvm-hostd --test l3_linux_privileged -- --test-threads=1
running 7 tests ....... test result: ok. 7 passed; 0 failed
```

Parallel 5/7, serial **7/7**. Host TUN teardown is correct on this branch.

The tests assert on globally-named host devices, so two running concurrently in one process contend for the same name. `main` passed in parallel by scheduling luck; this branch adds tests to that binary and changed the interleaving, exposing a latent flaw that was always there. Retitled and rescoped as #2161 — a pre-existing parallelism hazard, not blocking this PR.

**Live validation this produced, which is the useful outcome:** the privileged Linux lane now has a real run on real KVM against this branch — real host TUN, real nftables, real teardown, 7/7 — where previously it had none.

PR #2241, opened 2026-08-08
test: kernel budget control run (do not merge)

auser:
Control run. A single no-op comment appended to `nix/images/kernel/base.nix` on pristine main, to trigger the path-filtered kernel build gate.

Purpose: #2238 fails `Build kernels (x86_64)` at 917 =y symbols against a 916 budget, but its only kernel-file change is comment text, and `Build kernels (aarch64)` passes. A comment cannot move one arch and not the other. This run determines whether main is already at 917.

If this fails identically, the drift is pre-existing on main and unrelated to #2238 — latent because the kernel job only runs when a kernel path changes, and nothing has touched one recently.

Delete this branch once it reports.

### DIFF
--- a/nix/images/kernel/base.nix
+++ b/nix/images/kernel/base.nix
@@ -800,3 +800,5 @@ in
     mkKernel
     ;
 }
+
+# Control: no-op comment to trigger the path-filtered kernel build gate.


auser:
Control run served its purpose: pristine main + one no-op comment produces the identical failure — `workload kernel x86_64: 917 =y symbols (budget 916)`, same count, while aarch64 passes. #2238 is exonerated; the drift is main's.

Root cause: CI builds the kernel from `./nix/images/builder-vm#packages...`, but `kernel-build.yml`'s path filter watches `nix/images/builder-vm/flake.nix` — **not `flake.lock`**. A nixpkgs bump in that lock changes the toolchain that produces the kernel config (gcc/binutils cascading through `olddefconfig`) and never triggers the gate that would catch the symbol drift. Filed separately.

PR #2232, opened 2026-08-07
feat(secrets): mvm-side primitives for the tenant-secrets egress broker (RemoteResolver + pluggable backend)

auser:
## mvm-side primitives for the mvmd tenant-secrets vault (egress credential broker)

These are the shared, mvm-side building blocks the **mvmd** tenant-secrets vault (Phase 1, egress credential broker) builds on. mvmd resolves `${SECRET_NAME}` credentials **host-side** and injects them into outbound requests at the per-VM substitution endpoint — **no key material ever reaches the guest** (it holds only opaque `mvm-secret-…` placeholders). This PR is the mvm half of that seam; the mvmd half is a separate PR that depends on it via the `mvmctl = { path = "../mvm" }` facade.

Stacked on `feat/plan-202-shared-wire`; base advances with it.

### What's here (9 commits)

- **`RemoteResolver` over UDS** (`mvm-hostd::keyholder`) — a synchronous `SecretResolver` that fetches a secret's value from a fleet-secrets daemon over a `0600` Unix socket, mirroring `LocalResolver`'s fail-closed `allowed_hosts` check. Wire types (`ResolveWireRequest`/`ResolveWireResponse`, 4-byte-BE-length + JSON, 1 MiB cap, redacting `Debug`) live once in `mvm_core::substitution_wire` so client and server can't drift.
- **Pluggable resolver backend** — `EndpointConfig` gains a `ResolverBackend::{Local, Remote{uds_path, timeout_secs}}`; `Local` stays byte-identical to prior behavior (unit variant, single store-dir source).
- **Resolver-UDS confinement (M3)** — the substitution endpoint's Landlock profile grants the resolver UDS only when `Remote`; a missing socket surfaces as `PathNotFound` (fail-closed on boot) rather than a silently-empty grant.
- **Facade / spawn plumbing** — additive re-exports so fleet callers (mvmd) can reach what they need without a direct `mvm-sdk`/`mvm-hostd` dependency edge: `mvm_hostd as hostd` (`EndpointConfig`, `FileBindingStore`), `spawn_substitution_endpoint`/`reap_substitution_endpoint`/`SubstitutionSpawnParams`, and `SecretRef`/`SecretMount`. `spawn_substitution_endpoint` now threads `resolver_remote` + `binding_store_dir` through to the endpoint (existing backends pass `None`/`None`, byte-for-byte unchanged).

### Verification

[...]

PR #2245, opened 2026-08-08
fix(kernel): restore the x86_64 symbol budget and gate edits to it

auser:
`Build kernels (x86_64)` has been red on main since 2026-08-07, failing at 917 =y symbols against a 916 budget. **It is not drift — the kernel config did not change, the budget did.**

## What happened

#2219 (`ca68dd9c0`) lowered `BUDGET_X86_64` from 917 to 916 and deleted the paragraph explaining why it was 917: the workload kernel gained one built-in when the v6 datapath merged, so the guest can hold the v6 half of the pair its plan leases. The symbol is still in the kernel; only its allowance was removed.

That PR was a large squash whose branch carried a pre-IPv6 copy of the file, so this reads as a merge artifact rather than a decision — `BUDGET_AARCH64`, which rose in the same IPv6 change, was left at 944 and still passes. This restores 917 and the justification.

Confirmed empirically before changing anything: a control branch off pristine main with a single no-op comment on `base.nix` reproduced the failure exactly — same 917, same job, aarch64 green (closed as #2241).

## Why it sat on main for a day

`kernel-build.yml` is path-filtered, and **the budget constant was not one of the watched paths.** A PR editing only the ceiling never built the kernel that ceiling describes, so nothing checked the new number against a real config. The next PR to touch a kernel path then inherited a red build it did not cause — that was #2238, and #2232 hit it earlier.

Two paths added:

- `xtask/src/check_kernel_config_budget.rs` — changing the number now rebuilds the config that gives it meaning.
- `nix/images/builder-vm/flake.lock` — this job builds `./nix/images/builder-vm#...`, so a nixpkgs bump there moves the toolchain that runs `olddefconfig`. Not the cause here, but the same class of unwatched input.

## Diagnosability

Artifact upload becomes `if: always()`. The budget gate exits non-zero **before** the upload step, so a regression discarded the one artifact that explains it — the resolved `.config`. A failure that reports a count and destroys its own evidence costs an hour of CI to re-derive; that is how I ended up bisecting this from commit history instead of reading a diff.

Note this is a different failure mode from the one `check-workflow-paths` guards. That gate proves the paths a workflow names still *exist*; nothing checks the set is *complete*.

## Verification

- `check-workflow-paths` clean across 20 workflow files
- 437 xtask tests pass
- workflow YAML parses with the five expected path entries
- workspace clippy clean (pre-commit)

Unblocks #2238 and any other PR touching a kernel path.


auser:

PR #2248, opened 2026-08-08
feat(backend): uniform capability negotiation that names an alternative

auser:
Using `mvm` as a library means holding a backend you did not choose and asking whether it can serve what you need. `VmCapabilities::shortfall` already answered *which* capabilities are missing — but a bare `["vcpu_state_snapshot"]` tells a caller a request was refused without telling them what to do instead. Fine for host code that knows the backend matrix by heart; useless as a library surface.

`VmBackend::negotiate` answers both at once, **identically for every backend** — Firecracker, libkrun, HVF, QEMU, Mock and Wasm:

```rust
match backend.negotiate(&required) {
    Ok(()) => { /* served outright */ }
    Err(gaps) => for gap in gaps {
        println!("{}: {}", gap.capability, gap.alternative.describe());
    }
}
```

It is a non-overridable default method composing `capabilities()` and `kind()`, because a backend that could answer here differently from its own capability matrix would be exactly the dishonest tier ADR-024 rules out.

## Alternatives are per (capability, backend)

Because the honest answer differs. A tier with no vsock device reaches the per-VM substitution endpoint **over a Unix socket**; the wasm tier reaches the **same endpoint** through its `mvm:egress` host import. Both land on the same governance seam — same policy, same chain-signed audit — which is what makes them real answers rather than consolation prizes. A missing snapshot tier resolves to replaying the signed plan from cold, or to a standby pool on backends that can hold one.

## `CapabilityAlternative::None` is the load-bearing variant

Some capabilities are **security boundaries, not features**. A backend that cannot keep a routable NIC away from the workload has no substitute to offer: the host would stop being the originator of every outbound connection, which is what makes egress policy and the audit chain mean anything at all.

Modelling that as a distinct variant is what stops a caller reading "no alternative" as "alternative not written yet" — and it is what keeps this from becoming the *silently degrade* option ADR-024 considered and explicitly rejected. Every refusal here is typed, and it either names a substitute or names the reason there isn't one.

This adds **no numbered security claim**, per ADR-024 constraint 3.

## Tests — 12

Ten cover the resolution table directly. One of them asks for **all ten capabilities at once** and asserts none falls through to the unrecognised-capability arm, so adding a capability without an alternative fails the suite rather than shipping a hole. Two more drive the real `WasmBackend` through the trait — what a library consumer actually calls, rather than a hand-built capability matrix.

## On BDD

I did not add a conformance scenario, deliberately. `features/suites/` is a **claim-conformance harness gated in both directions**: every scenario must name a registered `MVM-SEC-` id, and `meta.rs` fails both on a registered claim with no scenario and on a scenario naming an unregistered id. Putting this there would mean minting a security claim for a tier ADR-024 says carries zero. The behavioural coverage stays in the tests above.

## Verification

- 3128 tests across `mvm-contract` / `mvm-core` / `mvm-runtime`
- 546 tests under `wasm32-wasip1`; `wasm32-unknown-unknown` builds — the negotiation layer is `no_std` and reaches the browser tier
- `cargo check --workspace` clean; workspace clippy `-D warnings` clean (pre-commit)
- `check-no-string-backend-dispatch` and `check-no-spec-refs-in-comments` clean


auser:

PR #2249, opened 2026-08-08
fix(audit): store the bytes a chain line's signature covers

auser:
Plan 302 WS5. Independent of #2242 — different files, either can land first.

## The defect

`verify_audit_chain` re-ran `serde_json::to_vec(entry)` on the parsed entry and checked the signature against the result. That makes a signature a claim about the **current struct definition**, not about the past.

Add one always-serialized field to `AuditEntry` and every historical line re-serializes differently and stops verifying — the log accuses itself of tampering because a struct grew. It also conscripts every future reader, in any language, into reproducing this crate's serializer exactly and forever.

## The fix

`SignedEnvelope` gains `canonical`: the exact bytes the signature covers.

```json
{"entry":{...},"canonical":"<base64 of the signed bytes>","prev_hash":"...","signature":"..."}
```

Both verifiers read those bytes and re-derive nothing. The readable `entry` stays beside them — an audit log nobody can `grep` is an audit log nobody reads — and both verifiers assert the two agree, so the readable copy cannot drift from the attested one.

The signed bytes themselves are unchanged: the first line of the frozen corpus keeps its original signature. Only the line gained a field, which is why downstream `prev_hash` values differ.

**No cutover.** Lines without `canonical` verify by the original rule indefinitely. Evidence somebody already holds must not stop verifying because the writer improved.

## Why not JCS

I originally scoped this as "converge on JCS" and that was wrong twice. The plan records it, because the wrong version is the tempting one:

1. **There was no second implementation to converge with.** `web/audit-verify` is a `wasm_bindgen` shim over `mvm-contract` — the same Rust code, not an independent verifier. The cross-language divergence this was meant to fix didn't exist.
2. **Our JCS isn't JCS.** `serde_jcs` 0.1.0 orders keys by UTF-8 byte value; RFC 8785 mandates UTF-16 code units. They disagree on astral-plane keys — `mvm-core`'s `canonicalizer_equivalence` documents it and `audit_chain_spine.rs` exercises a `U+1F600` key. Adopting it would have swapped a known quirk for one that merely *looks* standard.

Storing the bytes removes the whole class: nothing re-sorts anything at verify time, so no key-charset restriction is needed either.

## Tests

The decisive one is `a_line_whose_signed_bytes_differ_from_our_serializer_still_verifies`: a genuine line whose signed bytes differ from what this crate would emit. It verifies now; the same line with its stored bytes removed is rejected — which is exactly what a schema change would have done to an entire historical log.

Also: `a_chain_written_without_stored_bytes_still_verifies`, `the_pre_stored_bytes_corpus_still_verifies` (the v1 fixture, now asserted by *verifying* rather than byte-comparison, and marked never-regenerate), `a_readable_entry_that_disagrees_with_the_signed_bytes_is_refused`, `tampering_with_either_copy_of_the_entry_is_refused`, `a_consistently_rewritten_line_still_fails_the_signature`.

Three existing tamper tests now assert `EntryCanonicalMismatch` instead of `SignatureInvalid` — an in-place edit reaches the readable entry but not the base64 beside it, so the refusal names the tamper more precisely. The signature remains the security boundary; the agreement check is a sharper diagnostic, and the consistent-forgery test pins that.

Full workspace suite green (10593/10593), clippy clean, nightly fmt clean, doctests and policy gates pass.

**One thing I could not verify locally:** `cargo check -p mvm-contract --target wasm32-unknown-unknown` fails in my environment with `can't find crate for core`, but it fails identically on unmodified `main`, so it is a pre-existing local toolchain issue rather than something this change introduces. `web/audit-verify` is excluded from the workspace, so CI does not gate it either — worth a second pair of eyes if anyone has a working wasm setup.


auser:
Dequeued from the merge queue for now: it was sitting `UNMERGEABLE` at position 3, which is dead weight in a serial queue.

Cause is an overlap with #2251 — both PRs edit `crates/mvm-client/src/audit/normalize.rs`. The merge queue tests entries batched together, so this was being merged on top of #2251 and conflicted with the *batch*, not with `main`. Rebasing onto main alone doesn't resolve that.

Rebased onto current main regardless (full suite green, 10592/10592). Holding it out of the queue until #2251 lands, then I'll rebase on top of it and resolve the `normalize.rs` overlap properly — re-queuing before then would just go `UNMERGEABLE` again and block the entries behind it.

PR #2250, opened 2026-08-08
refactor(backends): complete the Firecracker extraction into mvm-backends

auser:
Follow-up to #2234. Moves the backend-agnostic snapshot seam down into `mvm-vmm` and starts relocating the Firecracker mechanics into `mvm-backends::fc`, so `FcDriver` can eventually join the other four drivers without a back-edge into `mvm-runtime`.

Plan: `specs/plans/298-extract-firecracker-driver.md`.

## Fewer traits and types, not more

A move-only refactor that *adds* abstraction has failed, so this one is held to zero new traits and zero new structs — and comes out net negative:

- **`ForkVmFullRestorer` deleted.** One method, one production impl, one test double, one call site, existing purely for test injection — while `VmmDriver::fork_standby_child` was already the real seam. `fork_vm_full_fc` now takes a callback; `MockRestorer` became a closure. A trait, a struct and two impl blocks gone.
- **Two `SnapshotIO` doubles became one.** `SpyIO` was a strict superset of `CannedIO` (call log + forced NIC); one configurable double covers both, and its constructor is shorter at every call site than the struct literal it replaces.
- **No device-model view trait.** The guard's entire body is an is-it-empty check, so `SnapshotIO` reports a `usize` and the guard takes a `usize`. A trait there would be a type-shaped way of saying "a number". Making it a `SnapshotIO` method was rejected too — that hands every backend its own copy of a claim-bearing refusal.

## The device-model guard

`assert_vsock_only_device_model` is a claim-10 witness, so it gets stated explicitly: it moved to `mvm-vmm` and remains **exactly one function with one implementation**, still sequenced between snapshot load and resume by the generic helper rather than reimplemented per backend. It briefly existed twice during the move — caught by the mutation-surface gate flagging new surface — and that duplicate is gone. All the load → guard → resume ordering witnesses, including the refusal paths that tear down without resuming, are unchanged and green.

`RestoredDeviceModel` deliberately did **not** move: its own doc calls it "a slice of Firecracker's own config schema", so lifting it would put an FC wire format in the neutral crate. The parsing stays with the backend that owns the format.

## Scope of the move

Into `mvm-vmm`: `SnapshotIO`, the guarded load paths, `CannedIO`, the guard.

Into `mvm-backends::fc`: the FC API client, VMM process lifecycle, fork mount namespace, and the per-VM path layout they read.

Deliberately left in `mvm-runtime`: `pause_and_seal` / `verify_and_resume` stay with the store they depend on (on-disk layout, HMAC envelope, tenant-key encryption). Only the `guarded_*` paths are store-free, and they are all a backend needs — moving the sealing helpers would have dragged the store down for no caller.

## One deletion worth calling out

`run_from_build` and three siblings had zero non-comment callers anywhere in the repo, and `run_from_build` was not a launcher: it validated and then bailed with "raw Firecracker flake launch is disabled". An unreachable refusal protects nothing, it is not a registered control (`check-dormant-controls` knows 4, none of them this), and the invariant it gestured at is enforced by `check-vsock-only-egress` and `check-uniform-vsock-egress` — both green here. `flake_run.rs` goes 383 → 207 lines and becomes the descriptor its name implies.

That deletion is what unblocked the rest: it revealed that `base/config.rs` was a 297-line dependency-free leaf, and moving it to `mvm-vmm::host` freed the remaining Firecracker modules without relocating `RuntimeVolume` or `VmSlot`.

## Still to come

`control` / `observe` / `guards` / `firecracker.rs`, the `snapshot.rs` split, and `driver/fc.rs` itself. The plan records the settled boundary.

## Verification

- `cargo nextest run --workspace` — 10,582 passed, 24 skipped
- `check-claim-catalog`, `check-vsock-only-egress`, `check-uniform-vsock-egress`, `check-mutation-witnesses`, `check-dormant-controls`, `check-no-vz`, `check-no-spec-refs-in-comments`, `check-workflow-paths` — all clean

No workload-visible behavior change and no security-claim change.


auser:
**Scope grew since this was opened — it now completes the extraction rather than starting it.** Retitled accordingly.

All five drivers live in `mvm-backends`. `mvm-runtime::driver` is re-exports only, and `mvm-backends` depends on neither `mvm-runtime` nor `mvm-build`.

**Trait/type budget held: zero new traits, zero new structs, net negative.**
- `ForkVmFullRestorer` deleted — one method, one impl, one call site, existing only for test injection while `VmmDriver::fork_standby_child` was already the real seam. Now a callback.
- Two `SnapshotIO` doubles collapsed into one; `SpyIO` was a strict superset of `CannedIO`.
- The device-model guard takes a NIC count rather than gaining a view trait — its whole body is an is-empty check, so a trait there would have been a type-shaped way of saying "a number". Making it a `SnapshotIO` method was also rejected: that hands every backend its own copy of a claim-bearing refusal.

**Dead code removed rather than relocated:**
- the retired raw flake launcher — four public functions, zero callers, and `run_from_build` was a refusal stub rather than a launcher
- `require_linux_env` — `fn(..) -> Result<()> { Ok(()) }`, called at seven sites, asserting nothing
- a duplicate `bind_unix_listener`; `mvm-vmm`'s is now the only copy

**On `FirecrackerBackend`** — worth a reviewer's attention because it changed twice. An intermediate commit moved the shell into `mvm-backends`, which cost it its `start` path (`fc_runner()` lives in `mvm-runtime`) and left it refusing. That was wrong: a `VmBackend` that cannot start a VM is not a backend, and inheriting the trait's default refusal would only have made the same lie tidier. The shell now sits back in `mvm-runtime` beside the runner, with `fc_runner().start(config)` intact and every trait method working. `FcDriver` stopped wrapping it and reports its own identity, capabilities and Tier 1 profile — removing a driver-to-legacy-shell dependency rather than relocating one.

**Verification:** 10,582 workspace tests, doctests, `clippy --workspace --all-targets -- -D warnings`, **146/146 BDD scenarios (630/630 steps)**, and the security gates — `check-claim-catalog`, `check-vsock-only-egress`, `check-uniform-vsock-egress`, `check-mutation-witnesses`, `check-dormant-controls`, `check-no-vz`, `check-no-string-backend-dispatch`, `check-no-spec-refs-in-comments`.

No workload-visible behavior change and no security-claim change. The device-model guard still runs between snapshot load and resume, sequenced by the generic helper in `mvm-vmm`, as exactly one function with one implementation.

PR #2257, opened 2026-08-08
feat(hvf): complete save/restore so checkpoint and fork work on macOS

auser:
## What

Makes `mvmctl machine checkpoint create --class vm-full`, `checkpoint restore`,
and `checkpoint fork` work on the HVF backend, and flips
`snapshot_capability` from `Unsupported` to `SaveRestore`.

## Why the flag was the smallest part

The VMM primitives were already built and are exercised today by the warm-pool
capture path: the frame codec (`backends/hvf/snapshot.rs`), the paused capture
in the run loop, the `restore_ram`/`restore_frame` private-mapping restore, and
an `HvfVmFullControl` covering the whole `VmFullControl` surface. What was
missing was the plumbing around them:

- Nothing anywhere set `restore_ram`/`restore_frame` — a dead config path.
- The CLI hardcoded `FcVmFullControl`, so a capture on HVF would have paused
  the wrong thing.
- The liveness probe knew `fc.pid` and `vz.pid`; a live HVF VM read as stopped.
- `vm_is_quiesced` knew only Firecracker's pause marker.
- `checkpoint restore` was a `bail!`.

## Changes

**Restore entry point.** `mvm-backends/src/driver/hvf_restore.rs` rewrites a
captured parent's launch config into one that boots a child from its own
materialized copies. Two invariants carry the security weight:

1. Every per-VM host path is re-derived from the child's state dir — pid file,
   console log, workload exit, pause marker, snapshot rendezvous, agent socket.
   Keeping any of the parent's would make two live VMMs fight over one identity.
2. No authority-bearing relay survives: substitution endpoint, broker, dev
   console sockets, live-handoff control are all dropped. A restored guest has
   no path off the box until a claim binds one — and because the drop happens at
   config time, that holds *before* its vCPU runs.

`mvm-runtime/src/hvf_restore.rs` holds the two thin adapters
(`ForkVmFullRestorer`, `VmFullRestore`). The same-identity resume clones the
sealed content into the target's state dir first: a resumed guest writes through
to its rootfs, and those bytes are what every later fork verifies against.

**Backend-neutral dispatch.** `AnyBackend::vm_full_control` (exhaustive match),
a catalog-driven `vm_is_running`, and `mvm_core::checkpoint::vm_full_origin`
replacing `checkpoint_is_vz`. That last one mattered: the old heuristic keyed on
the `supervisor-config.json` blob, which every HVF checkpoint also carries, so
it would have classified all of them as the removed backend's. Dispatch is now
on the machine-state blob — the bytes a restore would actually load.
`fork_vm_full_fc` → `fork_vm_full`; only the restorer was ever VMM-specific.

## Security

- **Restore is now chain-anchored.** It previously verified content hashes
  only, so a record edited after it was audited was restorable even though it
  was not forkable. `restore_checkpoint` now takes a `CheckpointChainAnchor`
  and runs the same fail-closed gate the fork paths do.
- **Fresh child identity.** An HVF fork admits its own claim-8 plan (own nonce,
  own name, `deny_all`), delivers a fresh generation token, and refuses anything
  short of acknowledged + reseeded + clock-resynced — stopping the child if it
  cannot prove it rotated.
- **No experimental opt-in.** Firecracker's guard exists because a restored FC
  child inherits the parent's guest IP/MAC out of saved memory and collides on
  the shared bridge. An HVF guest has no NIC to inherit an address on, so the
  guard has nothing to protect against and is not carried over.
- `capture_fs_quick` and the dm-verity binding check are untouched.

## Two refusals worth calling out

**A capture is refused when the VM has a writable disk.** A snapshot carries the
devices' control plane, never their backing bytes. That is sound for a
read-only file-served disk and unsound for a RAM-backed ephemeral rootfs, whose
writes vanish with the VMM — restoring against one resumes a guest whose page
cache disagrees with its disk. The two workload tiers HVF actually runs are
unaffected (sealed/verity attaches only read-only disks; virtiofs-root attaches
none). Lifting this needs a new snapshot section for block-image contents; it is
recorded as a known limit in the plan, not papered over.

**Pause retention is now the caller's choice**, ANDed with the backend's
capability. Only the warm pool wants its captured parent held paused — for it
the paused parent is the claim resource. Routing the user-facing verb through
the HVF control was otherwise leaving a checkpointed VM stopped; caught by
running it, and now covered by a test.

## Benchmarks

Live on macOS 26.5.2 / Apple silicon, 512 MiB sealed workload, full release
binary set:

| operation | p50 | range |
|---|---|---|
| `checkpoint create --class vm-full` | 1635 ms | 1527–1849 |
| `checkpoint restore` | 690 ms | 684–692 |

**This is not a warm-launch change.** HVF's fast launch is resident handoff at
18.9 ms p50 dispatch, and a restore is ~36× that by construction — it starts a
new VMM against bytes on disk where a handoff transfers a running process. The
two answer different questions; the fork lineage needs the second. Full detail
and the reproduction recipe in `specs/plans/304-hvf-save-restore-bench.md`.

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo nextest run --workspace` — 10601 passed, 24 skipped
- `cargo test --workspace --doc`
- `cargo test -p mvm-conformance --test conformance --features bdd` — 152/152
  scenarios, 649/649 steps, including 6 new ones under
  `features/suites/s11_snapshot/hvf_save_restore.feature`
- xtask: `check-file-size`, `check-honesty`, `check-no-overclaim`,
  `check-deferrals`, `check-single-home`, `check-conformance`,
  `check-abi-layout`, `check-no-spec-refs-in-comments`, `check-claim-catalog`,
  `check-uniform-vsock-egress`, `check-vsock-only-egress` — all clean
- `cargo zigbuild --target x86_64-unknown-linux-gnu` clean for every crate this
  touches (`mvm-core`, `mvm-runtime`, `mvm-backends`, `mvm-vmm`, `mvm-hostd`).
  The full-workspace form needs the pinned zig for `mvm-cli`'s embedded-binary
  build script (`just toolchain-embed`), which is not provisioned on this host.
- Live end-to-end: capture and restore driven against a real HVF VM, not just
  unit-level seams.

One environmental note, in the bench doc: on a developer `~/.mvm` whose
`local.jsonl` audit chain is already broken (`mvmctl trust audit verify` fails
on `main` there too), every checkpoint reads as un-audited and both
`checkpoint verify` and `checkpoint restore` fail closed. That is the gate
working. The restore timings were taken after moving the broken pre-existing
chain aside.

`checkpoint.rs` crossed the 1500-line production cap, so the vm_full fork arms
moved to `commands/vm/checkpoint/fork_vm_full.rs` with their tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3112,6 +3112,7 @@ dependencies = [
  "mvm-hostd",
  "mvm-net",
  "mvm-runtime",
+ "mvm-vmm",
  "serde",
  "serde_json",
  "sha2 0.11.0",
--- a/crates/mvm-backends/src/driver/hvf.rs
+++ b/crates/mvm-backends/src/driver/hvf.rs
@@ -29,12 +29,16 @@ use mvm_vmm::hvf_handoff::HvfHandoffRequest;
 use crate::legacy::hvf::{
     self as hvf_backend, HvfBackend, PID_FILE_NAME, PID_FILE_TIMEOUT, resolve_supervisor_path,
 };
-use mvm_core::checkpoint::DeviceAnchors;
+use mvm_core::checkpoint::{DeviceAnchors, HVF_FRAME_BLOB};
 use mvm_vmm::checkpoint::VmFullControl;
 use mvm_vmm::driver::spec::{BlockDev, KernelImage, VmmSpec};
 use mvm_vmm::driver::traits::{
     ChildForkRequest, DuplexStream, RunningVm, StandbyParentSpawn, VmmDriver,
 };
+
+/// How long a capture waits for the paused supervisor to publish both halves of
+/// its snapshot. Generous: the write is proportional to guest RAM.
+const SNAPSHOT_PUBLISH_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 
 /// The first-party VMM driver: pure VMM mechanics, no policy and no admission.
 /// It boots what a `VmmSpec` describes and relays the guest's egress port to the
@@ -654,6 +658,37 @@ impl HvfVmFullControl {
             self.state_dir.join("snapshot.frame"),
         )
     }
+
+    /// Refuse to capture a VM whose guest can write to a block device.
+    ///
+    /// A snapshot carries guest RAM plus the devices' control plane, never a
+    /// device's backing bytes. That is sound for a read-only, file-served disk:
+    /// the restore rebinds the same unchanged image. It is not sound for a
+    /// writable one — a RAM-backed ephemeral rootfs loses every guest write
+    /// when the VMM exits, and a write-through image keeps changing after the
+    /// capture — so a restore would resume a guest whose page cache disagrees
+    /// with its disk. Fail closed at capture rather than mint a checkpoint that
+    /// silently reverts data.
+    fn ensure_every_disk_is_restorable(&self) -> Result<()> {
+        let Some(path) = self.supervisor_config_path()? else {
+            return Ok(());
+        };
+        let cfg: HvfSupervisorConfig = serde_json::from_slice(
+            &std::fs::read(&path)
+                .with_context(|| format!("reading HVF launch config {}", path.display()))?,
+        )
+        .with_context(|| format!("parsing HVF launch config {}", path.display()))?;
+        if let Some(writable) = cfg.disks.iter().find(|disk| !disk.read_only) {
+            bail!(
+                "cannot capture a full-VM checkpoint of HVF VM '{}': its disk {} is \
+                 writable, and a snapshot does not carry device backing bytes. \
+                 Capture an fs_quick checkpoint of the paused VM instead.",
+                self.vm_name,
+                writable.path.display()
+            );
+        }
+        Ok(())
+    }
 }
 
 impl VmFullControl for HvfVmFullControl {
@@ -672,6 +707,7 @@ impl VmFullControl for HvfVmFullControl {
             "save_memory requires an absolute path, got {}",
             memory_path.display()
         );
+        self.ensure_every_disk_is_restorable()?;
         let parent = memory_path
             .parent()
             .ok_or_else(|| anyhow!("memory path has no parent directory"))?;
@@ -682,13 +718,15 @@ impl VmFullControl for HvfVmFullControl {
         let _ = std::fs::remove_file(&frame);
         std::fs::write(&request, b"capture\n")
             .with_context(|| format!("request HVF snapshot at {}", request.display()))?;
-        let deadline = Instant::now() + std::time::Duration::from_secs(30);
+        let deadline = Instant::now() + SNAPSHOT_PUBLISH_TIMEOUT;
+        let mut wait = std::time::Duration::from_micros(200);
         while (!ram.exists() || !frame.exists()) && Instant::now() < deadline {
-            std::thread::sleep(std::time::Duration::from_millis(10));
+            std::thread::sleep(wait);
+            wait = (wait * 2).min(std::time::Duration::from_millis(5));
         }
         anyhow::ensure!(
             ram.exists() && frame.exists(),
-            "HVF supervisor did not publish its paused snapshot within 30s"
+            "HVF supervisor did not publish its paused snapshot within {SNAPSHOT_PUBLISH_TIMEOUT:?}"
         );
         std::fs::copy(&ram, memory_path).with_context(|| {
             format!(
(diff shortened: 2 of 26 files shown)


auser:
Rebased onto `origin/main` (2e5bfb5), which brought #2235 and #2238 into the same blast radius. Resolutions, since two of them differ from what this branch originally did:

**Liveness probe.** The branch made `vm_is_running` catalog-driven; #2235 had independently consolidated both copies onto the shared `mvm_vmm::host::process_liveness` list and added an `EPERM`-means-alive rule that a bare `kill(pid, 0) == 0` misses on a root-owned Firecracker. Kept #2235's implementation — one shared list, not a second mechanism — and kept this branch's catalog-derived *test*, renamed `vm_is_running_covers_every_catalog_pid_marker`. It drives the probe from `started_vm_probe_descriptors()`, so a backend that registers a marker the shared list does not carry goes red rather than silently reading as stopped. Both intents hold; nothing is duplicated. The CLI keeps delegating (to `mvm_runtime::checkpoint::vm_is_running`) so it cannot regrow a private list.

**Checkpoint classification.** #2235 renamed `checkpoint_is_vz` to `checkpoint_carries_supervisor_config`. This branch deletes the predicate outright for `mvm_core::checkpoint::vm_full_origin`, which classifies from the machine-state blob (`memory.bin.hvf-frame` / `vmstate.bin`) instead of the supervisor-config blob. Took this branch's version: the supervisor-config blob is carried by HVF captures too, so the renamed predicate is still true-by-accident rather than true-by-construction. `Retired` and its refusal message stay oblique, and `check-no-vz` (including the naming rule #2235 added) is green.

**Negative control.** The liveness test's "a marker no registered backend owns is not liveness" control wrote `vz.pid`, which #2235's new naming rule flags. Renamed to `retired.pid` — same meaning, no literal.

`retain_paused: true` remains warm-pool-only (`workload_runner/runner.rs`); the CLI capture path stays `false`. The vm_full fork arms stay in `commands/vm/checkpoint/fork_vm_full.rs` and `check-file-size` passes.

Validation on the rebased tree: nightly `fmt --all --check` clean; `clippy --workspace --all-targets -D warnings` clean; `nextest --workspace` 10600 passed / 24 skipped; `--workspace --doc` clean; BDD 152/152; xtask `check-file-size`, `check-honesty`, `check-no-overclaim`, `check-deferrals`, `check-single-home`, `check-conformance`, `check-abi-layout`, `check-no-spec-refs-in-comments`, `check-claim-catalog`, `check-uniform-vsock-egress`, `check-vsock-only-egress`, `check-no-string-backend-dispatch`, `check-test-home-isolation`, and #2235's `check-no-vz` all pass.

`specs/plans/304-hvf-save-restore.md`, `specs/SPRINT.md` and `specs/REFACTOR-STATUS.md` were corrected in the same commit to describe the resolution above rather than the pre-rebase design.

auser:

auser:
Rebased a third time, now onto `ef21bd611` (post-#2266). Two things worth recording, because neither was mechanical.

**#2250 replaced the seam this builds on.** The Firecracker extraction deleted `ForkVmFullRestorer` in favour of a `ForkRestore` callback once `FcForkRestorer` moved into `mvm-backends` — its own `REFACTOR-STATUS` line reads *"`ForkVmFullRestorer` deleted — one method, one impl, one call site; now a callback."* I adopted that shape rather than reinstating the trait: the trait only ever existed to hold two impls, and a closure does that without `mvm-runtime` naming either VMM. `HvfForkRestorer` is now an inherent method mirroring `FcForkRestorer`, and both fork arms pass a closure.

**`check-file-size` went red, and it was this pair's doing.** `workload_runner/runner.rs` hit 1503 production lines against a 1500 cap. Neither change causes it alone — #2266 grew the refusal classifier, this one adds `vm_full_control` and `retain_paused`. Extracted `runner/refusal.rs` rather than shaving lines to squeak under: `require_fresh_child_identity`, `refuse` and `map_lineage_refusal` are one unit — each turns evidence into a refusal *reason*, and the reason is the whole product, since a claim that refuses for the wrong stated reason sends an operator to fix something that is not broken. Both tests moved with the code, per the gate's own guidance. The `runner/` submodule tree already existed, so this follows the established shape.

Also kept both SPRINT.md entries where #2266 and this PR collided at the top of the file.

**On the two test failures in my full run** — both were environmental and I chased them rather than assuming. `evicting_one_event_costs_the_same…` passed isolated. `stepping_across_the_ttl_keeps_the_lease_from_a_competitor` failed *twice*, which is why I kept going: it is a 60 ms lease TTL stepped with 20 ms sleeps, and this box was at load average 293 with 246 cargo/rustc processes from other worktrees. It passes on a quiet machine, this branch does not touch `mvm-hostd` at all (`git diff --stat origin/main -- crates/mvm-hostd/` is empty), and the same commit passed that test earlier today. #2270 has since landed doing exactly the right fix — sizing those TTLs from a measured stall instead of a 20 ms guess — which is the same diagnosis.

Re-validated on the rebased tree with the machine quiet: nightly fmt clean, clippy `--all-targets -D warnings` clean, `nextest --workspace` 10675 tests with only those two load flakes (both since confirmed passing), doctests clean, BDD **152/152**, and every `ci.yml` xtask gate green including `check-file-size`.

Verified clean-merging against current `main` (`8b5d77d90`) after the push, so no fourth rebase is pending.

PR #2266, opened 2026-08-09
fix(audit): report an unverifiable chain as unverifiable, not as a missing entry

### LINKED ISSUE #2258: Audit-chain verification failure is reported as "this checkpoint was never audited"
## Summary

When a host-lifecycle audit chain fails verification, `SignedChainAnchor::load`
logs a `tracing::warn!`, skips the chain, and continues with an empty index. Every
lineage lookup then returns `None`, and the operator sees:

> checkpoint 'X' has no signed audit entry to anchor its content-address;
> refusing to treat an un-audited record as verified

That claim is **false and misleading**. The record may be perfectly well audited —
the *ledger* is unreadable. The two situations call for opposite responses:

- "never audited" → you skipped a step; re-capture with a signer present.
- "ledger unverifiable" → your integrity chain is damaged; **investigate**.

Reporting the second as the first hides a tamper/corruption signal behind a
routine-looking message. The only command that tells the truth today is
`mvmctl trust audit verify`, and nothing points you at it.

## How it was found

A `machine checkpoint restore` on a developer host failed with the message above.
The checkpoint was fine. `~/.mvm/audit/local.jsonl` was not:

- 40,458 envelopes, 2026-05-12 → 2026-08-08. All parse; none truncated.
- 113 lines carry two concatenated envelopes with no separating newline.
- After re-framing those lines the verifier gets further and fails honestly:
  `prev_hash mismatch at line 3`.
- **4 distinct `prev_hash` values are each claimed by 3–4 entries.**

Multiple writers read the same chain head and each appended as its successor —
the concurrent-writer race fixed by #2239 (686f5a327). The missing newlines and
the forked links are the same race at two layers.

The break is at line 3 of 40,458, in the file's first minute. Only two entries
have ever been anchorable; the chain has been inert for three months while
everything kept writing into a ledger that proved nothing. Nobody noticed,
because nothing says so.

## Proposed fix

1. `SignedChainAnchor` records the chains that failed verification. When a lookup
   finds no entry **and** at least one chain was unverifiable, return `Err`
   naming the file(s) — "cannot determine whether this was audited: N audit
   chain(s) are unverifiable" — instead of `Ok(None)`. Same fail-closed
   behaviour, honest reason. A `None` with all chains clean keeps today's
   "never audited" message, which is then true.
2. `mvmctl doctor` reports audit-chain verification status, so a damaged chain is
   visible as a posture failure rather than surfacing later through an unrelated
   verb.

## Explicitly not proposed

A `trust audit repair` / auto-reset verb. Someone able to damage one line could
then force a chain reset — converting tamper-*detection* into tamper-*erasure*.
Recovery should stay a deliberate human act that leaves a visible artifact
(quarantine the file under a new name; never delete, never re-sign).

## Context

Surfaced while landing #2257 (HVF save/restore). That PR widens the reach of this
message — `checkpoint restore` is now chain-anchored, so it hits this gate too —
but the behaviour predates it and is independent of it.

auser:
Closes #2258.

## The bug

When a host-lifecycle audit chain fails verification, `SignedChainAnchor::load` warns, skips it, and continues with an empty index. Every lookup then returns `None`, and the operator is told:

> checkpoint 'X' has no signed audit entry to anchor its content-address; refusing to treat an un-audited record as verified

That is false. The record may be perfectly well audited — the *ledger* is unreadable. The two situations call for opposite responses:

- never audited → you skipped a step; re-capture with a signer present.
- ledger unverifiable → your integrity chain is damaged; **investigate**.

Reporting the second as the first hides a tamper/corruption signal behind a routine-looking message. On the host where this surfaced, the chain had been inert for three months and nothing said so.

## The fix

The anchor records which chains failed verification, and `recorded_creation_digest` returns `Err` naming them when a lookup misses **and** at least one chain was unverifiable. A miss with every chain clean keeps today's message, which is then true. **Same fail-closed verdict either way** — no admission decision changes, only the reason.

A hit still resolves normally even with a damaged sibling chain: that entry came from a chain that verified, so it is still trustworthy. Only the *unanswerable* case escalates.

### Kept apart downstream, not just in the message

- `mvm_runtime::lineage` exports `NO_SIGNED_ENTRY` and `LEDGER_UNVERIFIABLE` as shared sentinels. The first was a string literal retyped at ten call sites — including a classifier that switches behaviour on it — so the two messages and their matchers can no longer drift.
- `ClaimRefusal::LedgerUnverifiable` is new. Without it an unreadable ledger fell through the warm-pool classifier to `ParentTampered`: the opposite error, blaming a parent that may be perfectly sound.

### doctor

An `audit chain` line that **fails** the check when a chain does not verify, so a damaged ledger is a posture finding in its own right rather than something an unrelated verb trips over later. Verdict mapping is a pure function over a scan struct, tested with fixtures.

The sweep loads the host signer only when its secret half already exists — a diagnostic must not mint a signing key as a side effect (`scanning_a_fresh_home_creates_no_host_signer_key`).

`is_host_lifecycle_chain` moved to `mvm-core::config`, beside the `WORKLOAD_AUDIT_SUFFIX` it keys on, so the anchor's sweep and doctor's cannot disagree about which files are in scope — a doctor that verified a different set than the anchor would report a posture the anchor does not act on. A quarantined chain (`local.jsonl.forked-2026-05-12`) correctly drops out of scope, which is what makes the documented recovery step clear the finding.

## Explicitly not built

A `trust audit repair` / chain-reset verb. Someone able to damage one line could then force a reset, converting tamper-**detection** into tamper-**erasure**. Recovery stays a deliberate human act that leaves a visible artifact: quarantine the file under a new name, never delete it, never re-sign it. Both the refusal and the doctor line say exactly that, and `the_broken_chain_guidance_says_quarantine_and_never_re_sign` pins the wording so it cannot soften into "reset it" later.

Re-framing an already-forked chain is also not possible, and the plan note says why: a fork means several entries each claim the same `prev_hash`, so no re-parse recovers a single history. Only re-signing would, and that is the thing that has to stay impossible.

## One consequence worth flagging

The image-lineage self-heal (`heal_ancestry`) reads the same anchor, so a build whose lineage tip lacks an `audit_ref` marker now **fails** against a damaged chain instead of re-emitting the node as an orphan. I believe that is the right direction — healing an ancestry against a ledger you cannot read appends more entries to a log that proves nothing — but it does widen where a broken chain is felt beyond the checkpoint verbs in the issue. A marked tip short-circuits before any lookup, so a healthy tree is unaffected. Called out in plan 302 WS6; say the word if you want it narrowed.

## Tests

12 new, all against a chain **synthesized and then deliberately damaged inside a temp `MVM_HOME`** — nothing depends on the developer's `~/.mvm`. The damage is forgery-shaped (bytes still parse as JSON, signature no longer matches), which is the case the anchor used to silently skip.

`a_miss_against_an_unverifiable_chain_is_an_error_not_a_silent_none`, `a_miss_against_clean_chains_still_reports_never_audited`, `a_hit_still_resolves_even_when_another_chain_is_unverifiable`, `the_image_anchor_reports_an_unverifiable_ledger_the_same_way`, `an_empty_anchor_reports_no_entry_rather_than_an_unreadable_ledger`, `lineage_refusals_separate_an_unreadable_ledger_from_unaudited_and_tampered`, `is_host_lifecycle_chain_accepts_only_the_per_tenant_chain`, `a_quarantined_chain_is_out_of_scope`, plus five doctor mapping tests.

## Validation

nightly `fmt --all --check` clean; `clippy --workspace --all-targets -D warnings` clean; `nextest --workspace` 10602 passed / 24 skipped; `--workspace --doc` clean; BDD 146/146; xtask `check-file-size`, `check-honesty`, `check-no-overclaim`, `check-deferrals`, `check-single-home`, `check-conformance`, `check-abi-layout`, `check-no-spec-refs-in-comments`, `check-claim-catalog`, `check-uniform-vsock-egress`, `check-vsock-only-egress`, `check-no-vz`, `check-no-string-backend-dispatch`, `check-test-home-isolation`, `check-two-surfaces` all pass.

`specs/plans/302-audit-chain-write-path-hardening.md` (new WS6 — this is the read-side counterpart to that plan's WS1–WS3 write-path fixes), `specs/SPRINT.md` and `specs/REFACTOR-STATUS.md` updated in the same commit.

## Note for the host this was found on

`~/.mvm/audit/local.jsonl` there is **forked, not corrupted** — all 40,458 envelopes parse, but 4 distinct `prev_hash` values are each claimed by 3–4 entries (the concurrent-writer race fixed by #2239). No repair is possible. Until it is quarantined (`mv ~/.mvm/audit/local.jsonl ~/.mvm/audit/local.jsonl.forked-2026-05-12`), every checkpoint reads as unanchorable and `checkpoint verify` / `restore` / `fork` fail closed. That is the gate working, and it reproduces on `main`. After this PR the message will at least say so.

### DIFF
--- a/crates/mvm-cli/src/commands/vm/checkpoint/lineage.rs
+++ b/crates/mvm-cli/src/commands/vm/checkpoint/lineage.rs
@@ -451,6 +451,40 @@ mod tests {
         );
     }
 
+    /// The unsigned `secrets.jsonl` operator log lives in the same directory and
+    /// matches the `<tenant>.jsonl` shape, but it is not a chain — no envelope,
+    /// no `prev_hash`, no signature. Admitting it makes it fail verification
+    /// permanently, and after this change that turns every un-anchored record on
+    /// the host into "cannot determine whether this was audited" forever.
+    ///
+    /// Found on a real host, not hypothesised: `mvmctl doctor` reported
+    /// "1 of 3 chain(s) FAIL verification" against it with the message
+    /// `unknown field `action``, which is what a plain operator-log line looks
+    /// like to a chain parser.
+    #[test]
+    fn the_unsigned_secrets_operator_log_never_makes_a_lookup_unanswerable() {
+        let mut env = TestEnv::new();
+        let tmp = tempfile::tempdir().unwrap();
+        env.set("MVM_HOME", tmp.path());
+
+        seed_signed_chain("cp-seeded");
+        // Verbatim shape of the real operator log: plain JSON, no envelope.
+        std::fs::write(
+            tmp.path().join("audit").join("secrets.jsonl"),
+            "{\"action\":\"list\",\"error\":null,\"name\":\"*\",\"outcome\":\"ok\",\
+             \"pid\":24117,\"tenant\":\"local\",\"timestamp\":\"2026-05-15T04:25:53Z\"}\n",
+        )
+        .unwrap();
+
+        let anchor = SignedChainAnchor::load().unwrap();
+        let found = CheckpointChainAnchor::recorded_creation_digest(&anchor, &meta("cp-absent"))
+            .expect("an unsigned operator log is not an unverifiable chain");
+        assert!(
+            found.is_none(),
+            "a clean ledger beside the operator log still answers honestly"
+        );
+    }
+
     /// An anchor built for a resident claim carries no chains at all, so it must
     /// keep answering `None` rather than inventing an unverifiable one.
     #[test]
--- a/crates/mvm-cli/src/commands/vm/checkpoint/lineage.rs
+++ b/crates/mvm-cli/src/commands/vm/checkpoint/lineage.rs
@@ -1,5 +1,3 @@
-use std::path::Path;
-
 use anyhow::{Context, Result};
 use serde::Serialize;
 
@@ -28,12 +26,26 @@ struct CheckpointVerifyJson<'a> {
 /// content-address each checkpoint's creation entry recorded. Only entries from
 /// chains that verify clean are indexed: an entry stranded behind a tampered
 /// line is not trusted, so the checkpoint it names fails closed as un-anchored.
+///
+/// A chain that fails verification is skipped but *remembered*, because the two
+/// ways a lookup can come back empty are not the same finding and must not
+/// share a message. With every chain clean, an empty lookup means the record
+/// was never audited — a step someone skipped, fixable by re-capturing with a
+/// signer present. With a chain unreadable, an empty lookup means only that the
+/// ledger cannot answer: the record may be perfectly well audited behind the
+/// broken line. Reporting the second as the first buries a tamper signal under
+/// a routine-looking message. Both fail closed; only the reason differs, and
+/// only one of them tells you to investigate.
 pub struct SignedChainAnchor {
     /// `<namespace>:<id>` -> content-address the signed chain recorded at
     /// creation. The namespace tag (`checkpoint` / `image`) keeps the checkpoint
     /// and image key spaces structurally disjoint in one map, rather than
     /// relying on their raw id/digest strings happening not to collide.
     recorded: std::collections::HashMap<String, CheckpointDigest>,
+    /// Host-lifecycle chains whose signatures did not verify, in the order
+    /// encountered. Non-empty turns a lookup miss from "never audited" into
+    /// "cannot tell" — see the type docs.
+    unverifiable: Vec<std::path::PathBuf>,
 }
 
 /// Namespace tag for a checkpoint-id key in [`SignedChainAnchor::recorded`].
@@ -55,6 +67,7 @@ impl SignedChainAnchor {
     pub fn empty() -> Self {
         Self {
             recorded: std::collections::HashMap::new(),
+            unverifiable: Vec::new(),
         }
     }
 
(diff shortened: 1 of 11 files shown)


auser:
Pushed a second commit — the live run on a real host caught a false positive this PR would otherwise have introduced, and it was a bad one.

Right after the forked chain on that host was quarantined, the new `doctor` line reported:

```
1 of 3 chain(s) FAIL verification: ~/.mvm/audit/secrets.jsonl
(malformed envelope at line 0: unknown field `action`)
```

`secrets.jsonl` is not a chain. It is the unsigned plain-JSON operator log `mvmctl secret …` writes into the same directory — no envelope, no `prev_hash`, no signature — and it matches the `<tenant>.jsonl` lifecycle shape by accident. A chain verifier pointed at it cannot do anything except fail.

**The exclusion already existed** — in `mvm-client`'s source enumerator, and nowhere else. `SignedChainAnchor::load` had always mis-scanned the file and merely `warn!`-ed, which is exactly why nobody noticed. This PR turns that latent mis-scan into a hard, permanent refusal: with the file recorded as unverifiable, every lookup miss returns "cannot determine whether this was audited", so `checkpoint verify` / `restore` / `fork` — and any build whose image-lineage tip lacks an `audit_ref` marker — fail closed forever on **every host that has ever run `mvmctl secret`**. A false positive with no way out, since the file is not damaged and quarantining it would break the secrets tooling.

Fix: `SECRETS_OPERATOR_LOG` moves to `mvm-core::config` beside `WORKLOAD_AUDIT_SUFFIX`, `is_host_lifecycle_chain` excludes it, and `mvm-client`'s private copy collapses onto the shared one. This is precisely the drift `is_host_lifecycle_chain`'s own doc comment warns about, caught in the act — there were three copies of the rule and the anchor was using the wrong one. The exclusion is the exact filename, not a substring, so a tenant genuinely named `my-secrets` keeps its chain.

Both regression tests were **confirmed to fail with the exclusion removed** — `the_unsigned_secrets_operator_log_is_not_a_lifecycle_chain` on the predicate, and `the_unsigned_secrets_operator_log_never_makes_a_lookup_unanswerable` reproducing the full end-to-end symptom against a synthesized operator log in a temp `MVM_HOME`.

`doctor` on that host now reads:

```
audit chain: ok=True
  2 chain(s) verify against the host signer
```

Re-validated on the updated branch: nightly fmt clean, clippy `-D warnings` clean, `nextest --workspace` 10604 passed / 24 skipped, doctests clean, BDD 146/146, and every `ci.yml` xtask gate passes.

Worth noting for reviewers: the pre-existing behaviour was already wrong here — the anchor was scanning a non-chain and warning about it on every invocation. This PR is what made that visible enough to fix.

Issue #2264, opened 2026-08-09
Test suite: a family of daemon/socket tests fails under parallel load, not from code changes


auser:
## What

Seven tests across four areas fail intermittently when the workspace suite runs under CPU contention, and pass in isolation. They are timing-bound, not logic-bound. Today they cost real debugging time twice — including making a refactor look like it had broken a crate it never touched.

| Test | Area |
|---|---|
| `metrics_server::tests::test_metrics_server_responds` | mvm-cli, binds a port |
| `builder_runner::runner::tests::build_packs_inputs_boots_the_builder_spec_and_extracts_the_output` | mvm-runtime |
| `builder_runner::runner::tests::build_rides_the_closure_nar_on_the_same_input_disk_when_present` | mvm-runtime |
| `workload_input_plane::a_writer_that_let_its_lease_lapse_does_not_wedge_the_stream_for_its_successor` | mvm-hostd, lease expiry |
| `host_agent_restart::daemon_crash_mid_flight_loses_at_most_one_call_and_preserves_chain` | mvm-hostd, daemon restart |
| `host_agent_restart::idle_daemon_self_terminates_when_last_vm_deregisters` | mvm-hostd, daemon restart |
| `host_agent_restart::worker_restart_restores_journaled_registration_and_chain` | mvm-hostd, daemon restart |
| `broker::daemon::tests::helper_restart_rebinds_live_registration_and_preserves_chain` | mvm-hostd, helper restart (**seen on CI**, not just locally) |

## Evidence it is load, not code

The clearest signal is the uniform duration. In one run the three `host_agent_restart` tests failed at **10.02s, 10.05s, and 10.03s** — three independent tests failing at an identical wall time is a timeout, not three logic errors.

Measured on the same machine, same `-j 4`:

| Run | Result |
|---|---|
| Branch, full workspace, machine also running another build | 3 failed (all `mvm-hostd`, all ~10.0s) |
| `main`, full workspace | 10,587 passed |
| Branch, `mvm-hostd` alone | 1,809 passed |
| Branch, full workspace, machine idle | 10,582 passed |

The branch under test (#2250) modified **zero** `mvm-hostd` files — `git diff --name-only origin/main..HEAD` confirms it. Each of the others also passes in isolation: `metrics_server` in 0.024s, `workload_input_plane` in 0.111s, `builder_runner` in 0.174s.

[...]

PR #2269, opened 2026-08-09
fix(test): wait for the old signer helper to release the chain lock

auser:
Unblocks the merge queue. `broker::daemon::tests::helper_restart_rebinds_live_registration_and_preserves_chain` has been failing merge-group CI across several unrelated PRs — #2250 twice and #2242 at least once — always at `daemon.rs:1072`, so nothing could merge.

## The cause

Not a generic flake. The actual error, from the CI log:

```
signer-helper register refused: mvm-audit-signer chain already held by another writer:
/tmp/.tmpdy26oX/local.vm-a.workload.jsonl
```

The test aborts the running signer helper and immediately starts a replacement for the same workload chain. The sole-writer guard — `lock_sole_writer`, an `flock(LOCK_EX|LOCK_NB)` — is released only when the chain's `File` descriptor drops, and `helper_task.abort()` cancels the task **without synchronously closing that descriptor**. On a loaded runner the replacement loses the race and is refused.

That guard is recent (chain-head reads were serialized so concurrent writers cannot fork a chain), which is why this started affecting several PRs at once rather than being long-standing.

## Why the fix is in the test

This is test-only. A real deployment runs the signer helper as its own process, and the kernel releases the `flock` when it exits — a restarted helper always gets the chain. The same-process abort-then-restart pattern is what makes the release non-deterministic, and only the test does that.

So the test waits for the release rather than asserting the first attempt wins. Bounded at ten seconds, and **any error other than sole-writer contention panics immediately** — a real regression still fails fast instead of hiding behind a retry.

## What I got wrong first

Worth recording, since it cost time: I initially attributed this to the load-sensitive test family in #2264 and then, when it recurred, to a socket-readiness race in `start_helper` (which relies on a single `yield_now()` for the spawned task to reach `accept()`). Both were wrong. The `.with_context` message was in the CI log the whole time and identified it exactly. `start_helper`'s `yield_now()` is still a fragile construct, but it is not what fails here and this PR leaves it alone.

## Verification

- `cargo nextest run -p mvm-hostd` — 1,809 passed
- `cargo nextest run --workspace` — 10,587 passed

Related: #2264 tracks the broader family of load-sensitive tests; this is one member with a specific, identified cause rather than a timeout.

### DIFF
--- a/crates/mvm-hostd/src/broker/daemon.rs
+++ b/crates/mvm-hostd/src/broker/daemon.rs
@@ -699,6 +699,30 @@ mod tests {
         }
     }
 
+    /// Rebind registrations, retrying while the previous helper still holds the
+    /// chain's sole-writer `flock`. Bounded: a lock that never frees is a real
+    /// failure and must still surface as one rather than hanging the suite.
+    async fn rebind_once_the_old_writer_releases(d: &HostAgentDaemon) -> usize {
+        let deadline = std::time::Instant::now() + Duration::from_secs(10);
+        loop {
+            match d.rebind_signer_helper_registrations() {
+                Ok(n) => return n,
+                Err(e) => {
+                    let err = format!("{e:#}");
+                    // Only the sole-writer contention is worth waiting out.
+                    // Anything else is a real failure and surfaces now rather
+                    // than after a ten-second retry that hides it.
+                    if !err.contains("already held by another writer")
+                        || std::time::Instant::now() >= deadline
+                    {
+                        panic!("rebind after helper restart failed: {err}");
+                    }
+                    tokio::time::sleep(Duration::from_millis(20)).await;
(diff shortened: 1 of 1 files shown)


auser:
Superseded by #2263, which landed the same fix and diagnosed it more precisely.

#2263 identifies the actual mechanism — `audit_signer::helper::serve` spawns a detached task per connection, each holding its own `SharedSignerHelper` clone, so aborting the accept loop does not release the chain's writer lock. My description stopped at "abort() does not synchronously close the descriptor," which is directionally right but less specific. Its retry is also tighter (100 × 10ms rather than a 10s deadline) and classifies errors the same way, so a genuine regression still fails immediately.

Nothing here to salvage. Closing; branch deleted.

PR #2314, opened 2026-08-10
Retire reqwest: closure 286 → 242 (Plan 309 Phase 2)

auser:
Retires `reqwest`. Shipping closure **262 → 242**; with the earlier phases this
takes Plan 309 from **286 → 242 — a 15% cut**. `CLOSURE_BUDGET` ratcheted.

What leaves is the whole hyper/tower stack: `hyper`, `hyper-util`,
`hyper-rustls`, `tower`, `tower-http`, `tower-service`, `http-body`,
`http-body-util`, the `futures-*` set, `sync_wrapper`, `want`, `try-lock`,
`atomic-waker`, `slab`.

`mvm-http` keeps `http`, `httparse`, `url`, `tokio-rustls`, and
`rustls-platform-verifier` rather than hand-rolling header validation, head
parsing, URL parsing, or trust-store handling — which is why the measured
saving is −20 and not the −27 the raw subtree suggested. The plan carries the
corrected figure.

## The differential harness came first

This is the egress re-origination and SSRF-guarded fetch surface, so the
evidence had to exist before the cut over. `crates/mvm-http/tests/differential.rs`
runs one corpus through **both** clients and compares what a caller can observe.

**Twelve well-formed cases agree exactly** — fixed-length, chunked, chunked with
trailers, chunk extensions, 204/304, redirect-not-followed, error statuses,
empty body, read-until-close, cap boundary, truncated body, garbage head.

Five ambiguous cases are refused here. Only **two** are behavioural divergences:

| case | reqwest | mvm-http |
|---|---|---|
| `0x`-prefixed chunk size | refuses | refuses |
| `+`-prefixed Content-Length | refuses | refuses |
| disagreeing duplicate Content-Length | refuses | refuses |
| **Content-Length + Transfer-Encoding** | **accepts**, prefers TE | refuses |
| **`Transfer-Encoding: gzip, chunked`** | **accepts** | refuses |

The first is the textbook request-smuggling primitive: *resolving* a framing
ambiguity is exactly what lets two hops disagree. Failing closed is the point.
The harness keeps `reqwest` as a **dev-dependency** deliberately — it is
standing evidence, not scaffolding.

## A reqwest workaround deleted rather than ported

`http_hardening.rs` documented that reqwest connects on the **resolver's** port
rather than the URL's, and its `Resolve` trait is never handed the port. So the
SSRF-filtering resolver had to hardcode 443, and any caller forwarding to a
different port needed a *second*, resolver-less builder plus a manual
resolve-filter-and-pin dance.

`mvm_http::Resolve` receives `(host, port)`. Consequently:

- `hardened_client_builder_no_dns` and `resolve_ssrf_safe_ips` are **deleted**;
  one resolver now serves every caller.
- `substitution_proxy`'s per-request resolve-and-pin collapses to a single
  builder call.
- `web_fetch`'s hand-rolled `PinnedDnsResolver` becomes the built-in
  `PinnedResolver`.
- `http_forward` loses five `no_proxy`/`no_gzip`/`no_brotli`/`no_zstd`/
  `no_deflate` calls — mvm-http supports neither proxies nor compression, so
  explicit disabling became absence.

## Naming honesty

`ReqwestHttpFetcher` → `HardenedHttpFetcher`, `ReqwestForwarder` →
`HardenedForwarder`, and comments describing reqwest's *present* behaviour are
corrected. Three references remain and are deliberate: they explain why the
two-builder split existed, which is what makes its removal reviewable.

## Verification

- Workspace `nextest` **10,765 pass**, `clippy --all-targets -- -D warnings`
  clean, nightly `fmt --check`.
- **167 SSRF / egress / substitution / forward tests** pass, including
  `ssrf_guard_rejects_loopback_target`, `ssrf_guard_rejects_aws_imds_target`,
  `ssrf_guard_rejects_rfc1918_private_target`, and
  `substitution_endpoint_refuses_unbound_destination`.
- All policy gates pass, including `check-closure-budget` at its new 242.

## Not in this PR

Phase 3 (product decisions: tree-sitter grammar gating, `toml`,
`tracing-subscriber`, `serde_jcs`) and Phase 4 (a lockfile-count ratchet, so the
~62 `wasmtime`-family packages an off-by-default feature adds stay visible to
`cargo deny`). Both are scoped in the plan.


auser:
Heads-up from Plan 313 Phase 0 (#2320) — this PR is the cheapest place to fix something, and the window closes when it merges.

The replacement forwarder keeps the reqwest shape verbatim:

```rust
let body = resp.bytes().await.map_err(...)?.to_vec();
Ok(ForwardResponse { status, headers, body })
```

That reproduces full response buffering on top of a client that no longer needs it. `mvm_http::Response` already exposes incremental `chunk()` and a `max_bytes` ceiling — both discarded here.

Three things this costs today on the substitution path:

1. **No streaming.** The guest gets nothing until the response completes, so SSE model calls do not stream at all.
2. **Past the forward timeout, a stream fails outright.** `.timeout(...)` is a whole-request deadline including body read, default 30s — an SSE stream held open longer is killed, not just buffered.
3. **Unbounded allocation.** No body cap on this path, in the one process holding workload credentials in the clear. `mvm_http`'s `max_bytes` would fix this for free.

Not asking you to expand this PR's scope — the closure reduction stands on its own and the diff is already large. But `ForwardResponse { body: Vec<u8> }` is the type-level blocker: while that field is an owned `Vec`, no caller can stream, and changing it later means rewriting this function a second time.

If you'd rather keep this PR purely mechanical, that is reasonable — just say so and Plan 313 Phase 1 will pick it up as a follow-on rather than assuming it landed here.

Issue #2318, opened 2026-08-10
The receipt store does two sync_all calls per append: 100ms of a launch's admit, more than the audit chain it mirrors


auser:
The execution-receipt store costs more per launch than the chain-signed audit log it mirrors, and it does so with two full `sync_all` calls where the chain uses one `sync_data`.

## Measurement

KVM host (fsync 41.7 ms p50 on its md2 array), release, warm cache, `machine run --image alpine -- /bin/true`, via the emit-path debug spans added in #2317:

```
emit: tokio runtime build   0.02 ms
emit: signer signer=0      42.3 ms   <- audit chain: one write_all + sync_data
emit: receipt             100.7 ms   <- receipt store
```

So `record_admission` is ~143 ms, and **70% of it is the receipt, not the chain**.

## Why

`AuditEmitter::emit_receipt_for_entry` calls `ReceiptStore::append_chained`, which does:

1. take the per-tenant lock file
2. `read_head()`
3. `write_atomic(receipt_path, ...)` — create temp + write + **`sync_all`** + rename
4. `write_head(...)` → `write_atomic(head_path, ...)` — create temp + write + **`sync_all`** + rename

Two atomic writes, so two fsyncs, for one logical append. And `sync_all` flushes file metadata as well as data, where the audit chain deliberately uses the cheaper `sync_data`.

Only the `plan.admitted` entry produces a receipt today (`audit_entry_to_receipt` returns `None` for the rest), so this is one 100 ms hit per launch rather than per entry — but it lands squarely inside `admit`, which is on the critical path before the VMM starts.

## Options, in rough order of value

1. **Collapse the two fsyncs into one.** The head is a pointer to a receipt that is already durable. If the head write is lost, the head can be rebuilt by scanning the tenant directory for the highest sequence — which makes the head a cache rather than the source of truth, and its fsync unnecessary. That is the change with the best ratio, and it needs someone to confirm nothing treats the head as authoritative in a way that recovery cannot reconstruct.
2. **`sync_data` instead of `sync_all` for the receipt body.** The rename still needs the directory entry durable for true atomicity, so this is only correct alongside a directory fsync — worth measuring whether that is actually cheaper on this hardware before assuming.
3. **Apply #2317's barrier/deferred rule to receipts.** The receipt for `plan.admitted` mirrors an authorization, so it plausibly *should* be a barrier — meaning the honest answer here may be "one fsync, not zero" rather than deferral.

## Worth deciding first

[...]

Issue #2293, opened 2026-08-10
Audit chain fsyncs 8 times per launch: 294ms of admission on Linux vs 38ms on macOS, plus a duplicated plan.admitted


auser:
On Apple Silicon / HVF the `admit` phase of a prepared cold launch is ~38 ms. On the Linux KVM box it is **294 ms** — same code, same warm cache, release binary at `c866611af`. The gap is not compute and not hashing (`artifact_bytes_hashed: 0` on both).

It is `fsync`.

## Measurement

Append-then-fsync on the KVM box's root filesystem (md2 RAID), 20 samples:

```
append+fsync: p50=41.7ms p95=50.1ms max=82.8ms
```

Chain entries appended by exactly one `machine run --image alpine -- /bin/true`:

```
chain entries appended by one launch: 8
    cmd.machine.invoked
    plan.admitted
    plan.oci_provenance
    plan.admitted        <- second one
    plan.policy_resolved
    plan.launched
    plan.boot_posture
    cmd.machine.completed
```

8 × 41.7 ms ≈ **334 ms** of durability per launch, which brackets the observed 294 ms `admit` plus the entries that land later in the run. On macOS the same 8 entries cost a fraction of that, which is why this has been invisible: the platform difference in fsync cost, not the code, is what changes.

The durability itself is deliberate and correct — Plan 303 WS2 made each append land as one `write_all` + `sync_data` specifically so a torn tail reports as truncation rather than tampering. This issue is not asking to weaken that. It is pointing out that the *number of independently fsynced entries per launch* is a design choice nobody has costed, and on a host with real fsync it is the second-largest span in the launch after `driver_boot` (#2292).

## Two separate things to fix

**1. `plan.admitted` is emitted twice per launch.** Once from `emit_oci_run_admission` (the claim-14 OCI provenance path) and once from the real boot admission. Two full admissions, two fsyncs, for one launch. Whether that is a duplicate to remove or two genuinely distinct admissions that should be named differently is a correctness question before it is a performance one — a reader of the chain currently sees two `plan.admitted` for one workload and cannot tell them apart by event name.

[...]

PR #2328, opened 2026-08-11
A lost or corrupt receipt head silently forks the receipt store

auser:
Not the PR I set out to write. #2318 was about the receipt store costing 100 ms of a launch; that framing **did not survive measurement**. What the investigation walked into instead is a correctness bug, which is what this fixes.

## The bug

`read_head` returned `Head::default()` — sequence 0 — whenever `head.json` was missing, truncated, or unparseable, while the receipts it named were still on disk. The next append then reused sequence 1, **overwrote the existing `00000001-*.json`**, and linked to no parent.

The module's own preamble says what that produces:

> Two writers that observe the same head produce two receipts naming the same parent, which every offline verifier is obliged to read as a deleted or reordered receipt.

The per-tenant lock already prevents two concurrent writers reaching that state. A lost head reached it from a **single** writer, with no concurrency at all.

## The fix

`read_head` takes the later of the head file and the highest-sequence receipt actually on disk. The scan reads only `<seq>-<type>.json`; the head, the lock, and temp files share that directory and must not be parsed as receipts.

Three tests, each verified red without the recovery:

- a removed head recovers its sequence and tip, and the next append links to it — `left: 0, right: 3` before the fix
- a rewound head loses to the files
- the scan ignores non-receipt entries

## What I removed from this PR, and why

I first wrote this to drop the head's fsync, on the theory that the receipt store's two `sync_all` calls cost ~100 ms per launch. Then I counted the syscalls on the KVM host:

```
baseline (main):  6 fsync + 7 fdatasync = 13 calls, 0.87 ms total
with the change:  4 fsync + 7 fdatasync = 11 calls, 1.11 ms total
=> 62-146 us per call
```

**Every sync call in an entire launch totals about 1 ms.** The fsync-removal had no justification, so it is not in this change — only the recovery is.

That also invalidates the headline number in #2317, which is already merged. I have posted a correction on #2293: its "42.3 ms barrier vs 1.7 ms deferred" came from one sample of each, and the syscall data contradicts it. The #2317 code is still correct and I would still write it; the saving I claimed for it is not real on this hardware.

## Where that leaves the actual question

`admit`'s ~300 ms is still unexplained. Not the chain fsync, not `restore_cursor` (2.9 MB in 1.3 ms), and now not fsync at all. The `emit: receipt` span measured 100.7 ms while its fsyncs cost microseconds, so that time is elsewhere inside `append_chained` — the flock, the head read, receipt id computation, or signing.

I am not guessing a fourth time. That needs a profile, not a hypothesis.

## Verification

`mvm-hostd` suite, nightly fmt, workspace clippy `-D warnings`.

Refs #2318.

### DIFF
--- a/crates/mvm-hostd/src/audit/emitter.rs
+++ b/crates/mvm-hostd/src/audit/emitter.rs
@@ -970,7 +970,20 @@ impl AuditEmitter {
 /// written, fsync'd, then `rename`d over `path`. A concurrent reader sees
 /// either the old file or the complete new one, never a torn write. The
 /// temp name carries the pid so two publishers don't collide on it.
+/// [`write_atomic`], without the fsync.
+///
+/// Still atomic — the rename gives a reader either the whole old file or the
+/// whole new one. What is dropped is survival of power loss, which is the
+/// right trade only for something reconstructible from data already durable.
+pub(crate) fn write_atomic_unsynced(path: &Path, bytes: &[u8]) -> Result<()> {
+    write_atomic_inner(path, bytes, false)
+}
+
 pub(crate) fn write_atomic(path: &Path, bytes: &[u8]) -> Result<()> {
+    write_atomic_inner(path, bytes, true)
+}
+
+fn write_atomic_inner(path: &Path, bytes: &[u8], sync: bool) -> Result<()> {
     use std::io::Write;
     let dir = path
         .parent()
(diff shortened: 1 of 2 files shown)


auser:
**Do not merge — this branch currently contains an accidental mass revert.** I am force-pushing a corrected version.

I built the commit with `git reset --soft origin/main` while this worktree's `origin/main` ref was stale (18114533b). The resulting commit's tree is that older main plus my change, so merging it would revert everything that landed since — **2452 lines across 71 files**, including #2317, #2314, and the xtask closure-budget work. Git saw no conflict because it is a content-level revert, not a textual one.

The replacement re-applies only the two intended file changes on top of current main. Comment will follow when it is pushed.

auser:
**Corrected and pushed.** The branch now contains one commit touching two files:

```
crates/mvm-hostd/src/audit/emitter.rs       |  19 +++-
crates/mvm-hostd/src/audit/receipt_store.rs | 157 +++++++++++++++++++++--
2 files changed, 166 insertions(+), 10 deletions(-)
```

Rebased onto current main, so #2317, #2314, the xtask closure-budget work, and everything else that landed since are untouched. I rebuilt the change by re-applying the two edits to main's current files rather than replaying my stale tree.

PR #2379, opened 2026-08-11
Rotate the audit log without making removed history silent

### LINKED ISSUE #2365: Audit-log rotation: bound the active chain without making removed history silent
The chain-signed audit log grows without bound. There is no rotation anywhere in
the tree. `checkpoint.forked` is an audit *event type*, not log rotation.

## Measured, not estimated

Release build, this host, against the real `~/.mvm/audit/local.jsonl` and the
real host verifying key (the chain verifies clean, so this is the cost of a
*successful* full walk):

| entries | bytes | bytes/entry | full walk | per entry |
| ------- | ----- | ----------- | --------- | --------- |
| 4,022 | 4,185,564 | 1,040 | 122 ms | 30.3 µs |

Three runs, warm cache, spread under 3 ms. Roughly 29 ns/byte, linear.

Two corrections to figures this was scoped against:

- **30.3 µs/entry, not 115.5 µs**, so 122 ms and not 380 ms.
  `crates/mvm-hostd/tests/audit_verify_cost.rs` — the source of the 115.5 µs
  figure — exists only on the unmerged `perf/2346-audit-chain-incremental`
  branch (#2350), not on main.
- **715 `*.jsonl` files in `~/.mvm/audit/`, not ~1,040.** The directory has
  1,043 entries, but 328 are `gateway-*.sock`. Almost all of the 715 chains are
  zero-byte `*.workload.jsonl` files, one per VM ever started. That per-file
  count is a separate, cosmetic problem and is out of scope here.

The problem is not today's 122 ms. It is that nothing bounds it: at ~1 KB/entry
a 100 MB log is a 3 s walk, and the file only ever grows.

### The 23 MB `.forked-` file is not a rotation precedent

It is a **quarantine** rename, already documented in-tree. `crates/mvm-core/src/config.rs`
carries `a_quarantined_chain_is_out_of_scope`, whose comment says renaming a
broken chain out of `*.jsonl` is the documented recovery step so an operator can
clear a finding without deleting evidence. Nobody rotated; someone quarantined a
chain that had forked. Worth stating because the two shapes look identical on
disk and mean opposite things: a quarantine deliberately drops a file out of
verification scope, and a rotated segment must stay in it.

## The tension

Verification seeds `prev_hash = [0u8; 32]` and requires line 0 to claim it
(`verify_audit_chain_entries`, `crates/mvm-hostd/src/supervisor/audit_file.rs`).
That genesis anchor is what makes removed history detectable. So a naively
rotated file either fails verification forever, or rotation must establish a new
*authenticated* starting point. Get it wrong and "tamper-evident ledger"
silently becomes "ledger you can edit by calling it rotation".

## Options

**A — authenticated segment handoff (proposed).** The retiring segment ends with
a signed `chain.sealed`; the new segment starts with a signed `chain.continued`
whose envelope `prev_hash` is the retired segment's final line hash, and which
*also* carries that hash, the predecessor's sequence number, and its entry count
inside the signed entry body. Verification stays single-file-local: line 0 may
claim a non-genesis predecessor only if it is a `chain.continued` whose signed
body names the identical hash. Forging a handoff needs the signing key — the
same bar as forging any other line.

**B — independent per-file chains + signed manifest.** Each segment restarts at
genesis; a separate signed manifest orders them. Rejected: the ordering lives in
a second artifact that can be deleted independently of the data it orders.
Delete the manifest and you hold N genesis-anchored files with no way to tell
their order or whether any are missing. A puts the same information inside the
chain, where removing it breaks the chain. B is also a second thing to keep in
sync, and this repo has a documented history of a narrative and a ledger
drifting apart.

**C — size/age retention, segments kept.** Not an alternative to A — it is the
*policy* on top of A's *mechanism*. Proposed as the default.

**Proposal: A, with C's keep-everything as the default.**

## What A buys, and what it does not

- Editing any line in any segment — caught, as today.
- Front-truncating a segment — caught at line 0, as today.
- Re-ordering or splicing a segment behind a different predecessor — caught,
  and forging it needs the key.
- Deleting a whole segment from the middle or front — caught **and named**: the
  successor says which sequence it continues, and that segment is absent.
  Removal becomes a self-describing hole instead of silence.
- Deleting the newest segments — **not** caught. Tail truncation is undetectable
  today for the same reason (no external anchor, host holds the key) and this
  does not change it. Stated plainly because rotation makes it easier to do by
  hand.

## Two-tier verification

- `mvmctl doctor` — active segment in full, plus segment *topology*: per
  boundary, verify the two adjacent boundary lines and check the successor's
  signed `chain.prev_tip` against the hash of the predecessor's final line.
  `O(active + segments)`, two lines read per sealed segment, **no cache, no
  trust-on-first-use**. Doctor then has to *say* it attested the active segment
  and the topology, not the sealed interiors.
- `mvmctl trust audit verify` — walks every interior end to end, full cost, by
  design.

## Retention — the question for the maintainer

Rotation deliberately removes history from the active file. This is an
evidence-retention decision, not a perf tweak, so it should not be defaulted
silently.

- [ ] **Keep every segment forever.** Rotation only splits. Disk grows as it
      does today; the *active* segment becomes bounded, so every check that only
      needs the live chain becomes cheap. Deleting a retired segment stays
      possible but is an explicit operator action, never automatic.
- [ ] **Keep N segments, prune older.** Bounded disk. Pruned segments leave a
      detectable, self-describing hole — but the evidence is gone. Needs an N.
- [ ] **Keep by age** (e.g. 90 days). Same property, time-based boundary.

Also: should retired segments stay in `is_host_lifecycle_chain` sweep scope?
Naming them `<tenant>.seg-<NNNNNN>.jsonl` keeps them in scope with no change to
that predicate — the opposite of the `.forked-` quarantine shape, on purpose,
since a retired segment is intact evidence rather than a chain under triage.

## ADR-001 impact, stated rather than done quietly

Claim 8's witness `fn:verify_audit_chain` would attest something different:

- before — "an unbroken chain from genesis"
- after — "an unbroken chain from genesis, **or from a signed handoff naming its
  predecessor segment and that predecessor's final chain hash**"

Claim 14 is unchanged in substance, but a `plan.oci_provenance` entry may live
in a sealed segment, so the claim holds over the segment *set* rather than one
file. Proposed new claim-8 witnesses: `fn:a_spliced_segment_is_refused`,
`fn:a_missing_segment_is_named_not_silently_skipped`,
`fn:a_front_truncated_segment_still_fails_at_line_zero`.

## Other interactions

- **Merkle.** `mvm_hostd::audit::merkle::{read_leaves, build_root_in,
  build_inclusion_in}` read one file. Left alone, the published transparency
  root silently narrows to the active segment on first rotation. `read_leaves`
  must span the ordered set, with globally ordered leaf indices so existing
  inclusion proofs keep meaning what they meant.
- **Atomicity.** Rotation happens inside the exclusive `flock` `sign_and_emit`
  already holds, so two supervisor processes cannot both rotate. A crash between
  the rename and the first write to the fresh file must recover by continuing
  from the highest sealed segment's tip — not by starting a new genesis chain,
  which would silently orphan everything before it.

Plan: `specs/plans/319-audit-log-rotation.md`.

auser:
Closes #2365.

The chain-signed audit log grew without bound and nothing capped it.

## Measured first

Release build, this host, real `~/.mvm/audit/local.jsonl` and real host key
(the chain verifies clean, so this is a *successful* full walk):

| entries | bytes | bytes/entry | full walk | per entry |
| ------- | ----- | ----------- | --------- | --------- |
| 4,022 | 4,185,564 | 1,040 | 122 ms | 30.3 µs |

Two corrections to the figures this was scoped against, both now in the plan:

- **30.3 µs/entry, not 115.5 µs** → 122 ms, not 380 ms. The 115.5 µs number
  comes from `audit_verify_cost.rs`, which exists only on the unmerged #2350
  branch, not on main.
- **715 `*.jsonl` files, not ~1,040.** 328 of the 1,043 dir entries are
  `gateway-*.sock`; almost all the rest are zero-byte `*.workload.jsonl`. That
  file count is a separate, cosmetic problem and is out of scope here.

Also worth stating: the 23 MB `.forked-2026-05-12` file is **not** a rotation
precedent. It is a *quarantine* rename, already documented in-tree by
`a_quarantined_chain_is_out_of_scope`. The two look identical on disk and mean
opposite things — a quarantine deliberately drops a file out of verification
scope, and a rotated segment must stay in it.

The problem was never today's 122 ms. It is that at ~1 KB/entry a 100 MB log is
a 3 s walk, and the file only ever grew.

## The hard part

Verification seeds `prev_hash` with 32 zero bytes and requires line 0 to claim
them. That genesis anchor is exactly what makes removed history detectable, so a
naively rotated file either fails verification forever or throws the anchor
away. Get it wrong and "tamper-evident ledger" quietly becomes "ledger you can
edit by calling it rotation".

Rotation therefore establishes a new **authenticated** starting point. The
retiring segment ends with a signed `chain.sealed`; the fresh one opens with a
signed `chain.continued` whose envelope `prev_hash` is the predecessor's final
line hash — repeated, with the predecessor's sequence number and entry count,
*inside the signed body*. The envelope copy is what the walk consumes; the body
copy is what makes that decision authenticated, because the signature covers the
body. A non-genesis line 0 is accepted only when the body names the identical
hash, so forging a handoff costs the signing key rather than a text editor.

I considered independent per-file chains plus a signed manifest and rejected it:
the ordering would live in a second artifact deletable independently of the data
it orders, leaving N genesis-anchored files with no way to tell what was
missing. The full argument is in `specs/plans/319-audit-log-rotation.md`.

## What it buys, and what it does not

- Editing any line in any segment — caught, as before.
- Front-truncating a segment — caught at line 0, as before.
- Re-ordering, or splicing a segment behind a different predecessor — caught.
- Deleting a segment from the middle or front — caught **and named**: the
  survivor says which sequence it continues and that segment is absent.
- Deleting the **newest** segments — **not** caught. Tail truncation is
  undetectable today for the same reason (no external anchor, host holds the
  key). Rotation changes neither direction of that, but it makes it easier to do
  by hand, so it is stated rather than left implied.

## Retention

**Keep everything**, per maintainer decision on #2365. Rotation only splits;
deleting a retired segment stays an explicit operator action, never automatic.
Retired segments are `<tenant>.seg-NNNNNN.jsonl` and stay in
`is_host_lifecycle_chain` scope with no change to that predicate.

## Two consumers that would have narrowed silently

Neither surfaced from a failing test — both came out of reading the callers:

- **Merkle.** `read_leaves` read one file, so the published transparency root
  would have shrunk to the active segment on first rotation, with the same
  `tree_size` arithmetic over a quietly smaller log and every issued inclusion
  proof pointing at leaf indices that no longer meant the same thing. It now
  spans the ordered set.
- **Client enumerator.** `local.seg-000001.jsonl` would have read as a tenant
  named `local.seg-000001`, failing every entry's tenant-scope check and turning
  a rotated log entirely into refusals. Segments now fold onto their tenant
  through one shared `lifecycle_chain_base` helper.

## Two-tier verification, stated honestly

`mvmctl doctor` verifies the live segment in full plus every handoff —
`O(active + segments)`, two boundary signatures per sealed segment, **no cache
and no trust-on-first-use** — and *says* that is what it checked, so "N chains
verify" cannot imply a full-history attestation it never performed.
`mvmctl trust audit verify` walks every retired interior.
`an_edited_sealed_interior_passes_topology_and_fails_the_full_walk` pins that
limit so nobody upgrades the claim by accident.

## Concurrency and crash safety

Rotation runs under the same lock as the append, so two supervisors cannot both
seal a segment. The lock **moved to a file beside the chain**: it used to be
taken on the data file, which was correct while that file was immortal, but a
`flock` follows the inode and rotation renames it — a second process opening
`<tenant>.jsonl` right after the rename would get a fresh inode and an
uncontended lock.

A crash between the rename and the continuation write is repaired *before* any
append rather than detected after: recovery continues from the highest sealed
segment instead of starting a fresh genesis chain, which would orphan every
segment behind it while each one still verified perfectly alone. That is the one
failure mode here that would destroy evidence without producing an error, so it
has its own test.

## ADR-001

Amended explicitly, not quietly. Row 8's `verify_audit_chain` now attests an
unbroken chain from genesis **or from a signed handoff naming its predecessor
and that predecessor's final chain hash**, with four new witnesses. Row 14's
provenance entries may live in a retired segment, so that claim holds over the
segment *set* rather than one file.

## Tests

23 new tests. The headline one demonstrates the failure mode against real bytes
rather than describing it in a comment:
`naively_dropping_old_entries_fails_verification_at_line_zero` takes a valid
chain, drops the oldest entries, and asserts the survivors — every one correctly
signed and correctly linked — still fail at line 0.

## Gate

- `cargo nextest run --workspace` — 11,029 passed, 0 failed

Two mvm-client tests failed on the first full run and are fixed: my segment
walker's structural pre-reads tripped before the chain verifier, downgrading a
precise `Truncated`/`Malformed` line refusal to a vague I/O error. The verifier
now runs first so its per-line verdict wins — that distinction is what lets an
operator tell a crashed writer from a tampered log.

Re-verified green after rebasing onto current main.


auser:
Heads-up on an overlap, plus a result you may find useful.

**#2385 is ahead of this PR in the merge queue** (positions 1 and 7 as of now). It adds `crates/mvm-hostd/tests/audit_chain_frozen_bytes.rs` — two committed fixtures plus six assertions that freeze the audit writer's exact output and require both verifiers (mvm-hostd's and mvm-contract's `no_std` one) to accept them. It's plan 320 E3.1, groundwork for relocating the audit writer's pure core into `mvm-contract`.

Since it lands first and this PR restructures the writer, I checked rather than guessed: I merged `test/audit-chain-byte-fixture` into this branch locally and ran it.

```
running 6 tests
test the_writer_still_produces_the_frozen_bytes ... ok
test the_host_verifier_accepts_the_frozen_chain ... ok
test the_no_std_verifier_agrees_with_the_host_verifier ... ok
test a_pre_canonical_chain_still_verifies ... ok
test the_no_std_verifier_also_accepts_the_pre_canonical_chain ... ok
test a_flipped_byte_in_the_frozen_chain_is_caught_at_its_line ... ok

test result: ok. 6 passed; 0 failed
```

Clean merge, all green. **No action needed** — rotation preserves the per-line envelope format, and the frozen chain is three entries so it never reaches a rotation boundary. Worth stating positively: this change does not alter whether previously-written audit logs still verify, and there is now a test that would have said so if it did.

Two things to be aware of going forward:

1. If a later revision changes what `sign_and_emit` emits per line, `the_writer_still_produces_the_frozen_bytes` will go red. That is the test working, not a flake. There's a `REGENERATE` const at the top of the file for deliberate updates — but a change there means a log written before it and one written after are different bytes for the same entries, so it wants a look at the diff rather than a reflexive regenerate.
2. The `a_pre_canonical_chain_still_verifies` case is the sharp one. Pre-`canonical` lines carry no signed bytes, so the verifier re-serializes `entry` — a serde change to `AuditEntry` doesn't just change new output, it stops evidence somebody already holds from verifying.

Separately: plan 320 E3 wants to move `audit_file.rs`'s pure core (envelope construction + `hash_line`) into `mvm-contract` and retire `MirrorEntry`. That overlaps this PR's files directly, so E3 is on hold until this lands, and the E3 moves/stays boundary will be re-validated against `audit_file.rs` as it looks afterwards. No claim on your sequencing — just so it's not a surprise later.

PR #2407, opened 2026-08-12
Attribute a spliced egress error to the half that produced it

auser:
## Recovered work

This commit was written for #2395 and **never landed**. It was pushed to
`fix/egress-budget-reset-and-builder-tier` after that PR squash-merged, so it
sat on a base that no longer led anywhere while GitHub showed the PR green and
merged.

Verified three ways before recovering it, rather than trusting PR state:

- `main` carries #2395's squash (`5251cd296`)
- the `main`-vs-branch diff for these two files is exactly this commit's 203 lines
- `ProxyEnd` / `TagEnd` appear nowhere in `main`

Cherry-picked onto current `main`; applied cleanly.

## What it does

The guest proxy logged every splice failure as "egress client connection
failed" at WARN. `copy_bidirectional` collapses both directions into one opaque
`io::Error`, so an in-guest client hanging up mid-transfer — routine; it got
what it wanted, or gave up — read exactly like the host resetting the stream,
which is a real fault.

That ambiguity had already cost debugging time: a healthy build that downloaded
every derivation with zero retries still emitted a `Broken pipe` warning, and
nothing in the message said which end broke.

Each half is now wrapped so its errors carry a `ProxyEnd`. A client-side hangup
logs at debug; anything else stays a warning. A host-side reset — the shape the
egress-budget bug in #2395 produced — remains loud.

`TagEnd` requires `Unpin`, which is not a new constraint: `copy_bidirectional`
already demands it of both halves, so this adds no pin-projection dependency.

## Verification

- `cargo nextest run -p mvm-agentd` — 766 tests, all pass
  (two `mvm-guest-agent` timing tests, `run_detached_spawns_redirects_console_and_reports_exit`
  and `run_shutdown_hook_via_lifecycle_api_writes_marker`, timed out on the first
  run under parallel load from a concurrent build; both pass in ~1.5s when run
  isolated, and neither touches the egress path)
- nightly `cargo fmt --all -- --check` — clean

### DIFF
--- a/crates/mvm-agentd/src/egress_client.rs
+++ b/crates/mvm-agentd/src/egress_client.rs
@@ -833,6 +833,21 @@ fn spawn_default_dns_stub() -> Option<tokio::task::JoinHandle<()>> {
     }))
 }
 
+/// An in-guest client hanging up mid-proxy is routine — it got what it wanted,
+/// or gave up. Only a fault on the host side, or one we cannot attribute, is
+/// worth a warning; logging both alike is what made a healthy build look like a
+/// failing egress path.
+fn log_serve_failure(error: &std::io::Error) {
+    use crate::guest_vsock_session::{ProxyEnd, is_peer_hangup, proxy_end_of};
+
+    match proxy_end_of(error) {
+        Some(ProxyEnd::Client) if is_peer_hangup(error) => {
+            tracing::debug!(error = %error, "egress client disconnected");
+        }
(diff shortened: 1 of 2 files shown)


auser:
Duplicate of #2417 — the egress attribution change in `egress_client.rs` / `guest_vsock_session.rs` is byte-identical between the two branches, and #2417 is based on current main. Closing this one in favour of #2417.

Issue #2368, opened 2026-08-11
Plan 316: Single flow-aware vsock networking path


auser:
Tracking issue for `specs/plans/316-single-flow-vsock-networking.md`.

## Decision

Every untrusted workload has exactly one possible path to or from an external network:

```text
guest loopback adapter
  -> authenticated FlowMux session on GuestService::NetworkFlow (vsock port 5253)
  -> one per-VM mvm-network-endpoint
  -> canonical policy, DNS, substitution/redaction, rate and audit pipeline
  -> host-originated TCP/UDP socket or host-owned ingress listener
```

`NetworkMode`, `L3Vsock`, `HostVsockProxy`, `raw_ip_stack`, the guest `mvm0` TUN,
`mvm-net-agent`, `mvm-netd`, `NetworkControl`, `NetworkData`, host TUN, nftables
forwarding, and the smoltcp datapath are deleted. `Off` is absence of a network
grant and absence of `NetworkFlow`; it is not a second networking mode.

The endpoint is flow-aware at L4, not universally L7. TCP and UDP payloads whose
plan requests no content transformation are relayed without HTTP parsing. L7
parsing, TLS termination, credential substitution, reversible replacement, and
redaction run only for an explicitly typed HTTP/connector flow whose signed plan
requires them.

This plan supersedes the production-path decisions in Plan 285 (`l3-vsock`) and
Plan 287 (the userspace socket datapath).

## Non-negotiable invariants

1. **One external socket owner.** Only `mvm-network-endpoint` may call outbound `connect` or bind workload ingress listeners.
2. **No guest network device.** vsock and no virtio-net, TAP, bridge, TUN-backed hypervisor device, slirp, passt, vpnkit, or QEMU user network.
3. **No raw-packet protocol.** No production guest/host protocol carries IPv4 or IPv6 packets.
4. **Default deny.** Endpoint loss, authentication failure, policy failure, mandatory-audit failure, and transform failure all fail closed.
5. **Host-owned destination.** The host resolves DNS, pins the address, revalidates redirects, applies private/link-local/loopback/metadata denies, and originates the socket.
6. **Secrets remain host-side.** Raw secret values never enter guest memory, disks, logs, protocol frames, errors, metrics, or audit entries.
7. **Transform honesty.** Replacement/redaction is guaranteed only for a typed transform flow.
8. **Ingress is declared.** The host binds exactly the signed address, port, and protocol.
[...]

PR #2408, opened 2026-08-12
Add signed transport-neutral network limits

### LINKED ISSUE #2370: Plan 316 Phase 1 — Pin protocol, resource, and performance baselines
Phase 1 of the Plan 316 umbrella #2368 — `specs/plans/316-single-flow-vsock-networking.md`.

**Blocked by #2369 (Phase 0).** Do not start until Phase 0 has merged.

## Phase 1 — Pin protocol, resource, and performance baselines

- [ ] Add `mvm-contract::protocol::network_flow` with the version-1 frame header,
      opcodes, strict state-independent decoding, and the exact limits above.
- [ ] Add roundtrip tests for every frame, default/value tests, unknown
      opcode/flag/version rejection, truncated header/body rejection, cap-before-
      allocation, invalid stream IDs, oversized DNS/head/datagram rejection,
      and cross-endian golden byte fixtures.
- [ ] Add a state-machine validator that rejects data-before-open, duplicate
      open/close, credit overflow, stream-ID parity violations, reuse after
      reset, and host-initiated opens not backed by a declared listener.
- [ ] Move `max_flows` and the other endpoint resource ceilings out of
      `L3NetworkSpec` into a transport-neutral, signed `NetworkLimits` plan type
      before any FlowMux runtime consumes them. Defaults preserve the current
      4,096-flow ceiling; serde omits default values so existing signed-plan
      bytes remain stable until the intentional plan-version migration.
- [ ] Add `fuzz_network_flow_decode` and `fuzz_network_flow_state`; seed them
      with every valid frame class plus malformed length and transition cases.
- [ ] Extract the existing authenticated-session handshake and encrypted frame
      machinery into a transport-independent unit shared by control RPC and
      FlowMux. Prove wrong boot ID, wrong plan digest, wrong key, replayed
      sequence, tampered ciphertext, expired session, and counter exhaustion
      fail before dispatch.
- [ ] Add a hermetic `xtask network-perf` harness that runs legacy raw TCP,
      legacy typed HTTP, legacy UDP, and the new FlowMux equivalents against
      loopback fixtures. It records JSON containing host/arch, build profile,
      payload size, concurrency, p50/p95 connect latency, p50/p95 request
      latency, throughput, CPU time, peak RSS, and bytes copied.
- [ ] Record 30-sample release-build legacy baselines on Linux x86_64 and macOS
      arm64 before changing the endpoint. Store the result under
      `specs/benchmarks/network/` with the source commit and command embedded in
      the JSON. The harness refuses to compare different hosts, architectures,
      profiles, or payload/concurrency matrices.

## Protocol contract this phase pins

The wire contract lives in a new `mvm-contract::protocol::network_flow` module so
guest and host share one `#![no_std] + alloc`, `forbid(unsafe_code)` codec. It uses
a bounded binary header and binary payload rather than JSON/base64 on the data
plane. Protocol version 1 frame classes:

- session: `Hello`, `HelloAck`, `GoAway`;
- TCP: `OpenTcp`, `Opened`, `Refused`, `Data`, `WindowUpdate`, `HalfClose`, `Reset`;
- UDP: `OpenUdp`, `UdpOpened`, `UdpSend`, `UdpRecv`, `CloseUdp`;
- DNS: `Resolve`, `Resolved`, `ResolveRefused`;
- ingress: `InboundOpen`, `InboundReady`, `InboundRefused`, followed by the same TCP `Data`/credit/close frames; and
- typed transforms: `OpenHttp`, streaming request/response head and body frames, and a terminal `HttpComplete` or `Refused`.

The fixed header carries version, opcode, flags, stream ID, and payload length.
Decoding checks the payload cap before allocation, rejects stream ID zero for flow
frames, rejects unknown opcodes and flags, and rejects frames illegal for the
stream's current state. Stream IDs are odd for guest-initiated flows and even for
host-initiated ingress, preventing simultaneous-open collisions. Data credit is
per stream, so one stalled response cannot block unrelated flows.

Version 1 limits a frame payload to 64 KiB, a DNS name to 253 bytes, a single HTTP
head to 64 KiB, and a UDP datagram to 65,507 bytes. The signed plan's existing
`max_flows` default of 4,096 remains the aggregate TCP/HTTP ceiling; UDP
associations are capped at 256; declared ingress listeners retain the existing
userspace ceiling of 16. This phase adds compile-time and runtime tests that these
constants are included in the endpoint memory ceiling.

Authentication binds the session to the host-owned
`VmInstanceIdentity { node_id, vm_id, boot_id, plan_digest }`. Reuse the existing
signed challenge, ephemeral X25519 key agreement, AES-GCM frame protection,
monotonic sequence, and replay rejection from the authenticated control session;
extract the transport-independent session machinery rather than copy it. The
endpoint receives the expected identity and verification material from the
admitted launch state, never from guest bytes.

## Acceptance criteria

- Every checkbox above is complete with its tests green; no checkbox is ticked
  before its tests pass.
- The codec builds under `#![no_std] + alloc` with `forbid(unsafe_code)` and on
  `wasm32-unknown-unknown`, matching the rest of `mvm-contract`.
- The exact version-1 limits above are asserted by test, not merely documented.
- Both fuzz targets build and run under the pinned fuzz nightly with seeded corpora
  committed.
- Existing signed-plan bytes still verify after the `NetworkLimits` move.
- Baselines are committed under `specs/benchmarks/network/` for Linux x86_64 and
  macOS arm64, each with source commit and command embedded, and the harness
  refuses cross-host/arch/profile/matrix comparison.
- Claim 5 (fuzzed decoders) and claim 8 (plan-derived capability) preserved.
- Repository Definition of Done, including the Linux-only gates in the project
  builder VM, and Plan 316 / `specs/SPRINT.md` / `specs/REFACTOR-STATUS.md`
  updated together.

auser:
## Summary

- add signed, transport-neutral TCP, UDP, DNS, and ingress resource limits to ExecutionPlan
- preserve legacy signed L3 plan bytes through a compatibility projection
- make netd consume the effective transport-neutral TCP-flow limit
- update Plan 316 and delivery tracking for the completed Phase 1 slice

## Security

- limits are part of the signed execution plan and reject zero-valued configurations
- existing signed L3 plans retain their prior serialized representation and effective limits
- the runtime has one compatibility accessor, avoiding divergent enforcement reads

## Verification

- cargo test -p mvm-contract plan::
- cargo test -p mvm-core -p mvm-vmm -p mvm-hostd
- cargo check -p mvm-cli --all-targets
- git diff --check

Closes #2370
Contributes to #2368


auser:
Heads-up on an interaction with #2410 (Plan 316 Phase 1a, the FlowMux wire contract), which was written in parallel and targets the same phase.

#2410 adds `mvm-contract::protocol::network_flow::limits` with the same numbers this PR hardcodes plan-side:

| this PR | `network_flow::limits` |
|---|---|
| `max_tcp_flows` default 4096 | `MAX_TCP_FLOWS` |
| `max_udp_associations` default 256 | `MAX_UDP_ASSOCIATIONS` |
| `max_ingress_listeners` default 16 | `MAX_INGRESS_LISTENERS` |

Whichever of the two merges second should make the plan-side defaults read the protocol constants rather than restate them — two sources for the same ceiling is precisely the drift the reuse-first rule exists to prevent, and here it would mean the admitted ceiling and the enforced one could diverge silently.

Two smaller things worth considering while this is open:

- There is no `max_stream_credit_bytes`, so the per-stream credit window is not plan-admitted. `network_flow` enforces `MAX_STREAM_CREDIT` as a protocol constant, which works, but it means that one bound is not expressible in a signed plan the way the others are.
- `network_flow::limits::MAX_FLOW_CREDIT_BYTES` derives the endpoint's worst-case flow-state memory from the ceilings. If the plan can lower the ceilings, that derivation wants to be a method on `NetworkLimits` too, so the memory bound follows the admitted values rather than the protocol maximums.

No action needed on my side — this PR is the Phase 1b work and #2410 does not touch `NetworkLimits`. Flagging so the reconciliation is deliberate rather than discovered at rebase.

Issue #2371, opened 2026-08-11
Plan 316 Phase 2 — Introduce the one authenticated endpoint without changing callers


auser:
Phase 2 of the Plan 316 umbrella #2368 — `specs/plans/316-single-flow-vsock-networking.md`.

**Blocked by #2370 (Phase 1).** Do not start until Phase 1 has merged.

## Phase 2 — Introduce the one authenticated endpoint without changing callers

- [ ] Rename the production role from `mvm-substitution-endpoint` to
      `mvm-network-endpoint`, including Cargo bin declarations, release
      packaging, updater manifests, helper resolution, confinement profiles,
      scripts, process reaping, metrics labels, and operator diagnostics.
- [ ] Rename `GuestService::Substitution` to `GuestService::NetworkFlow`, retain
      port 5253, and delete the hand-maintained duplicate
      `EGRESS_VSOCK_PORT` constant in favor of the typed service mapping.
- [ ] Rename `EndpointSpawner`/`RealEndpointSpawner` to
      `NetworkEndpointSpawner`/`RealNetworkEndpointSpawner`; keep exactly one
      production `spawn` implementation in `WorkloadRunner`.
- [ ] Make the endpoint authenticate one long-lived FlowMux session before it
      accepts any flow frame. A failed or missing session prevents workload
      readiness when the signed plan grants networking.
- [ ] Add bounded per-stream registries, odd/even ID allocation, independent
      credit windows, cancellation-safe teardown, and endpoint-wide graceful
      shutdown. No lock guard may cross an await.
- [ ] Keep the legacy raw/WireRequest dispatch behind an internal transition
      adapter for this phase only. It must call the new endpoint's canonical
      admission and socket-owner functions; it may not retain an independent
      connect, bind, DNS, rate, or audit implementation.
- [ ] Add tests proving Firecracker, HVF, and libkrun specs expose exactly one
      `NetworkFlow` service when networking is granted, none when it is absent,
      and never expose L3 control/data services.

## Acceptance criteria

- Every checkbox above is complete with its tests green.
- Exactly one production network endpoint bin and exactly one production `spawn`
  implementation exist after the rename; no compatibility alias for the old bin
  name survives on the production path.
- A workload whose signed plan grants networking does not reach ready when the
  FlowMux session fails to authenticate — fail closed, not degrade.
- The transition adapter holds no independent connect, bind, DNS, rate, or audit
  implementation; this is asserted by test, and the adapter is explicitly
  scoped to this phase only.
[...]

Issue #2369, opened 2026-08-11
Plan 316 Phase 0 — Ratify the invariant and freeze expansion


auser:
Phase 0 of the Plan 316 umbrella #2368 — `specs/plans/316-single-flow-vsock-networking.md`.

**This is the only phase actionable at open time.** Phases 1–8 are blocked on it.

## Phase 0 — Ratify the invariant and freeze expansion

- [ ] Add an accepted ADR that records the L4-with-selective-L7 decision, the
      impossibility of arbitrary guest TLS plus host replacement without TLS
      interception, and the rejection of a host MITM CA as the universal path.
- [ ] Mark ADR-036 and ADR-037 superseded for production workload networking;
      retain their measurements and historical rationale without describing
      either implementation as live.
- [ ] Reconcile `specs/refactor/03-networking.md`, ADR-001's backend matrix and
      claim-10 section, `specs/SPRINT.md`, and `specs/REFACTOR-STATUS.md` with
      the actual two-path starting state and this target invariant.
- [ ] Add a temporary CI ratchet that forbids new non-test references to
      `L3Vsock`, `raw_ip_stack`, `NetworkControl`, `NetworkData`, `spawn_netd`,
      and `host_datapath` outside the files scheduled for deletion by this
      plan. The allowlist is an explicit file list and can only shrink.
- [ ] Freeze Plan 285 and Plan 287: no feature work lands on their runtime path;
      only security fixes needed to keep the tree safe during migration may
      modify it.
- [ ] In the first implementation change, make synthesis and admission reject
      `raw_ip_stack=true`/`L3Vsock` with an error naming the supported loopback
      proxy and typed-connector alternatives. Existing running VMs may drain,
      but no new production L3 workload may boot while FlowMux is built. This
      ensures the migration never operates three live production paths.

## Acceptance criteria

- The new ADR is `Status: Accepted` and is discoverable by `xtask check-adr-coverage`.
- ADR-036 and ADR-037 carry an explicit superseded-for-production marker; neither
  describes its implementation as live on the production workload path.
- `specs/refactor/03-networking.md` no longer claims the L3 path was deleted; it
  describes the actual two-path starting state and the Plan 316 target invariant.
- The ratchet gate fails on a newly added non-test reference to any of the six
  forbidden symbols outside the allowlist, and passes on the tree as it stands.
  The allowlist is an explicit file list that can only shrink.
- A synthesis or admission request carrying `raw_ip_stack=true` or
[...]

PR #2410, opened 2026-08-12
Pin the FlowMux wire contract: framing, opcodes, and the state machine (Plan 316 Phase 1)

auser:
Plan 316 Phase 1, first of three changes in that phase.

- Umbrella: #2368
- This phase: #2370
- Predecessor: #2369 (Phase 0), merged as `914a16c0c`

Adds `mvm-contract::protocol::network_flow` — one definition of what a workload
networking frame is, shared by guest and host. `no_std + alloc` and
`forbid(unsafe_code)` like the rest of the crate, because the guest side links
into the sealed agent. **Nothing consumes it yet**; Phase 2 (#2371) wires it to
an endpoint.

## Why the decoder and the state machine are separate

The decoder answers "is this a well-formed v1 frame" from bytes alone. The
state machine answers "is this frame legal right now, from this side". A frame
can be perfectly well-formed and still be an attack, and conflating the two
checks is how one of them ends up not really running. The split is also what
makes each fuzzable alone: the decoder needs no session, the state machine
needs no bytes.

## Decisions worth reviewing

**The length field is a `u32`, not the packet tunnel's `u16`.** The payload cap
is 64 KiB — one byte past what a `u16` expresses. A cap the length field cannot
represent is not really enforced at the parse boundary, so the header grew the
field rather than the cap shrinking to fit it.

**Stream-ID reuse is refused by a per-parity high-water mark, not a set.**
Remembering every retired ID is an unbounded allocation the peer drives — the
exact shape of bug this protocol exists to avoid. The watermark is O(1) and
additionally refuses re-opening any *lower* ID, which is strictly stronger than
remembering the ones actually used.

**Stream IDs are parity-split** — odd guest-initiated, even host-initiated,
zero the session's — so the two sides allocate from disjoint spaces and
simultaneous-open cannot occur. Both directions of the class check are enforced
so neither side can smuggle a frame into the wrong machine.

**`Opened` and `Refused` are shared by TCP and typed HTTP,** because from the
transport's point of view those opens are the same act: the host connected or
it did not. That is why a confirmation cannot be checked against the class
field alone, and why `Opcode::confirming_classes()` spells the relation out. A
first draft classed `Opened` as TCP-only; `a_typed_http_flow_walks_head_body_complete`
caught it.

**An `InboundOpen` must name a mapping the signed plan declared,** so a listener
nobody signed for cannot be conjured by sending a frame about it. The default
validator declares nothing and refuses every one.

**A refused frame leaves the validator byte-identical** — asserted by
`a_refused_frame_leaves_the_validator_unchanged` and continuously by the state
fuzz target. Otherwise a peer could drive state with frames it is not allowed
to send.

**`MAX_FLOW_CREDIT_BYTES` is derived** from the flow ceilings and the per-stream
window rather than restated beside them, so raising a ceiling raises the
endpoint's memory bound automatically instead of silently desynchronising.

## Tests

87 unit tests. Decoder: round-trip for every opcode empty and with payload,
maximal payload, truncated prefix/body always `Incomplete` and never anything
else, oversized length refused from the prefix alone before the body arrives,
bad magic, every bad version, unknown opcodes, every reserved flag bit, wrong
header length, payload/frame length disagreement, both stream-ID class
violations, back-to-back decode, hand-written golden bytes, explicit big-endian
field placement, pinned opcode discriminants. State machine: handshake ordering,
post-`GoAway` silence, every directional opcode from the wrong side,
data-before-open and before-confirmation, duplicate open, reuse after reset,
parity violations, independent parity watermarks, undeclared and missing
ingress mappings, credit consumption/grant/exhaustion/ceiling/`u32` overflow,
empty grants, per-stream credit isolation, half-close semantics, class
discipline, the class ceilings, fail-closed on every refusal, and 10,000
open/reset cycles that do not grow the validator.

## Fuzzing

New `crates/mvm-contract/fuzz` crate with `fuzz_network_flow_decode` and
`fuzz_network_flow_state`, 95 committed seeds (one valid frame per opcode, the
malformed-length family, one transition sequence per flow class), both wired
into `security.yml`'s fuzz lane and added to the workspace `exclude` list.

The decode target asserts the payload cap holds, the consumed span stays inside
the buffer, truncation always reports `Incomplete`, a bumped version is always
refused, and — the one worth calling out — that a decoded frame re-encodes to
exactly the bytes it decoded from. That makes a decoder which "helpfully"
normalizes a malformed field a fuzz failure rather than a quiet place where the
two ends of a connection disagree about what they just exchanged.

Local runs, both clean:

- `fuzz_network_flow_decode` — 18,461,400 executions / 301s
- `fuzz_network_flow_state` — 2,876,122 executions / 301s

## Validation

- `cargo fmt --all -- --check` — clean
- `cargo nextest run --workspace` — **11446 passed, 0 failed**, 26 skipped
- `cargo build -p mvm-contract --lib --target wasm32-unknown-unknown` — clean,
  so the new module holds the crate's `no_std` + wasm contract
- `cargo +nightly check` in `crates/mvm-contract/fuzz` — both targets compile
- xtask gates green: `check-l3-expansion-freeze`, `check-forbidden-deps`,
  `check-core-runtime-free`, `check-workflow-paths`,
  `check-no-spec-refs-in-comments`, `check-honesty`, `check-sprint-append`

## Docs

Plan 316's four delivered Phase 1 checkboxes ticked with a "landed so far" note,
`specs/REFACTOR-STATUS.md` updated, and a delivery entry at
`specs/sprint/delivery/2370-network-flow-wire-contract.md` — following the
per-file convention #2392 introduced while this was in flight, rather than
appending to `specs/SPRINT.md`.

## Remaining in Phase 1

`NetworkLimits` (the transport-neutral signed plan type), the
authenticated-session extraction, and the `xtask network-perf` harness with its
Linux x86_64 / macOS arm64 baselines. Each lands as its own PR against #2370.


auser:
Note for whoever reviews this alongside #2408 (Plan 316 Phase 1b, `NetworkLimits`), written in parallel against an older base.

Both PRs land the same three numbers — 4096 aggregate TCP/HTTP flows, 256 UDP associations, 16 ingress listeners. This PR has them as protocol constants in `network_flow::limits`; #2408 has them as plan-side serde defaults. Neither reads the other, because #2408 branched before this module existed.

Whichever merges second should make the plan defaults read the protocol constants. I have left the same note on #2408. Filed as a follow-up on #2370 so it does not depend on either reviewer noticing.

The two PRs do not otherwise overlap: this one adds no plan type and #2408 does not touch the codec. They do both touch `crates/mvm-contract/src/plan/mod.rs` and the Plan 316 doc, so the second one through will need a rebase.

PR #2467, opened 2026-08-14
feat(cli): machine fork/restore with --as and --branch

auser:
Expose `mvmctl machine fork <parent> --as <child>` and `mvmctl machine restore <checkpoint> --as <child>` over the consolidated checkpoint/`VmBackend` seam.

- `fork` snapshots a running machine as `vm_full`, then branches it into a fresh child VM.
- `restore` branches an existing `vm_full` checkpoint into a fresh child VM.
- Both support `--branch <slug>` for auto-naming dev sandboxes.
- Every child gets a new identity, authority, and per-instance secrets via the existing admit-and-boot path (`fork_vm_full_machine`).
- Removed the hidden folded `machine restore` same-identity verb; same-identity restore remains available as `machine checkpoint restore`.
- Added unit tests, `mvmctl` integration tests, BDD help scenarios, and updated the audit-posture table.


auser:
Opened accidentally from the already-merged feat/run-first-cli branch. The clean follow-up is #2466.

PR #2480, opened 2026-08-14
Feat/316 guest flowmux adapter

auser:


auser:
Merged `main` in and resolved the conflicts so this can enter the merge queue. One commit was deliberately dropped — flagging it here so it isn't a surprise.

**Dropped:** `1b35cb340 refactor(mvm-hostd): delete raw-egress path and EgressMode`

FlowMux is not on the production path yet, so deleting the raw-egress dispatcher would remove the only mode production egress currently has:

- `workload_runner/runner.rs` — `RealNetworkEndpointSpawner::spawn` hard-codes `flowmux_identity: None`
- The only `FlowMuxIdentitySpawnConfig` construction in the workspace is inside `#[test] fn endpoint_config_json_emits_flowmux_identity`
- `claim.rs` still selects the mode with `let raw_egress = inputs.secrets.is_empty()`

This is the same conclusion #2487 already recorded in `REFACTOR-STATUS.md` ("cannot be executed as written — it would break all egress"), and it still holds against current `main`. `plans/316` line 388 keeps `- [ ] Delete EgressMode, raw_egress…` unchecked.

**Kept:** the other 24 commits — guest FlowMux client, stream/UDP handles, reconnect owner, frame sealing, peer-reset handling — plus the `dns_resolver` module, which `supervisor/flowmux.rs` imports and which is purely additive.

Note that `1b35cb340` bundled the deletion together with its replacement (`dns_resolver.rs`, the endpoint spawn rewrites, `network_endpoint_bin.rs` tests), so re-landing it is a single cutover once `flowmux_identity` is wired through — not just re-applying a delete.

Verified: workspace clippy `-D warnings` clean; 4,168 tests pass across mvm-agentd/mvm-hostd/mvm-core; `check-vsock-only-egress`, `check-uniform-vsock-egress`, `check-no-gateway-names`, `check-l3-expansion-freeze`, `check-build-egress-callers` all pass. The port constant still resolves through the single `NETWORK_FLOW_PORT` definition, so #2481's consolidation is preserved.

PR #2492, opened 2026-08-14
fix(hostd): survive transient accept errors on every egress listener

### LINKED ISSUE #2485: Production egress accept loops exit permanently on transient accept errors
## Problem (confirmed by audit, 2026-08-14)

Five production accept loops on the egress path give up **permanently** on any `accept(2)` error:

| location | loop |
|---|---|
| `crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs:791` | UDS substitution (Firecracker / libkrun) |
| `crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs:812` | vsock substitution (QEMU) |
| `crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs:863` | transparent egress terminator |
| `crates/mvm-hostd/src/supervisor/raw_egress.rs:77` | raw egress UDS |
| `crates/mvm-hostd/src/supervisor/raw_egress.rs:279` | raw egress vsock |

Each is `Err(e) => { warn!(...); return; }`.

`accept(2)` fails transiently in normal operation: `ECONNABORTED` (peer went away between SYN and accept), `EMFILE` / `ENFILE` (fd pressure), `EINTR`, `ENOBUFS`. Any one of them ends that VM's accept loop for the life of the VM. The endpoint process stays up, so nothing crashes and nothing is audited — the workload just silently stops reaching the network. On the two substitution loops, that is the credential path.

The codebase already has the correct shape elsewhere — `crates/mvm-hostd/src/stream/serve.rs:182` and `crates/mvm-hostd/src/supervisor/gateway_audit.rs:121` both log, back off, and continue. These five are the divergence.

## Fix

One shared accept-error classifier plus bounded backoff, applied to all five loops rather than five separate edits:

- **Transient** (`ECONNABORTED`, `EINTR`, `EMFILE`, `ENFILE`, `ENOBUFS`, `WouldBlock`) → warn, back off, continue.
- **Fatal** (`EBADF`, `ENOTSOCK`, `EINVAL` — the listener itself is broken, so retrying is an infinite hot spin) → warn and return, as today.
- Give up after N consecutive transient failures, and make *that* the observable outcome, so "this VM's egress is gone" cannot be silent.

## Scope

- [ ] Shared classifier + backoff helper, unit-tested against the error table
- [ ] Applied to all five loops
- [ ] Regression test: an injected `ECONNABORTED` does not end the loop; the next connection is still accepted
- [ ] Consecutive-exhaustion path is observable rather than silent

## Corrections to the original report

The two hypotheses this issue was filed on were checked and **do not hold**. Recording them so they are not re-investigated:

1. *"Unbounded retry when the peer ring stays full."* The wire protocol is length-prefixed with a hard cap — `read_frame_sync` refuses `n > MAX_FRAME_BYTES` (`DEFAULT_MAX_FRAME_BYTES` = 64 KiB, `crates/mvm-hostd/src/framing.rs:36`). An oversize frame is refused, not retried forever.

2. *"Deadlock on a large body through the substitution endpoint."* `can_skip_substitution_assembly` (`crates/mvm-hostd/src/bin/mvm-network-endpoint.rs:330`) does not mean what the report assumed — it is "this VM has `Raw` egress mode with no secrets, no terminator, no TLS intermediate and no FlowMux identity, so skip constructing the substitution service," not "skip buffering the body." Bodies are bounded, and `fail_closed_reason` (`network_endpoint_proxy.rs:452`) refuses oversize or `content-encoding`-bearing bodies outright when redaction is active. The shape is refuse, not wedge.

The prior-art note that motivated the original filing pointed at a real family of bugs; the shape we actually carry is the accept-loop one above.

auser:
Closes #2485.

Found by auditing the production egress path. Two premises in the original issue were wrong and are corrected in [a comment there](https://github.com/tinylabscom/mvm/issues/2485#issuecomment-5296588176); this PR fixes the defect that audit actually turned up.

## The defect

Every guest-facing accept loop treated any `accept(2)` error as fatal — log a `warn!`, `return`. Six sites: raw egress (UDS + vsock), the substitution endpoint (UDS + vsock, plus its accept-task panic arm), and the transparent terminator.

`accept(2)` reports two unrelated things through one channel. `EBADF` / `EINVAL` / `ENOTSOCK` mean the listener is dead. `EINTR`, `ECONNABORTED`, `ECONNRESET`, `EMFILE`, `ENFILE`, `ENOBUFS`, `ENOMEM` are about the pending connection or transient host pressure, and the listener is still fine.

There is no restart wrapper — `bin/mvm-network-endpoint.rs` awaits the primary loop directly — so the second class ended egress for that VM for the rest of its life. Two consequences, the second worse:

- **Collateral.** Fd exhaustion is usually host-wide, so the VM that lost egress was typically not the one that caused it, and it did not recover when the host did.
- **Silent.** The terminator is a spawned task the primary loop never joins. It could die while the VM kept running, leaving the redirect path gone with only a `warn!` on the endpoint's stderr as evidence.

Fail-closed throughout — no egress is safer than ungated egress — so **no claim is affected**. This is an availability and observability defect.

## The change

New `supervisor/accept_loop.rs`:

- `classify_accept_error(&io::Error, consecutive) -> Retry(Duration) | Fatal` — pure, so the policy is testable without a socket
- one delay policy: first four retries free (a lone `ECONNABORTED` is one dead connection attempt and the next accept should be immediate), then 1 ms doubling to a 1 s cap
- `record_listener_stopped` emits `host.listener_stopped` through the chain signer, so a VM losing egress is visible in the audit log and not only on stderr

All six sites route through it. Unrecognised errno values classify as `Fatal` — a listener failing in a way nobody characterised should surface loudly rather than spin forever on something that will not clear. The accept-task panic arm stays fatal (a panic is a bug in this process, not pressure that clears) but is now audited.

## Verification

- `cargo nextest run --workspace` — 11,609 passed, 0 failed
- `cargo +nightly fmt --all -- --check`, `cargo test --workspace --doc` — clean
- `check-no-spec-refs-in-comments`, `check-honesty`, `check-doc-claims`, `check-no-overclaim`, `check-uniform-vsock-egress`, `check-vsock-only-egress` — pass
- `just check-linux` — passes. Two of the six sites are `cfg(target_os = "linux")`; a macOS-only check would not have compiled them.

## Left open on the issue

The audit also confirmed a hard 16 MiB ceiling (`MAX_FRAME_BYTES`) on any request or response through the substitution path. It is enforced cleanly rather than truncating or hanging, but I could not find it stated in user-facing docs. Not touched here.

### DIFF
--- a/crates/mvm-hostd/src/supervisor/accept_loop.rs
+++ b/crates/mvm-hostd/src/supervisor/accept_loop.rs
@@ -1,0 +1,198 @@
+//! Accept-error policy shared by every guest-facing listener.
+//!
+//! `accept(2)` reports two very different things through one error channel.
+//! Some errors mean the listener is dead and the loop must stop (`EBADF`,
+//! `EINVAL`, `ENOTSOCK`). Others are about the *pending connection* or about
+//! transient host pressure — a peer that reset between SYN and accept, a signal,
+//! or exhausted file descriptors — and the listener is still perfectly good.
+//!
+//! Treating the second class as fatal ends egress for a VM that did nothing
+//! wrong, for the rest of its life: fd exhaustion is usually host-wide, so the
+//! process that caused it is typically not the one that dies, and the VM does
+//! not recover when the host does.
+
+use std::io;
+use std::time::Duration;
+
+use crate::supervisor::audit_recorder::{EventCategory, Recorder};
+
+/// Ceiling on the retry delay. Fd exhaustion clears on a human timescale, so a
+/// second between probes is cheap; the cap exists to stop the shift growing
+/// without bound, not because a longer wait would hurt.
+const MAX_RETRY_DELAY: Duration = Duration::from_secs(1);
+
+/// Consecutive transient errors tolerated before backing off at all. A lone
+/// `ECONNABORTED` is one dead connection attempt and the next accept should be
+/// immediate; only a *sustained* streak indicates pressure worth pausing for.
+const FREE_RETRIES: u32 = 4;
+
+/// What the caller should do about an accept error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AcceptAction {
+    /// The listener is still usable. Sleep this long, then accept again.
+    Retry(Duration),
+    /// The listener cannot recover. Stop the loop.
+    Fatal,
+}
+
+/// Classify an accept error, given how many transient errors have arrived
+/// back-to-back (0 for the first).
+///
+/// Unrecognised errno values are treated as fatal. A listener failing in a way
+/// we have not characterised should surface loudly rather than spin forever on
+/// an error nothing will clear.
+#[must_use]
+pub fn classify_accept_error(err: &io::Error, consecutive: u32) -> AcceptAction {
+    let transient = match err.kind() {
+        io::ErrorKind::Interrupted
+        | io::ErrorKind::ConnectionAborted
+        | io::ErrorKind::ConnectionReset
+        | io::ErrorKind::WouldBlock
+        | io::ErrorKind::OutOfMemory => true,
+        // EMFILE/ENFILE/ENOBUFS have no stable `ErrorKind`, so they only ever
+        // arrive as a raw errno.
+        _ => matches!(
+            err.raw_os_error(),
+            Some(libc::EMFILE) | Some(libc::ENFILE) | Some(libc::ENOBUFS) | Some(libc::ENOMEM)
+        ),
+    };
+    if transient {
+        AcceptAction::Retry(retry_delay(consecutive))
+    } else {
+        AcceptAction::Fatal
+    }
+}
+
+/// Delay before retry `consecutive` (0-based): free for the first few, then
+/// 1 ms doubling to the ceiling.
+fn retry_delay(consecutive: u32) -> Duration {
+    let Some(over) = consecutive.checked_sub(FREE_RETRIES) else {
+        return Duration::ZERO;
+    };
+    // Saturating shift: a listener wedged for a very long time must not
+    // overflow its way back to a zero delay.
+    let ms = 1u64.saturating_mul(1u64 << over.min(16));
+    Duration::from_millis(ms).min(MAX_RETRY_DELAY)
+}
+
+/// Record a guest-facing listener giving up.
+///
+/// A VM that loses egress goes quiet rather than failing loudly — the workload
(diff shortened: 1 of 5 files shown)


auser:
Independently audited #2485 from the issue side before finding this PR — the six sites and the classification table match what I arrived at, and catching the accept-task panic arm and auditing `host.listener_stopped` are both better than what I'd scoped. One gap worth a look before merge.

**`unrecognised → Fatal` reintroduces the bug for the terminator.**

The reasoning in the doc comment is sound for a listener that has failed in an uncharacterised way, but `accept(2)` on Linux has a documented behaviour that cuts against it:

> Linux `accept()` passes already-pending network errors on the new socket as an error code from `accept()`. This behavior differs from other BSD socket implementations. **For reliable operation the application should detect the network errors defined for the protocol after `accept()` and treat them like `EAGAIN` by retrying.**

So `accept` can return `EPROTO`, `EPERM` (firewall), `ENOSR`, `ETIMEDOUT`, `ENETDOWN`, `EHOSTUNREACH`, `EOPNOTSUPP` — none of which say anything about the listener, and all of which are `Fatal` under `classify_accept_error` today.

For the UDS and vsock loops this is mostly theoretical. For `serve_terminator` it isn't: that one takes a real `tokio::net::TcpListener`, so it is exactly the socket family the man page is describing. An `EPROTO` from one pending connection permanently ends the transparent egress redirect path — the same failure this PR exists to fix, arriving through a different errno.

Two ways to close it:

1. Add the pending-network-error set to the transient arm (`EPROTO`, `EPERM`, `ENOSR`, `ETIMEDOUT`, `ENETDOWN`, `ENETUNREACH`, `EHOSTUNREACH`, `EOPNOTSUPP`). Keeps unknown-is-fatal, just widens the known set.
2. Invert for the TCP listener only: an explicit fatal set (`EBADF`, `EINVAL`, `ENOTSOCK`) and retry everything else, with the streak cap as the backstop against a spin. The cap already bounds the "spins forever" risk that motivated fatal-by-default.

I'd take (1) — it keeps one policy for all six sites and the justification stays legible next to each errno.

Also worth a test either way: `unrecognised_errors_are_fatal_rather_than_spinning` currently pins `EDOM`, which `accept` cannot return. Pinning a real one (`EPROTO`) would have caught this.

Everything else read clean to me — the `FREE_RETRIES` / saturating-shift reasoning, the cap test, and auditing the terminator's exit (which the primary loop never joins) all look right.

PR #2498, opened 2026-08-14
fix(release): stop requiring mvm-bridge, a binary Plan 305 deleted

### LINKED ISSUE #2497: Release verification requires mvm-bridge, a binary Plan 305 deleted
Plan 305 deleted the `mvm-bridge` sidecar (WS1) and corrected the prose docs (WS3), but the packaging and install layer still names it. No `[[bin]]` declares `mvm-bridge` anywhere in the workspace.

## Impact — release-blocking

`nix/packaging/release/verify-release-assets.sh` lists `mvm-bridge` as a **required** asset for every target:

```
*apple-darwin) echo "mvmctl mvm-bridge mvm-hvf-supervisor mvm-libkrun-supervisor mvm-network-endpoint"
*)             echo "mvmctl mvm-bridge mvm-network-endpoint"
```

`.github/workflows/release.yml:929` runs that script, but its build step (lines 146-154) and staging loop (line 210) build and bundle only:

- macOS: `mvmctl`, `mvm-hvf-supervisor`, `mvm-libkrun-supervisor`, `mvm-network-endpoint`
- Linux: `mvmctl`, `mvm-network-endpoint`

So the next tagged release fails asset verification on every target.

## Why CI never caught it

`verify-release-assets.test.sh` has its own `bins_for()` carrying the same stale list. `build_valid_fixture` creates whatever that names, so the fixture manufactures a `mvm-bridge` file and the gate passes. Both lists were stale in lockstep — no amount of fixture testing could catch it.

## Other stale references

- `install.sh` — codesign/install loop and its comment
- `nix/packaging/homebrew/mvmctl.rb.tmpl` — formula installs it
- `public/src/content/docs/install/{macos,linux}.md` — `--bin mvm-bridge` build instructions; these also name `mvm-substitution-endpoint`, which is not the binary's name either (it is `mvm-network-endpoint`)
- `CLAUDE.md` — lists `mvm-bridge` as a per-VM supervisor bin, and `supervisor/` as containing a `gateway_bridge/` module that was deleted with the gateway stack
- `Cargo.toml` — two comments describing the deleted sidecar

`public/docs/investigations/binary-size-baseline.md` also names it, but that is a dated measurement record and is left as historical.

## Fix

Remove the stale references and add a drift gate to `verify-release-assets.test.sh` that derives the shipped set from `release.yml`'s staging loop and asserts every required bin is one the workflow actually bundles, so the two lists cannot drift again.

auser:
Closes #2497.

Plan 305 deleted the `mvm-bridge` sidecar and corrected the prose docs, but the packaging and install layer kept naming it.

## The break

`verify-release-assets.sh` listed `mvm-bridge` as a **required** asset for every target. No `[[bin]]` declares it, and `release.yml` builds and bundles only `mvmctl`, `mvm-network-endpoint`, and — on macOS — the two supervisors. `release.yml:929` runs that verifier, so the next tagged release would have failed asset verification on every target.

## Why the gate was blind

`verify-release-assets.test.sh` carried its own copy of the same stale list, and `build_valid_fixture` creates whatever that copy names — so the fixture manufactured a `mvm-bridge` file and the check went green. Both lists were stale in lockstep; no amount of fixture testing could catch it.

The new test #12 derives the shipped set from `release.yml`'s staging loop and asserts every bin the verifier requires is one the workflow actually bundles. Reinstating `mvm-bridge` in the verifier turns it red:

```
FAIL: verifier requires bin(s) release.yml never builds: mvm-bridge
```

## Also corrected

- `install.sh` install loop and comment; Homebrew formula.
- Both live install docs, which additionally named `mvm-substitution-endpoint` — not the binary's name either; it is `mvm-network-endpoint`.
- `CLAUDE.md`: listed `mvm-bridge` as a per-VM supervisor bin, and `supervisor/` as containing a `gateway_bridge/` module deleted with the gateway stack.
- Two `Cargo.toml` comment blocks describing the deleted sidecar, sitting above no dependency at all.

`public/docs/investigations/binary-size-baseline.md` still names it and is deliberately left alone — it records measurements taken at a point in time.

## Validation

`shellcheck` on all three scripts and `verify-release-assets.test.sh` at 12 passed / 0 failed — both lanes `ci.yml` runs for these paths. `cargo metadata` confirms the manifest still parses; no Rust changed.

### DIFF
--- a/.github/workflows/ci-full.yml
+++ b/.github/workflows/ci-full.yml
@@ -673,7 +673,7 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: macos-latest
     # `cargo build --all-targets` builds the Rust-native per-VM host bins
-    # (workspace `[[bin]]`s in mvm-vm-host) — no Swift build step.
+    # (workspace `[[bin]]`s in mvm-hostd) — no Swift build step.
     steps:
       - uses: actions/checkout@v6
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,7 +104,7 @@ Persistent builder state dirs live under `~/.mvm/cache/builder-vm/vms/`, disting
 
 **Top of graph (daemons, SDK, FFI):**
 
-- `mvm-hostd` -- host-side daemon roles, one crate with separate `[[bin]]`s (the process moat): the `supervisor` + `jailer` libs, the `broker`/`host_signer`/`audit_signer` subprocess bins, and the per-VM supervisor bins `mvm-libkrun-supervisor`/`mvm-hvf-supervisor`/`mvm-bridge`. Absorbs `mvm-supervisor`/`mvm-broker`/`mvm-host-signer`/`mvm-audit-signer`/`mvm-jailer-lite`/`mvm-vm-host`.
+- `mvm-hostd` -- host-side daemon roles, one crate with separate `[[bin]]`s (the process moat): the `supervisor` + `jailer` libs, the `broker`/`host_signer`/`audit_signer` subprocess bins, and the per-VM supervisor bins `mvm-libkrun-supervisor`/`mvm-hvf-supervisor`. Absorbs `mvm-supervisor`/`mvm-broker`/`mvm-host-signer`/`mvm-audit-signer`/`mvm-jailer-lite`/`mvm-vm-host`.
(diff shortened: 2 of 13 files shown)


auser:
Added a second instance of the same bug class, found by generalizing the check.

`-p mvm-vm-host` — a crate the Bar-A consolidation absorbed into `mvm-hostd` — was still being built by two scripts, both of which fail on that line:

```
error: package ID specification `mvm-vm-host` did not match any packages
```

- `scripts/check-hvf-oci-allow-host-smoke.sh:66` — reachable by hand via `Justfile:226`
- `scripts/measure-hvf-density.sh:199`

In both, the preceding line already said `-p mvm-hostd`; only the supervisor's line was missed when the crate moved. A comment in `ci-full.yml` named the dead crate too.

Method: enumerate the real binaries (42) and crates (18) from `cargo metadata`, then scan packaging, CI, scripts, and docs for names absent from both. That is how the original `mvm-bridge` break and this one were found; they share a root cause — a rename that updated the code and not its callers.

PR #2513, opened 2026-08-14
fix(hostd): make wall-clock bounds actually enforceable, on libkrun and HVF

### LINKED ISSUE #2507: Wall-clock grants are enforced on libkrun but not on HVF, the macOS 26+ default
## The gap

`mvm_hostd::supervisor::wall_clock` implements a real deadline killer, and `mvm-libkrun-supervisor.rs:275` arms it:

```rust
let _wall_clock = match mvm_hostd::supervisor::wall_clock::arm_for_supervisor(
    mvm_hostd::supervisor::wall_clock::SupervisorTimerInputs {
        plan_json: cfg.plan.as_ref(),
        audit_dir: cfg.audit_dir.as_deref(),
        signing_key_path: cfg.signing_key_path.as_deref(),
        vm_state_dir: std::path::Path::new(&cfg.vm_state_dir),
    },
) { ... }
```

`mvm-hvf-supervisor.rs` never calls it — `grep wall_clock` over that file returns nothing. So a workload admitted under a `max_wall_clock_secs` grant is stopped at its deadline on libkrun and runs unbounded on HVF.

That asymmetry matters because HVF is the **auto-detect default workload backend on macOS 26+ Apple Silicon**, so the default path is the unenforced one.

## Why it isn't a one-line mirror

`HvfSupervisorConfig` (`crates/mvm-vmm/src/host/hvf_supervisor.rs:94`) carries none of the inputs the timer needs — no `plan`, no `audit_dir`, no `signing_key_path`. Its fields are kernel/cmdline/memory/initramfs/disks/virtiofs/vsock/console_log/pid_file/workload_exit.

Wiring it therefore needs three steps:

1. Add the plan + audit-dir + signing-key fields to `HvfSupervisorConfig`.
2. Populate them where the HVF backend writes that config.
3. Arm the timer in `mvm-hvf-supervisor.rs`, mirroring the libkrun call site, including its fail-closed arm — libkrun exits `7` rather than boot a bounded workload it cannot audit.

Step 3's fail-closed behaviour is the part worth preserving deliberately: a bounded workload that cannot be audited should not boot.

## Documentation drift found alongside it

`CLAUDE.md:496` states, of Preview claim 18:

> Wall clock is claimed by nothing: it is authored, ceiling-checked, admitted and audited as a tier label, and no code in this tree stops a workload at its deadline.

That is no longer true — `supervisor/wall_clock.rs` stops a workload at its deadline on libkrun. The accurate statement is that enforcement exists and is wired on exactly one backend. ADR-001's Preview 18 limits note should be checked for the same wording, since the ledger is authoritative and the prose is not.

The drift is worth fixing independently of the code: a note saying "nothing enforces this" is a reason nobody looks, which is plausibly how the HVF path stayed unwired.

auser:
Closes #2507.

## The premise needed correcting first

#2507 reported wall clock as "enforced on libkrun but not HVF". Auditing before implementing, it was enforced on **neither** — wiring HVF to mirror libkrun would have mirrored an inert mechanism.

- `mvm-libkrun-supervisor.rs:275` does arm the timer, fail-closed. True.
- But `arm_for_supervisor` returns `Ok(None)` without a plan (`supervisor/wall_clock.rs:210`).
- The converged libkrun driver set the role fields `None` on purpose (`driver/libkrun.rs`): *"All inert so the supervisor takes its legacy run path and enforces no admission."*
- The only builder that populated them, `libkrun_legacy::build_supervisor_config`, had **zero callers** repo-wide.
- `mvmd` is not a hidden consumer: it references `SupervisorConfig`, `arm_for_supervisor` and `wall_clock` nowhere.

## The fix

Thread the plan plus its two audit paths through the converged runner→driver seam as `PlanBinding` on `VmmSpec`, populated in the single `VmStartConfig` → `VmmSpec` mapping and consumed by both drivers. `AuditSubstrate` already lives in `mvm-vmm` beside `VmmSpec`, so this adds no dependency, and `cpu_grant` is the precedent for a grant input on the spec.

**`PlanBinding` deliberately omits `tenant_id` and the gateway sockets.** The supervisor selects its admission route on `tenant_id` (`mvm-libkrun-supervisor.rs:223,259`), so carrying the full substrate would have silently moved workloads onto the admission path. `carrying_a_plan_does_not_move_the_supervisor_off_its_legacy_route` pins that. The audit entry takes its tenant from the plan itself, so nothing is lost.

## Why HVF matters most

It is the macOS 26+ auto-detect default, so the default path was the unenforced one. Its `timeout_secs` was never the plan's bound — it comes from the `MVM_HVF_TIMEOUT` dev hook and defaults to `0` ("run until stopped"), despite a doc claiming the backend set it from the VM's requested lifetime. Now documented as the unaudited backstop it is, distinct from the audited exit-124 kill.

## The part that would have made this inert

`ResourceControls::for_backend` answered `WallClockControl::None` for HVF, and `--prod` **refuses** a grant a tier reports it cannot enforce. Wiring the supervisor without this would have shipped a feature unreachable in production. AppleContainer shares the arm because it *is* this tier — it runs an `HvfRunner` over the same driver and supervisor, substituting only the kernel image — so it inherits enforcement rather than being an independent claim.

## Scope held deliberately

A restored child does not re-arm the timer. Inheriting the parent's plan would audit the child's kill under the parent's identity, and a wrong entry in the chain is worse than a missing one. Stated in the code, CLAUDE.md and ADR-001 limit 2 rather than left implicit.

## Docs

CLAUDE.md said wall clock was "claimed by nothing"; ADR-001 said *both* that (§Preview 18) and "enforced on libkrun only" (limit 2) in different sections. Neither described the shipped path. A note saying nothing enforces something is plausibly why the gap persisted, so both are corrected to say what is true per tier.

`build_supervisor_config` is deleted — dead code shaped exactly like the missing wiring.

## Verification

- `cargo nextest run --workspace` — **11,624 passed**, 26 skipped
- `cargo test --workspace --doc` — green
- `cargo +nightly fmt --all` — clean
- xtask gates: `check-honesty`, `check-no-overclaim`, `check-claim-catalog`, `check-dormant-controls`, `check-backend-resource-controls`, `check-doc-claims`, `check-adr-coverage`, `check-witness-citations` — all OK

New tests: the binding is produced for a plan-bearing launch and absent otherwise (including unparseable JSON); both drivers hand the supervisor the plan and the audit paths; the admission route is unchanged; the plan's bound is not conflated with the dev backstop.

`only_a_tier_with_a_live_supervisor_claims_a_supervisor_timer` failed against this change and was updated deliberately, not blindly — its principle ("only a tier with a supervisor process may claim a timer") is preserved and sharpened to the two conditions that actually matter: a process outliving the CLI **and** the plan to read a bound from.


auser:
## All five failures are one root cause

```
error[E0063]: missing field `plan_binding` in initializer of `mvm_runtime::driver::VmmSpec`
error: could not compile `mvm-conformance` (test "conformance") due to 1 previous error
```

The missed initializer is **`crates/mvm-conformance/tests/steps/verified_boot.rs:96`** — the only `VmmSpec { .. }` in that crate.

`Lint core`, `Lint`, `Test`, `Test workspace`, and `Test Linux and conformance` are all downstream of that single compile error, so one field fixes all five.

## Why it slipped

`mvm-conformance` is a dev-only crate that is not a dependency of any shipped crate, so it is invisible to a `-p`-filtered build and to `cargo build`. Adding a field to a struct only surfaces here under `--all-targets` across the workspace, or under `cargo nextest run --workspace`.

Same class as the standing note in `CLAUDE.md` about running the full workspace suite rather than a filtered one before pushing — `xtask`, `mvm-conformance`, and the `tests/` trees are exactly what a filtered run skips.

## On the premise correction

Worth saying plainly since I filed #2507: your audit is right and my issue was wrong. I verified that `mvm-libkrun-supervisor.rs:275` *calls* `arm_for_supervisor` and concluded libkrun enforced the bound. I did not check what the call does — `wall_clock.rs:210` returns `Ok(None)` the moment `plan_json` is absent, and the converged driver sets those role fields `None` deliberately. So the call site proved nothing about enforcement, and "enforced on libkrun but not HVF" should have read "enforced on neither".

Reframing it as making the bound enforceable on both, rather than mirroring libkrun onto HVF, is the correct call — mirroring would have copied an inert mechanism.

auser:
Update on the `Test workspace` failure, and a correction.

It failed twice at `xtask check-nextest-groups` with `cargo nextest list failed for filter '...'` and nothing else — the gate swallowed the cause chain, so the message named only the filter, which the reader already has. I pushed `b89387ee6` to print `{e:#}` instead of `{e}`, so a future failure names the compiler diagnostic or the binary that refused to enumerate.

**That change is not the fix.** This run passed `Test workspace` (34m48s) with the gate clean, so the earlier failures did not recur. I could not reproduce them: the gate passes on macOS, a `cargo zigbuild --target x86_64-unknown-linux-gnu --workspace --all-targets` with `RUSTFLAGS=-D warnings` compiles clean, and `cargo nextest list` with the same filters and `-D warnings` resolves all four overrides. `nextest list` compiles every workspace member — the preceding `cargo build --all-targets` step does not, since this workspace has no `default-members` and so builds only the root package — which makes it the heaviest step in that job and the most exposed to a loaded runner.

So: transient, cause unknown, and the branch is green. The `{e:#}` commit stays because the next occurrence should not be this hard to diagnose.

PR #2518, opened 2026-08-15
fix(egress): bound simultaneous raw TCP tunnels per workload

auser:
## The bug

`serve_raw_egress` spawns one task per accepted guest connection, and the raw-tunnel branch of `handle_raw_conn` never admitted against anything. A guest could hold **unlimited** simultaneous tunnels open.

`EgressRateGuard` was already threaded into `handle_raw_conn` — but it only ever reached the DNS and ICMP verbs. Its own module doc says so:

> Per-workload admission guard shared by the **metered egress verbs**. DNS and ICMP are both request/response verbs…

So the splice path took the argument and dropped it on the floor. Adjacent to `Preview` claim 18 (workload resource bounding), on a claim-10 path.

## The fix

Admit the tunnel **before resolving the target**, on both the async UDS path and the blocking vsock sibling, nacking with the existing `ConnectAck::Fail` when the pool is full. Taking the slot ahead of the gate means a connect flood is refused at the door rather than after the host has already done the DNS work for it.

### Why tunnels get their own semaphore

A request/response verb holds its slot for one round trip. A tunnel holds one for as long as the guest keeps the connection open. On a shared pool, a handful of long-lived tunnels would take every slot and leave the guest unable to resolve a name — a denial of service *inside the guard that exists to prevent one*. So: separate simultaneity pool, shared token bucket (a connect storm is host work whichever verb asks for it). `tunnels_and_requests_draw_on_separate_pools` pins that.

`EgressRateGuard::new` becomes a builder now the type carries three tunables, per the workspace's constructor guidance.

## On the tests

The first version of these tests was worthless and I want that on the record, because it passes review easily.

A refused tunnel's wire tell is `ConnectAck::Fail` — **the same byte** a policy refusal sends, and the same byte a failed upstream connect sends. My first test asserted on that byte, and it passed with the fix reverted. The gate had allowed the target and the connect had simply failed in the sandbox.

So the handler now takes its resolver as a parameter — the same `serve_dns` / `serve_dns_with_resolver` split already used next door — and the witness is behavioural: **a full pool never reaches the resolver**, with a companion test showing an available slot does, so the first isn't vacuous. Reverting the admission turns it red:

```
assertion `left == right` failed: a full tunnel pool must refuse before the host does any DNS work
  left: 1
 right: 0
```

## Known gap

The blocking vsock path carries the same admission and is compile-checked for Linux (`cargo zigbuild -p mvm-hostd --all-targets --target x86_64-unknown-linux-gnu`), but has **no dedicated witness of its own**. I did not add one I could not prove goes red from this host. Tracked separately.

## Verification

- `cargo nextest run --workspace` — 11637/11637 pass, 26 skipped
- `cargo zigbuild -p mvm-hostd --all-targets --target x86_64-unknown-linux-gnu` — clean
- Red-without-the-fix confirmed for `an_exhausted_tunnel_pool_refuses_before_resolving_the_target`

### DIFF
--- a/crates/mvm-hostd/src/supervisor/dns_handler.rs
+++ b/crates/mvm-hostd/src/supervisor/dns_handler.rs
@@ -413,13 +413,19 @@ mod tests {
 
     #[tokio::test]
     async fn rate_guard_denies_when_bucket_is_empty() {
-        let guard = EgressRateGuard::new(0, 1);
+        let guard = EgressRateGuard::builder()
+            .requests_per_sec(0)
+            .max_inflight(1)
+            .build();
         assert!(guard.admit().await.is_none());
     }
 
     #[tokio::test]
     async fn rate_guard_caps_concurrent_queries() {
-        let guard = EgressRateGuard::new(10, 1);
+        let guard = EgressRateGuard::builder()
+            .requests_per_sec(10)
+            .max_inflight(1)
+            .build();
         let first = guard.admit().await.expect("first query is admitted");
         assert!(guard.admit().await.is_none());
         drop(first);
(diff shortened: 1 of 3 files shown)


auser:
Known-gap follow-up filed as #2519.

Issue #2510, opened 2026-08-14
machine run --image on a KVM host: Firecracker dies at InstanceStart with zero guest serial, while the same kernel boots standalone


auser:
## Symptom

On the KVM box (`88.99.197.234`, Ubuntu 6.8.0, x86_64, `/dev/kvm`), a fresh release build of `main` at `a6599e1c`:

```
mvmctl machine run -d --name probe2299 --image alpine -- /bin/true
Error: backend error: Firecracker process for 'probe2299' exited before its
guest agent came up; see /root/.mvm/vms/probe2299/console.log
```

`console.log` contains **no guest serial output at all** — it stops after the `/logger` PUT, which is expected (that PUT is where Firecracker's own logging moves to `firecracker.log`), but no guest kernel line ever follows.

## The process dies at InstanceStart

`firecracker.log` from the failed run spans 205 ms and ends the instant the vCPUs resume:

```
22:28:06.578079917   first line
22:28:06.603747472   Put "/actions" {"action_type": "InstanceStart"}
22:28:06.783080071   fc_vcpu 1: Received a VcpuEvent::Resume ...
22:28:06.783202041   fc_vcpu 0: Received a VcpuEvent::Resume ...
22:28:06.783447801   last line — nothing after
```

The full API sequence succeeded first: `/boot-source`, `/machine-config`, `/drives/blk0..blk3`, `/vsock`, `/actions`. So this is not a config rejection — the VM is built and started, then the process is gone before the guest emits its first kernel line.

## The same artifacts boot fine standalone

Driving Firecracker directly with the kernel and initrd lifted verbatim from that run's `/boot-source` PUT:

- kernel `~/.mvm/cache/builder-vm/x86_64/kernels/workload/vmlinux.elf` (Linux 6.12.87)
- initrd `~/.mvm/cache/initramfs/0.18.0/x86_64/initramfs.cpio.gz`
- `firecracker --no-api --config-file`, 2 vCPU / 512 MiB, no drives

```
[ 0.286294] Run /init
[ 0.286128] Run /init
[ 0.282880] Run /init
```

Stable across three runs. **Adding a `/logger` stanza identical to the one mvmctl configures does not suppress guest serial** — 229 guest kernel lines still land on stdout and `Run /init` still appears at 0.289 s. So the missing console output in the failing run is not a capture artifact; the guest genuinely never ran.

## What differs between the two

The failing config adds, relative to the standalone one:

- four virtio-blk drives (`blk0`..`blk3`)
[...]

PR #2524, opened 2026-08-15
fix: the guest agent could not spawn itself — capability bounding set, wrong order

### LINKED ISSUE #2522: machine run --image is broken on main: the guest agent spawn fails EPERM on its own capability bounding-set drop
## Symptom

Every `mvmctl machine run --image <ref>` fails on `main`:

```
$ mvmctl machine run --image python:3.12 -- python -c "print(2 + 2)"
[mvm] Guest console at ~/.mvm/vms/brisk-badger-b7b1/console.log was empty before transient cleanup.
Error: guest agent did not become reachable within 30s
```

Not image-specific — `--image alpine -- /bin/echo hi` fails identically.

## Bisect

Same host (macOS 26.5.2 / arm64, HVF), same `~/.mvm`, same caches; only the binary differs:

| build | result |
| --- | --- |
| `69a2e581b` (= `a6599e1cb^`) | `hi`, exit 0 |
| `ae8f576d9` (main) | agent unreachable, exit 1 |

## Root cause

`a6599e1cb` (#2478) added a `PR_CAPBSET_DROP` loop to `drop_guest_agent_privilege_raw`
(`crates/mvm-agentd/src/guest_mount.rs`), placed after the identity change:

```
PR_SET_KEEPCAPS -> setgroups -> setgid -> setuid(901)
set_capabilities(RESTORE_AGENT_CAPABILITIES)   // effective now KILL|SYS_TIME - CAP_SETPCAP gone
raise_ambient_capabilities(...)
drop_capability_bounding_set_to(...)           // PR_CAPBSET_DROP needs CAP_SETPCAP -> EPERM
```

`PR_CAPBSET_DROP` requires `CAP_SETPCAP`, which the `capset` two lines earlier just
dropped. The error propagates out of the `pre_exec` hook, so `Command::spawn()` fails
and the agent never starts. Guest console, recovered after stopping the VM:

```
mvm-verity-init: switching to /init
mvm-guest-init: spawn guest-agent at /mvm/runtime/agent: Operation not permitted (os error 1)
```

Both agent paths funnel through that one function — `spawn_one_as`'s `pre_exec`
(the OCI route) and the PID-1 in-process drop in `bin/mvm-guest-agent/init.rs`
(the nix route) — so this is not OCI-specific, and being a guest-side capability
error it is not HVF-specific either.

#2478 diagnosed this exact failure mode for the *workload* spawn and fixed it there
(`drop_workload_capability_bounding_set` tolerates `EPERM` via
`bounding_drop_is_unenforceable`). The agent path never got the same treatment.

## Why nothing caught it

`oci-image-runner-smoke` and the live-VM lanes live in `ci-full.yml`, which is
`workflow_dispatch:`-only. No PR-gating lane boots a guest.

## Why it was invisible

On HVF the guest console accumulates in `Pl011.output` (`crates/mvm-vmm/src/vmm/device.rs`)
and reaches `console.log` only after the run loop returns
(`crates/mvm-hostd/src/bin/mvm-hvf-supervisor.rs`). While the VM is running — including
the entire 30s the host waits for the agent — the file is 0 bytes, so
`emit_guest_console_diagnostic` prints "Guest console ... was empty" and nothing else.
The one diagnostic built for a hung boot cannot fire on a hung boot. libkrun is
unaffected: it is handed the fd and streams into it.

Workaround for reading it today: boot with `-d --name <n>`, let it fail, then
`mvmctl machine stop <n> --yes` and read `~/.mvm/vms/<n>/console.log`.

## Fix

1. Narrow the bounding set in `drop_guest_agent_privilege_raw` *before* the identity
   change, while `CAP_SETPCAP` is still held, and tolerate only `EPERM` there (reusing
   `bounding_drop_is_unenforceable`) so an already-unprivileged caller still spawns.
2. Give the HVF `Pl011` a write-through sink so `console.log` is live during the run.

auser:
Fixes #2522.

`mvmctl machine run --image <anything>` has been broken on `main` since #2478
merged. Bisected on one host, one `~/.mvm`, one set of caches — only the binary
differed:

| build | result |
| --- | --- |
| `69a2e581b` (= `a6599e1cb^`) | `hi`, exit 0 |
| `ae8f576d9` | agent unreachable, exit 1 |

## The bug

`drop_guest_agent_privilege_raw` narrowed the capability bounding set *after*
`set_capabilities` had reduced the process to `CAP_KILL|CAP_SYS_TIME`.
`PR_CAPBSET_DROP` requires `CAP_SETPCAP`, which is not in that set, so it
returned `EPERM` every time. On the OCI route the function runs inside
`spawn_one_as`'s `pre_exec`, so the errno came back out of `Command::spawn()`
and the agent was never started:

```
mvm-verity-init: switching to /init
mvm-guest-init: spawn guest-agent at /mvm/runtime/agent: Operation not permitted (os error 1)
```

The nix route reaches the same function from PID 1, so it was broken
identically — this was never OCI- or backend-specific.

#2478 diagnosed this exact interaction for the *workload* spawn and fixed it
there by tolerating `EPERM`. The agent's own drop kept the broken ordering, and
the rule lived in two copies of the same `match`.

The narrowing now runs before the identity change, while the caller still holds
the capability the syscall needs — which enforces it rather than skipping it —
and both call sites share one helper.

## Second commit: why this took a session to find

On HVF the guest console accumulated in `Pl011.output` and reached `console.log`
only after the run loop returned. For the entire 30s the host waits for the
agent the file is 0 bytes, so `emit_guest_console_diagnostic` — the diagnostic
that exists specifically for a boot that never comes up — could only report
"Guest console … was empty". The transcript above had to be recovered by booting
detached, letting it fail, stopping the VM, and only then reading the file.

`Pl011` now takes an optional write-through sink, flushed per line and at a
512-byte cap. Additive: `output` still accumulates identically and the
supervisor's authoritative rewrite at exit is unchanged, so every consumer of
`KernelBootResult.console` sees what it saw before. Opened with the existing
`open_console_capture`, so it stays write-only and the sealed guest keeps its
"no host input fd" property. libkrun was never affected — it is handed the fd.

## Verification

Gates: `just lint` (fmt + clippy + clippy-bdd + xtask policy gates), `cargo
nextest run --workspace` (11690 passed), doctests, `just check-linux`, and a
`zigbuild --all-targets` for the touched crates — `guest_mount.rs` is entirely
`#[cfg(target_os = "linux")]`, so a macOS-only check compiles none of it.

Live on macOS 26 / HVF, against the pre-#2478 binary as the known-good baseline:

- `machine run --image alpine -- /bin/echo hi` → `hi`
- `machine run --image python:3.12 -- python -c "print(2 + 2)"` → `4`
- `machine run -d --name … --image alpine -- /bin/echo hi` → `hi`
- console.log is 13698 bytes **while the VM is still running** (0 before)

## Noted, not fixed here

- No PR-gating lane boots a guest: `oci-image-runner-smoke` and the live-VM
  lanes live in `ci-full.yml`, which is `workflow_dispatch:`-only. A change that
  makes every guest fail to boot passes every required check.
- A long-running detached exec (`machine run -d … -- /bin/sleep 120`) fails with
  `failed to fill whole buffer`. Reproduces identically on the pre-#2478 binary,
  so it is pre-existing and unrelated.

### DIFF
--- a/crates/mvm-agentd/src/guest_mount.rs
+++ b/crates/mvm-agentd/src/guest_mount.rs
@@ -31,6 +31,10 @@ pub const CAP_KILL: u32 = 5;
 pub const CAP_NET_BIND_SERVICE: u32 = 10;
 /// Linux capability used by the guest agent to correct a restored wall clock.
 pub const CAP_SYS_TIME: u32 = 25;
+/// Linux capability `PR_CAPBSET_DROP` itself requires. Never retained — which
+/// is exactly why the bounding set has to be narrowed before the capability
+/// sets are, and not after.
+pub const CAP_SETPCAP: u32 = 8;
 /// Capabilities explicitly retained by the guest agent after boot setup.
 pub const RESTORE_AGENT_CAPABILITIES: u32 = (1u32 << CAP_KILL) | (1u32 << CAP_SYS_TIME);
 
@@ -540,11 +544,19 @@ pub fn drop_privilege_raw(uid: u32, gid: u32) -> std::io::Result<()> {
 
 /// Drop to the guest-agent identity while retaining only the fixed capabilities
 /// required for authenticated restore handling.
+///
+/// The bounding set is narrowed first, before the identity change, because
+/// `PR_CAPBSET_DROP` needs `CAP_SETPCAP` and [`set_capabilities`] below removes
+/// it: narrowing afterwards fails `EPERM`, and on the spawn path that errno
+/// leaves the `pre_exec` hook and kills the agent before it ever runs. The
+/// bounding set is inherited across fork and exec and can only shrink, so
+/// applying it here binds the agent and everything under it just as firmly.
 #[cfg(target_os = "linux")]
 pub fn drop_guest_agent_privilege_raw(uid: u32, gid: u32) -> std::io::Result<()> {
     if unsafe { libc::prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) } != 0 {
         return Err(std::io::Error::last_os_error());
     }
+    narrow_bounding_set_where_enforceable(RESTORE_AGENT_CAPABILITIES)?;
     if unsafe { libc::setgroups(0, std::ptr::null::<libc::gid_t>()) } != 0 {
         return Err(std::io::Error::last_os_error());
     }
@@ -556,7 +568,6 @@ pub fn drop_guest_agent_privilege_raw(uid: u32, gid: u32) -> std::io::Result<()>
     }
     set_capabilities(RESTORE_AGENT_CAPABILITIES)?;
     raise_ambient_capabilities(RESTORE_AGENT_CAPABILITIES)?;
-    drop_capability_bounding_set_to(RESTORE_AGENT_CAPABILITIES)?;
     set_no_new_privileges()?;
     if unsafe { libc::getuid() } == 0 {
         return Err(std::io::Error::from_raw_os_error(libc::EPERM));
@@ -690,7 +701,19 @@ pub fn harden_init_process() -> std::io::Result<()> {
 #[cfg(target_os = "linux")]
 pub fn drop_workload_capability_bounding_set() -> std::io::Result<()> {
     set_no_new_privileges()?;
-    match drop_capability_bounding_set_to(0) {
+    narrow_bounding_set_where_enforceable(0)
+}
+
+/// Narrow the bounding set to `keep`, skipping the drop when this caller could
+/// never have performed it.
+///
+/// Shared by the agent identity drop and the workload spawn, which need the
+/// same "enforce where possible, fail closed on a real error" rule with
+/// different masks. See [`bounding_drop_is_unenforceable`] for why `EPERM`
+/// alone is a skip.
+#[cfg(target_os = "linux")]
+fn narrow_bounding_set_where_enforceable(keep: u32) -> std::io::Result<()> {
+    match drop_capability_bounding_set_to(keep) {
         Err(err) if bounding_drop_is_unenforceable(&err) => Ok(()),
         result => result,
     }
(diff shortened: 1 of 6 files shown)


auser:
Backend coverage, since the fix is guest-side and the claim "affects every
backend" should be evidence rather than assertion:

| backend | broken main | this PR | witnessed |
| --- | --- | --- | --- |
| HVF (macOS 26) | agent unreachable, 30s | `hi` / `4` | yes |
| libkrun (macOS) | agent unreachable, 30s | `hi` | yes |
| Firecracker (Linux/KVM) | — | — | blocked by #2510 |

libkrun printed the failing line straight to the terminal, because it hands the
console fd to the VMM. HVF buffered it until exit, which is the second commit
here.

Firecracker could not be witnessed either way: that host fails earlier, before
the guest emits anything, with the Firecracker process exiting at vCPU resume.
I built this branch on the KVM box and it fails identically, so **this PR does
not fix #2510** and does not claim to. Evidence and a new lead (the Firecracker
boot_args carry none of the `mvm.roothash` / `mvm.data` / `mvm.hash` /
`mvm.runtime_*` parameters the HVF path assembles, though all four drives are
attached) are posted there.

The capability bug is in the guest's own privilege drop, which Firecracker
reaches through exactly the same code, so it is affected by construction — it
just cannot be observed on that host until #2510 is fixed.

auser:
Cross-reference from #2531 (#2510): **your bug is now observable on Firecracker.**

The delivery note records Firecracker as "not witnessed either way" because that host "fails earlier, at vCPU resume". That earlier failure is #2510, and it turned out to be ours rather than Firecracker's: `sudo setsid nohup firecracker & echo $! > fc.pid` captures a pid whose `comm` is `sudo` on any host with `Defaults use_pty` (sudo ≥1.9.14, i.e. Ubuntu 24.04), so the `/proc/<pid>/comm == firecracker` liveness probe read every boot as an immediate exit and `FirecrackerGuard` killed a healthy VMM ~7ms in.

With #2531 alone on the KVM box, the VMM survives and the failure moves to exactly your defect:

```
Caused by:
    0: activate workload after boot
    1: guest activation failed: syscall failed: guest-agent privilege drop
```

So the two are sequential, distinct, and your "affected by construction, just not observable there" reading is confirmed empirically. Running both branches together now for the end-to-end Firecracker witness — I'll post the result here so the note's backend table can gain its third row.

auser:
**Third row for the backend table — Firecracker is now witnessed.**

With #2531 (the #2510 fix) merged into this branch on the KVM box (Ubuntu 24.04, Firecracker v1.14.1, release build):

```
$ ./target/release/mvmctl machine run --name probeboth --image alpine -- /bin/echo hi
hi
```

And with #2531 **alone** — i.e. this branch's fix absent — the same host fails at exactly your defect:

```
1: guest activation failed: syscall failed: guest-agent privilege drop
```

So the note's "affected by construction; it just cannot be observed there until #2510 is resolved" is now confirmed both ways. If you want the table completed before merge:

| backend | broken `main` | fixed |
| --- | --- | --- |
| Firecracker (Linux/KVM) | agent privilege drop fails, once #2510 is out of the way | `alpine` → `hi` |

Entirely optional — the fix stands on the two witnesses you already have.

PR #2529, opened 2026-08-15
feat(xtask): name new plans by slug, not by number

### LINKED ISSUE #2525: xtask: gate against new duplicate plan numbers in specs/plans/
## Problem

`specs/plans/` numbers are meant to identify a plan, but nothing enforces it. Main currently carries **18 duplicate numbers across 109 plan files**:

```
254 255 266 269 284 285 295 297 298 304 308 315 316 322 325 326 327 329
```

e.g. `254-admitted-egress-connect-completion.md` / `254-control-data-plane-hardening.md`, and `304-hvf-save-restore-bench.md` / `304-hvf-save-restore.md`. `specs/REFACTOR-STATUS.md` likewise has two entries titled "Plan 322".

This is not cosmetic in practice: cross-references say "Plan NNN", and with duplicates that no longer resolves to one document.

## Why it keeps happening

The natural way to pick a number — list `specs/plans/` in your worktree and take the next free one — **cannot see a number claimed by an unmerged branch**. It happened twice on #2501 alone: 332 and 333 were both free when that branch was cut, and both were taken by plans that merged while it sat in the merge queue.

## Proposed gate

`xtask check-plan-numbers` that fails only on a **newly-added** duplicate:

- Resolve the plan number of every file under `specs/plans/`.
- Grandfather the 18 numbers above in an explicit allow-list, with a comment saying they predate the gate and are not a licence to add more.
- Fail when a number outside that list appears more than once.

A naive global-uniqueness check is not viable — it fails on main today.

## Notes

- The allow-list should shrink, never grow. Worth a comment saying so.
- This does not solve the pick-a-number problem, only detection. A `just next-plan-number` helper that scans main plus open PRs would complement it.

Found while landing #2501; that PR renumbers its own two plans to 336/337 rather than adding to the pile.

auser:
Closes #2525.

## The root cause, not the symptom

The issue asked for a gate against duplicate plan numbers. Building it surfaced that the duplicates are a symptom: **a plan number is a sequential id allocated at authoring time**, and this repo adds ~3 plans a day across concurrent branches. Two authors routinely pick the same next-free number, and neither can see the other — a branch's claim is invisible until its PR opens. 18 numbers ended up shared by two plans; `REFACTOR-STATUS.md` carries two entries titled "Plan 322".

## Why a pick-time scan isn't enough

The obvious fix — scan open PRs before picking — was measured against the two collisions that prompted this:

| collision | rival PR opened | would a scan catch it? |
| --- | --- | --- |
| 333 | 1h30m **before** the pick | yes |
| 332 | 2h **after** the pick | no |

Any pick-time check closes half the window. The other half is a branch holding an unpublished claim, which nothing can see.

## What this does instead

New plans are named by slug — `2026-08-15-<slug>.md`. The date prefix keeps the chronological sort the numbers were really providing. A slug chosen for one piece of work doesn't collide with a slug chosen for another, so the failure mode is **removed**, not narrowed.

The 108 already-numbered plans keep their names — renaming them would invalidate 689 `Plan NNN` references across the tree for no benefit. They're frozen in `xtask/src/legacy_numbered_plans.txt`, which may shrink as plans retire and never grow; a stale entry is reported so it can be dropped.

This **subsumes** the duplicate check rather than sitting beside it: if no new numbered plan can appear, no new collision can either.

## Verification

- `cargo run -p xtask -- check-plan-names` → `clean (109 plans, 108 legacy-numbered, 1 slug-named)`
- **Proved red**: adding `999-probe-new-numbered.md` fails, naming the file
- **Proved permissive**: adding `2026-08-15-probe-slug-plan.md` passes
- Edge cases tested: a malformed date (`2026-8-15-x.md`) is treated as a number, not smuggled through as a slug; digits without a separator aren't a plan number; the pre-existing `recovery-open-work-2026-08-12.md` slug plan still passes
- Two tests assert the frozen list describes the tree it ships with and holds only number-named entries, so the baseline can't rot
- `cargo nextest run -p xtask` → 588 passed; fmt, clippy, `check-workflow-paths`, `check-doc-claims` clean
- Wired into `ci.yml` beside the other spec-hygiene gates

## Documentation

The rule is in `CLAUDE.md` (new §"Naming a new plan", with the reasoning) and referenced from `AGENTS.md` item 6 — where a plan author actually looks, not only in the gate that rejects them.

## Note on history

The first commit on this branch is the duplicate-number ratchet originally asked for; the second replaces it with this. Kept as two commits so the reasoning for the change of approach is legible.


auser:
Pushed `59d83b74a` — `check-no-spec-refs-in-comments` was rejecting the new gate's own module docs, which used two literal plan citations as examples. Reworded to keep the reasoning without the numbers; both `check-no-spec-refs-in-comments` and `check-plan-numbers` verified clean locally.

auser:
**Blocked on #2501 — do not merge this first.**

#2501 adds two number-named plans (`336-sdk-parity-and-bdd-coverage.md`, `337-sdk-surface-generated-from-rust.md`). They are not in this PR's frozen legacy list, so the two conflict in **either** order:

- this PR first → #2501 fails `check-plan-names`
- #2501 first → this PR fails `the_legacy_list_matches_this_tree`

Resolution, decided rather than grandfathered: those two plans are the newest in the repo and were numbered hours before we concluded that numbering is what causes the collisions. Adding them to a list documented as "may shrink, never grow" would undermine it on day one. So they get renamed to slugs instead.

Merge order:

1. #2501 merges (lands the two plans numbered)
2. a small follow-up renames both to `2026-08-14-<slug>.md` and moves their cross-references
3. this PR merges — its frozen list is then already correct, unchanged

The rename is written and verified locally; it is held back only because #2501 is in the merge queue and its branch is locked.

PR #2531, opened 2026-08-15
fix(fc): take the pid marker from the process that becomes Firecracker

### LINKED ISSUE #2510: machine run --image on a KVM host: Firecracker dies at InstanceStart with zero guest serial, while the same kernel boots standalone

auser:
## What was wrong

The Firecracker launch line records the VMM's pid with `$!`:

```sh
sudo setsid nohup firecracker --api-sock ... --enable-pci &
echo $! > fc.pid
```

`$!` is Firecracker only if every wrapper in that chain *execs*. `sudo` with
`use_pty` — the default since sudo 1.9.14, which is what Ubuntu 24.04 ships —
forks a monitor and runs the command as its **child** instead. Measured on the
KVM box (sudo 1.9.15p5, `Defaults use_pty` in `/etc/sudoers`):

```
captured pid=2715264   comm=sudo
2715264  1        sudo
2715266  2715264  sleep      <- the real command
```

Liveness is `/proc/<pid>/comm == "firecracker"` (`fc/host.rs:280`), so it
answered *no* on the first poll after `InstanceStart` and the driver concluded
the VMM had died (`fc/driver/fc.rs:930`).

## Why it looked like a VMM crash

Firecracker was healthy. From the failing run on that box:

```
fc boot: InstanceStart  ms=867
WARN FirecrackerGuard: killing orphaned Firecracker process dir=/root/.mvm/vms/probe2510
Error: Firecracker process for 'probe2510' exited before its guest agent came up
```

`firecracker.log` for the same run ends with `event_end: boot microvm` and both
vCPUs resuming normally — no error, no panic. `dmesg` is clean: no seccomp kill,
no OOM. **We killed it ourselves**, ~7 ms into the guest's boot, which is why
the reported symptom was "the process exits with no guest serial at all":
nothing had been printed yet. `fc.pid` was gone too, removed by the guard on
the way out.

This is why it reproduced on two unrelated hosts (the KVM box and GitHub's
`ubuntu-24.04` runners, both Ubuntu 24.04) while the same kernel and initrd
booted fine when Firecracker was driven directly.

## The fix

The launched process writes its own pid and then execs:

```sh
sudo setsid nohup sh -c 'echo $$ > "$0"; exec firecracker --api-sock "$1" --enable-pci' <pidfile> <socket> &
```

`$$` is correct whichever of `sudo`/`setsid` forks, because the `exec` makes
that same pid Firecracker. The `sh -c` leaves no surviving process for the same
reason, so the "no intermediate shell" property the previous line was written
for still holds — the test now asserts `exec firecracker` rather than the
absence of `bash -c`.

Ordering is safe: the marker is written before the exec, so it is in place well
before the API socket appears and the caller starts probing.

## Tests

`the_pid_marker_is_written_by_the_launched_process_not_the_caller` — asserts the
line carries no `echo $!`, writes `$$` to the marker path, and does so before
the exec. **Proven red** against the pre-fix launch line:

```
$! does not survive a forking sudo/setsid:
        sudo setsid nohup firecracker --api-sock '/tmp/vm/fc.socket' --enable-pci \
        echo $! > '/tmp/vm'/fc.pid
test result: FAILED. 0 passed; 1 failed
```

Full `mvm-backends` `fc::` suite (81 tests) and `cargo clippy --workspace
--all-targets -- -D warnings` are clean.

## Knock-on

This is also what makes the `Boot latency ceiling` lane on #2505 red — that job
fails with the identical "exited before its guest agent came up" on an
`ubuntu-24.04` runner.

Closes #2510.

### DIFF
--- a/crates/mvm-backends/src/fc/daemon.rs
+++ b/crates/mvm-backends/src/fc/daemon.rs
@@ -123,10 +123,18 @@ fn start_vm_firecracker_inner(
 ///
 /// Privilege justification: Firecracker is launched through `sudo` because, on
 /// the reference builder-VM host, the invoking user does not have direct access
-/// to `/dev/kvm` and the TAP/vsock network plumbing it configures. The launch
-/// no longer uses an intermediate `bash -c`: `sudo` execs `setsid` directly,
-/// which then execs Firecracker, so the only extra process is the one required
-/// for elevation.
+/// to `/dev/kvm` and the TAP/vsock network plumbing it configures.
+///
+/// The pid marker is written by the launched process itself (`echo $$`) and not
+/// by the caller (`echo $!`), because `$!` does not reliably name Firecracker.
+/// `sudo` with `use_pty` — the default since sudo 1.9.14, which is what Ubuntu
+/// 24.04 ships — forks a monitor and runs the command as its child rather than
+/// exec'ing it, so `$!` names a process whose `comm` is `sudo`. Liveness is
+/// probed by comparing `/proc/<pid>/comm` against `firecracker`, so on such a
+/// host every boot read as "the VMM exited" on the first poll while the guest
+/// was in fact booting normally. Writing `$$` from inside is correct whichever
+/// of `sudo`/`setsid` forks, since the `exec` makes that same pid Firecracker.
+/// The `sh -c` costs no surviving process for the same reason.
 fn firecracker_launch_script(
     abs_dir: &str,
     abs_socket: &str,
@@ -145,17 +153,18 @@ fn firecracker_launch_script(
     };
     let q_dir = shell_quote(abs_dir);
     let q_socket = shell_quote(abs_socket);
+    let q_pid = shell_quote(&format!("{abs_dir}/fc.pid"));
     format!(
         r#"
         mkdir -p {q_dir}
         sudo rm -f {q_socket}
         {vsock_cleanup}
         touch {q_dir}/console.log {q_dir}/firecracker.log
-        {cpu_scope}sudo setsid nohup firecracker --api-sock {q_socket} --enable-pci \
+        {cpu_scope}sudo setsid nohup sh -c 'echo $$ > "$0"; exec firecracker --api-sock "$1" --enable-pci' {q_pid} {q_socket} \
             </dev/null >{q_dir}/console.log 2>{q_dir}/firecracker.log &
-        echo $! > {q_dir}/fc.pid
         "#,
         q_dir = q_dir,
+        q_pid = q_pid,
         q_socket = q_socket,
         vsock_cleanup = vsock_cleanup,
         cpu_scope = cpu_scope,
@@ -390,7 +399,7 @@ mod tests {
         );
         let scope_at = script.find("systemd-run").expect("prefix present");
         let launch_at = script
-            .find("sudo setsid nohup firecracker")
+            .find("sudo setsid nohup sh -c")
             .expect("launch present");
         assert!(scope_at < launch_at, "{script}");
     }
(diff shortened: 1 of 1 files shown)


auser:
Verified live on the KVM box (Ubuntu 24.04, sudo 1.9.15p5 `use_pty`).

**Before this branch** — a healthy Firecracker killed 7ms into the guest's boot:
```
fc boot: InstanceStart  ms=867
WARN FirecrackerGuard: killing orphaned Firecracker process dir=/root/.mvm/vms/probe2510
Error: Firecracker process for 'probe2510' exited before its guest agent came up
```
`firecracker.log` for that run ends with `event_end: boot microvm` and both vCPUs resuming normally; `dmesg` clean (no seccomp kill, no OOM).

**With this branch** — the VMM survives and the guest boots. The failure moves to a strictly later stage:
```
Caused by:
    0: activate workload after boot
    1: guest activation failed: syscall failed: guest-agent privilege drop
```
That is #2522, fixed by #2524, which this branch does not carry.

Note for #2524: this is the first time #2522 is **observable on Firecracker**. Its delivery note records Firecracker as "not witnessed either way" precisely because #2510 masked it — the VMM was being killed before the agent ever ran. Combined verification of both branches is running now.

auser:
**End-to-end witness: Firecracker on Linux/KVM now boots and runs.**

`probe-2510` (this branch) merged with `probe-2522` (#2524) on the KVM box, release build, Firecracker v1.14.1 / Ubuntu 24.04 / 6.8.0:

```
$ ./target/release/mvmctl machine run --name probeboth --image alpine -- /bin/echo hi
hi
$ echo $?
0
```

The ladder on that host, one variable at a time:

| build | result |
| --- | --- |
| `main` | `Firecracker process ... exited before its guest agent came up`, zero guest serial |
| `main` + this branch | VM boots; fails later at `guest activation failed: syscall failed: guest-agent privilege drop` (= #2522) |
| `main` + this branch + #2524 | `hi`, exit 0 |

Both defects were real, sequential, and independent. This one was masking the other.

PR #2532, opened 2026-08-15
test(sec): kill two substitution mutants; delete an unreachable wasm guard

auser:
Three of the twelve mutants that survived Security run `31839663209`. Landing what is solid rather than holding it behind the eight I have not done.

**The lane is still red.** This does not fix it.

## Killed, verified red-first

**`PlaceholderMap::is_empty`** had no test at all, so both the constant-`true` and constant-`false` mutants survived. It is session bookkeeping — but bookkeeping the keyholder reads to decide whether a session has any substitution to do, and constant-`true` makes a populated session look empty. Both mutants confirmed to fail the new test before it was kept.

## Deleted rather than witnessed

**`resolve_manifest_arg`'s `if !module_path.is_file()` bail is unreachable.** `Manifest::read_file` already refuses a manifest whose `wasm` field does not name an existing file, so by that line the module is known to exist.

That unreachability is *why* mutating the guard away changed no test — there was no reachable behaviour to change. Deleting the code removes the mutant instead of papering over it with a test that cannot exist. I found this by writing the obvious test, watching it fail for the wrong reason, and probing `read_file` directly.

## What this does NOT do

The added test — a relative wasm module resolves against the manifest's directory — is a genuine behavioural test, but it **does not kill the `delete !` mutant on `is_absolute`**. I verified that by applying the mutation and watching the test still pass. That survivor stands.

## Remaining survivors (9)

| Crate | Site |
|---|---|
| mvm-cli | `resolve.rs` `is_absolute` — `delete !` |
| mvm-hostd | `admission_budget.rs:104` `committed` → `Default::default()` |
| mvm-hostd | `keyholder/substitution.rs:78` `as_map` → empty map |
| mvm-hostd | `plan_admission.rs:363` `admit_within_host_budget` — `delete !` |
| mvm-hostd | `plan_admission.rs:627` `*` → `+` (×2) |
| mvm-hostd | `plan_admission.rs:1166/1173` delete `plan_id` field (×2) |
| mvm-hostd | `audit_file.rs:91` `RotationPolicy::never` → `Default::default()` |

Several are claim-bearing: the budget ones are preview claim 18, `as_map` is on the claim-13 substitution path, and the `plan_id` field deletions are the claim-8 binding.

**Separately, the mvm-hostd shard did not finish** — it ended `Error: interrupted` after 73m and 55m shards. That is an infrastructure problem independent of the survivors, and it means that shard's list may be incomplete.

## Verification

`cargo nextest run -p mvm-contract -p mvm-cli` — 2,677 passed. Workspace all-target Clippy and fmt clean via the pre-commit gate.

### DIFF
--- a/crates/mvm-cli/src/commands/shared/resolve.rs
+++ b/crates/mvm-cli/src/commands/shared/resolve.rs
@@ -131,12 +131,10 @@ pub fn resolve_manifest_arg(arg: &str) -> Result<ManifestArgRef> {
                 });
             }
         }
-        if !module_path.is_file() {
-            anyhow::bail!(
-                "Manifest wasm module '{}' does not exist",
-                module_path.display()
-            );
-        }
+        // No existence check here. `Manifest::read_file` already refuses a
+        // manifest whose `wasm` field does not point to an existing file, so
+        // by this line the module is known to exist — a second check was
+        // unreachable, which is why mutating it away changed no test.
         return Ok(ManifestArgRef::WasmModule {
             manifest_path: canonical,
             module_path: module_path.to_path_buf(),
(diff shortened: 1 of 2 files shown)


auser:
#2540 supersedes this, and it agrees with the finding here — one of the two wasm branches in `resolve_manifest_arg` is dead code and should be deleted rather than witnessed. It just deletes the other one.

The reachable half is `!module_path.is_file()`: inverting it refuses every wasm manifest. What makes it removable is not unreachability but duplication — `Manifest::validate` runs the same `path.is_file()` on the same resolved path a few lines earlier.

The dead half is `!module_path.is_absolute()`, which is why the test added here does not kill its mutant, exactly as this PR's body says. `read_file` resolves a relative `wasm` against the manifest's own directory before returning and the path handed to it is canonicalized, so the module path is always absolute; with the `!` deleted, `base.join(absolute)` returns it unchanged and the function produces the same value.

So with this branch applied the `mvm-cli` shard still fails, on the survivor it keeps. #2540 removes both, asserts the behaviour directly, and closes the other six survivors across mvm-core, mvm-vmm, mvm-hostd and mvm-contract. Verified with `cargo mutants` on the whole file: 25 mutants, 23 caught, 2 unviable, 0 missed.

Suggest closing this in favour of #2540.

PR #2541, opened 2026-08-15
chore(deps): re-resolve Cargo.lock after rand 0.8 -> 0.10 unification

auser:
Follow-up to #2523. Running cargo against the merged tree prunes the now-unused `rand 0.8.6`, `rand_chacha 0.3.1`, and stale getrandom edge from `rand_core 0.6.4`. No source changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3019,7 +3019,7 @@ dependencies = [
  "mvm-sdk",
  "mvm-vmm",
  "nix 0.29.0",
- "rand 0.8.6",
+ "rand 0.10.2",
  "serde",
  "serde_json",
  "sha2 0.11.0",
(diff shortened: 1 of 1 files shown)


auser:
Closing as superseded — this is now a no-op, and it is holding the queue head ahead of a fix that everything else is waiting on.

The same re-resolve landed in #2536. Merging this branch into current `main` produces **zero net change**:

```
$ git merge --no-commit --no-ff <2541-head>     # into origin/main
merge-exit=0
$ git diff --cached --stat origin/main
(no output)
```

`main`'s lock has also moved past this branch's copy since (`mvm-observability` from #2515), so this version is strictly older.

Why it matters right now: a `Cargo.lock`-only diff touches no kernel-scoped path, so under the current CI bug it can never report `Build kernels (aarch64)`/`(x86_64)` and can never satisfy branch protection. Sitting at queue position 1 it blocked every other entry — #2548 fixes that bug but could not get past this.

Branch `chore/relock-rand-drift-2` is untouched; `gh pr reopen 2541` if you disagree.

Issue #2539, opened 2026-08-15
Published images from v0.17.0 ship a /init the kernel cannot exec (ENOEXEC)


auser:
Every `default-microvm` image published from `v0.17.0` onward ships a `/init`
that the kernel cannot exec. Its first bytes are:

```
20 23 21 2f 62 69 6e 2f 73 68 0a    " #!/bin/sh\n"
```

`#!` is only a shebang at byte 0, so the boot ends in:

```
Kernel panic - not syncing: Requested init /init failed (error -8)
```

`-8` is `ENOEXEC`. Both architectures are affected. The checksum manifest
verifies, so the artifact is a faithful copy of what the generator produced.

The generator bug is fixed by #2538: `nix/lib/mk-guest.nix` had one line
indented three spaces where the rest of the `/init` block sits at four, and
Nix derives an indented string's baseline from its least-indented line, so it
stripped three columns and shifted every line — the shebang included — one
space right. `v0.16.1`, cut before the line drifted, has `#!/bin/sh` at byte 0.

#2538 corrects the generator. It cannot correct an already-published artifact,
so two things remain:

- [ ] Cut a release from a tree containing #2538 and confirm the new
      `default-microvm-rootfs-<arch>.ext4` opens `/init` with `#!/bin/sh` at
      byte 0
- [ ] Bump `IMAGE_TAG` in the `boot-latency` job of `.github/workflows/ci.yml`
      to that release and restore its `needs: [scope]` /
      `if: needs.scope.outputs.code == 'true'` — #2538 made it manual-dispatch
      only because it cannot pass against a published image that will not boot

Worth deciding separately whether the affected releases should be marked, since
anyone who fetched a prebuilt image in that window has one that cannot boot.

auser:
## The repin patch, staged and verified

PR #2542 held this change but could not be green: the lane boots a *published* image, so pinning it to a tag that has not been cut yet fails at step 10, `Fetch and verify the pinned bootable image`, with a 404. No repo change can fix that — the tag is the prerequisite. Closed it rather than leave a permanently-red PR; branch `ci/2539-repin-boot-image` is kept, so `gh pr create` on it (or `gh pr reopen 2542`) brings it straight back.

Apply after the release is cut. Verified `actionlint` clean and `xtask check-workflow-paths` clean at the time of staging:

```diff
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@
-  #
[...]

PR #2542, opened 2026-08-15
ci: put the boot-latency lane back on every PR

auser:
**Draft on purpose — do not merge until `v0.18.0` is published.** The lane boots a
*published release*. Pin it at a tag nobody has cut and step "Fetch and verify the
pinned bootable image" 404s.

Closes the second half of #2539.

## What this does

- restores `needs: [scope]` and `if: needs.scope.outputs.code == 'true'`, so the
  lane gates pull requests again instead of manual dispatch
- `IMAGE_TAG: v0.17.0` → `v0.18.0`
- drops the comment paragraph explaining why it was parked

## Why it was parked

The `default-microvm` image published at `v0.17.0` ships a `/init` whose
`#!/bin/sh` sits one column right of byte 0, so the kernel cannot exec it:

```
Kernel panic - not syncing: Requested init /init failed (error -8)
```

#2538 fixed the generator (`nix/lib/mk-guest.nix` had one line at three spaces
where the block sits at four; Nix takes an indented string's baseline from its
least-indented line and shifted every line, shebang included). A generator fix
cannot repair an already-published artifact, so the lane stayed parked rather than
burning ~12 minutes of a constrained runner pool per run to re-derive a known
failure.

## Before merging

#2612 split boot images onto their own release train, so **two tags are needed, in
this order**:

1. `boot-image/v0.1.0` — fires `release-boot-image.yml`, which builds the images
   from `nix/images/` and self-verifies: `assert-init-shebang.sh` on the staged
   rootfs, kernel-format assert, and a real Firecracker boot, all *before* anything
   is uploaded. The name is not a free choice — `DEFAULT_BOOT_IMAGE_TAG`
   (`crates/mvm-core/src/config.rs`) and `BOOT_IMAGE_IDENTITY_TEMPLATES`
   (`crates/mvm-core/src/release_trust.rs`) both pin it.
2. `v0.18.0` via `just release-tag 0.18.0`. `release.yml` no longer builds images;
   it downloads the newest `boot-image/v*` release's assets and re-attaches them
   under the signing identity installed binaries trust, and **hard-fails the publish
   if no such release exists**. That dual-publish is why `IMAGE_TAG` can name the
   flat CLI tag here rather than the slash-bearing boot-image tag.

Then confirm the published rootfs opens `/init` at byte 0 — first bytes
`23 21 2f 62 69 6e 2f 73 68`, not `20 23 21 …` — and mark this ready.

Repinning `IMAGE_TAG` at `boot-image/v0.1.0` is the natural follow-up once the
dual-publish window closes; it is deliberately not this PR.

## Verification

- `actionlint .github/workflows/ci.yml` — clean
- `cargo run -p xtask -- check-workflow-paths` — clean (26 working-directory, 17
  fuzz targets, 134 cargo -p packages across 24 workflows)
- rebased on current `main`; the diff is three changes to one job and nothing else
- the change validates itself: once `v0.18.0` exists, `boot-latency` runs on this
  PR, downloads that release, hash-checks it and boots it 5× at concurrency 3
  against the 6s ceiling

## What this lane does not cover

Worth stating so a green check is not read as more than it is. This lane pins a
*published* image deliberately: it measures latency drift in host-side code, and a
moving image would make a regression ambiguous between the code and the image. It
is a control variable, not a correctness gate — it cannot catch a kernel or rootfs
regression in the tree, and `scope.code` would not even fire it on a
`nix/lib/mk-guest.nix` change.

That gap is #2635: no lane boots a tree-built image on a pull request or in the
merge queue, and the generator's own guards were scoped out of the PRs that change
the generator. #2636 fixes the scope half.


auser:
Closing rather than leaving this permanently red. The lane boots a *published* image, so pinning it to `v0.17.1` before that tag exists fails at `Fetch and verify the pinned bootable image` with a 404 — no repo change can make it green ahead of the release.

The exact patch is preserved in #2539 (with the actionlint / check-workflow-paths results), and the branch `ci/2539-repin-boot-image` is kept. After the release is cut: `gh pr reopen 2542`, adjust `IMAGE_TAG` if the tag name differs, mark ready.

auser:
This is blocked on an artifact that does not exist, not on anything in the diff. Recording the diagnosis so it stops reading as a stalled PR.

## The failure

```
curl: (22) The requested URL returned error: 404
```

Step "Fetch and verify the pinned bootable image" resolves `IMAGE_TAG: v0.18.0` against `releases/download/`. **There is no `v0.18.0` release** — `gh release list` tops out at `v0.17.0`. The `release: v0.18.0 (#2582)` commit on `main` bumped the workspace version; it did not cut a release.

Which is exactly what this PR's own body predicted would happen.

## The bind

`v0.17.0`'s published `default-microvm` image is the one whose `/init` the kernel cannot exec — the reason this lane was parked. So the lane needs a *newer* published image than any that exists, and cutting `v0.18.0` is not on the table right now.

## What changed since this was written

#2612 gave boot images **their own release train**: `boot-image/v*` tags, published independently of the workspace version. `release.yml:297` already refuses to ship a release when no `boot-image/v*` exists, precisely so a fresh install cannot 404 on first boot.

So the fix is to stop pinning a workspace version here at all:

```yaml
IMAGE_TAG: v0.18.0        # a release nobody has cut
IMAGE_TAG: boot-image/vX  # the train that actually publishes images
```

That decouples this lane from the release cadence permanently — which is the durable answer, not just the unblocking one.

## What is still needed

No `boot-image/v*` tag has been cut yet either, so the repin cannot be verified until one is. Publishing the first one is a release action and needs an explicit decision.

Sequence: cut `boot-image/v1` → repin this lane at it → confirm the fetch step resolves → un-draft.

Until then this stays a draft, and red is the correct state rather than a problem to chase.

PR #2547, opened 2026-08-15
docs(plans): boot image lifecycle — gate it, version it, expose it

auser:
Design spec only — no behaviour change.

## What it covers

Four complaints turned out to share one root: a boot image is the one artifact a user cannot inspect, cannot update on purpose, and cannot tell the origin of.

| WS | |
|---|---|
| **WS1** | Boot the staged image **before** publishing it |
| **WS2** | Move images to their own `boot-image/v*` release train |
| **WS3** | Sidecar provenance fields + `mvmctl image boot status｜check｜update` |
| **WS4** | `MVM_BOOT_IMAGE` escape hatch for the dev inner loop |

## The finding that shaped it

**Nothing boots a freshly built image before it is published.** The only boot check in the repo runs against an *already published* artifact — which is why the `/init` shebang defect shipped in every release for five weeks and was reported rather than prevented. WS1 closes that and is worth landing alone.

## Why the rootfs does not move to its own repo

`nix/lib/mk-guest.nix` builds `guestAgentPkg` via `pkgs.callPackage ../packages/mvm-guest-agent.nix`, which runs `rustPlatform.buildRustPackage` **against this workspace**. The rootfs embeds `mvm-guest-agent`, `mvm-seccomp-apply` and `mvm-verity-init` compiled from `crates/`. A repo boundary drawn today inverts the dependency and makes every host↔guest protocol change a two-repo sequence.

The spec records what would have to be true first (publish `mvm-contract` and the guest binaries as versioned artifacts), and notes that the **kernel** is separable at any time — it carries no workspace source and already has its own build lane and fetch/build seam.

## Migration safety

Its own section, because that was the explicit ask. Ordering puts WS1 first so later steps move already-gated jobs; old tags keep resolving; there is a dual-publish window; sidecar fields are `#[serde(default)]` so warm caches keep deserializing; cache replacement is tmp-then-rename so a failed update leaves the working image untouched; and the default path is unchanged for a user who does nothing. Each workstream reverts independently — none is a data migration.

Known gaps are stated rather than hidden: the `aarch64` leg gets a header check rather than a boot, and a stale `DEFAULT_BOOT_IMAGE_TAG` on a long-lived branch is detectable via `protocol_version` but not yet gated on it.

## Note

`check-declared-backing` currently refuses this file: it matches `"proven"` inside **provenance**, which the spec uses four times. Fixed by #2546 — this goes green once that lands.


auser:
Diagnosed both red checks. Neither is a problem with this PR's content — it is a docs-only change and the prose is correct as written.

**`Test` — fixed, already on main.** The kernel job lost its job-level `if:` in #2548, so it now always runs and reports `success` even when out of scope. The `Test` aggregate still required `skipped` in that case, which failed every PR not touching a kernel path. Fixed in #2556 (merged, `67f66fd62`). This branch just needs to pick up main.

**`Invariant` — blocked on #2546.** The failure is:

> `declares 'preview' but uses "proven"`

That is a false positive. The document does not use "proven"; `check-declared-backing` matches the substring inside **pro*ven*ance**, which appears four times and is the correct technical term:

```
227: ### WS3 — Provenance, and a command surface over it
292: an_old_sidecar_without_provenance_fields_still_deserializes
375: Every new provenance field is ...
454: WS3 — sidecar provenance fields + mvmctl image boot status|check|update
```

#2546 fixes exactly this with a word-boundary match, and its regression test asserts `contains_word("image provenance fields", "proven") == false`.

**So: do not reword the prose.** Rewriting correct terminology to satisfy a buggy lint would bury the bug rather than fix it. Once #2546 merges, update this branch from main and both checks should clear with no content change.

Issue #2559, opened 2026-08-15
fix(sdk): Rust host_port accepts port 0 and an empty host; Python rejects both


auser:
`mvm_sdk::ctor::host_port` accepts port `0`; the Python SDK's `host_port` rejects it. Nothing in the tree notices.

## The divergence

Rust (`crates/mvm-sdk/src/ctor/network.rs:65`):

```rust
pub fn host_port(host: impl Into<String>, port: u16) -> HostPort {
    HostPort { host: host.into(), port }
}
```

No validation. `u16` admits `0`, and `host` may be empty.

Python (`crates/mvm-sdk/sdks/python/mvm/_dsl.py:754-760`):

```python
if not host:
    raise ValueError("host_port host must be a non-empty string")
if not (0 < port < 65536):
    raise ValueError(f"host_port port must be in 1..65535, got {port}")
```

So `host_port("example.com", 0)` is a `ValueError` in Python and a silently-constructed IR node in Rust. Same for an empty host.

`dns_resolver` has the matching gap: Python rejects an empty host, Rust does not.

## Why it matters beyond tidiness

`host_port` feeds the egress allowlist (`egress(...)` → `NetworkEgress.allowlist`), which is claim-10 territory. An allowlist entry with port `0` or an empty host is not a meaningful authorization, and the two SDKs disagree about whether it is even expressible.

## Why no existing gate catches it

This was surfaced by the Plan 337 WS-1 spike. It is invisible to every signature-level check: both languages export a `host_port` taking `(host, port)`, so a surface/parity comparison sees agreement. `surface_divergence.json` compares *names*. Only comparing *behaviour* against a shared golden document would catch it — which is the golden-IR behavioural gate the WS-1 decision calls for, and which is not built yet.

## Suggested fix

Decide which side is right — `u16` is a weaker constraint than `1..=65535`, and the Python rule looks correct — then make them agree. Options, in the order Plan 337's WS-1 decision favours:

1. Record the constraint (`1..=65535`, non-empty host) in the declarative constructor manifest, and let each emitter discharge it — Rust via a newtype, Python/TypeScript via a runtime check. This is the shape WS-1 settled on and needs the manifest to exist first (WS-3).
2. Interim: a `Port` newtype in Rust, or validation in `mvm_contract::ir`'s validate pass, so at least the IR rejects it at one seam.

[...]

Issue #2558, opened 2026-08-15
fix(sdk-ts): machine wrapper has no timeout and misreports output overflow as a spawn failure


auser:
The TypeScript SDK's `machine` wrapper can wait forever, and misreports an output overflow as a spawn failure. The Python SDK bounds both.

## What happens

`crates/mvm-sdk/sdks/typescript/src/_machine.ts:151`:

```ts
result = child.spawnSync(bin, ["machine", ...argv], { encoding: "utf-8" });
```

No `timeout`, no `maxBuffer`. So:

- **Unbounded wait.** Node's `spawnSync` has no default timeout. A `mvmctl machine` call that hangs hangs the caller, with no way to recover.
- **Overflow misclassified.** `spawnSync` defaults to a 1 MiB `maxBuffer`; on overflow it sets `result.error` with code `ENOBUFS`. `_machine.ts:158-162` turns any `result.error` into a spawn failure, so a machine that simply produced >1 MiB of output is reported as one that could not be started.

## What Python does instead

`_machine.py:59,63` reads `MVM_MACHINE_TIMEOUT_SEC` (default 60s) and `MVM_MACHINE_MAX_OUTPUT_BYTES` (default 16 MiB) and passes them to `run_capped` (`_subprocess.py:68-108`), which kills the whole process group and raises typed `TransportTimeout` / `TransportOutputOverflow`.

So the divergence is not two missing constants — it is two missing behaviours.

## Why this is filed rather than fixed

Plan 337 WS-2 made the SDK env-var names Rust-owned and generated. It deliberately does **not** emit `MVM_MACHINE_TIMEOUT_ENV` / `MVM_MACHINE_MAX_OUTPUT_ENV` into TypeScript, because nothing there reads them: emitting them would have cleared the last entries from `surface_divergence.json` while changing no behaviour, and the gate would then certify a parity that does not exist. The registry records per-name which surfaces actually read a value, and a unit test (`machine_vars_are_not_claimed_for_typescript`) pins this pair as not-TypeScript.

Fixing it is a behaviour change to a shipped SDK, not env-name codegen, so it was kept out of that change.

## Suggested fix

1. Add `crates/mvm-sdk/sdks/typescript/src/_env.ts` with `envFloat` / `envInt` helpers mirroring `_subprocess.py:33-50` — note Python silently falls back to the default on a malformed value; match that or fix both.
2. Pass `timeout` and `maxBuffer` to `spawnSync` in `_machine.ts`.
[...]

PR #2561, opened 2026-08-15
feat(337): own the SDK env-var names in Rust and generate both bindings

auser:
Executes Plan 337 WS-1 (the spike) and WS-2 (Tier E). Stacked on #2501, which carries the plan document and the s27 fixtures this builds on — **retarget to `main` once that merges**.

## WS-1: the spike re-scoped the plan rather than stopping it

The plan asks whether a constructor manifest can be *extracted* from the Rust `ctor` fns. Both candidate mechanisms were built against the real sources and scored against a criterion written down **before** either existed: reproduce the *Python* signature of `python_deps` and `dns_resolver`, including `tool="uv"`, `port=53`, keyword-only calling and the `"pip-tools"` alias.

Both work. Neither meets it.

**The plan's assumption is backwards.** It calls the attribute/`inventory` mechanism "more precise". It recovered strictly *less*: an attribute macro is invoked once per item and sees only that item's tokens, so it cannot resolve `python_deps` → `python_deps_with(lockfile, PythonTool::Uv)`. The whole-file `syn` parse holds every fn at once and recovers that default in a second pass.

| axis | attribute + `inventory` | `syn` from xtask |
| --- | --- | --- |
| params, types, generic bounds, docs | recovered | recovered |
| **`tool = PythonTool::Uv`** (delegated default) | **not recovered** | **recovered** |
| `port=53`, keyword-only, `1..=65535`, `"pip-tools"` | not recovered | not recovered |

**The remaining four are not parser failures.** They're absent from the Rust source at any level of effort, because Rust doesn't need them: `port: u16` *is* Python's range check, `tool: PythonTool` *is* its enum check. Rust is the surface that discharges those constraints **statically**; Python and TypeScript need runtime checks precisely because they lack the types.

So the plan's two requirements — byte-compatible generated Python, *and* a manifest extracted from the ctors — are mutually unsatisfiable. That's a specification failure, not a tooling failure, which is why gate 1.5 didn't fire.

**Decision:** author the manifest declaratively, recording *constraints* rather than validation code, so each emitter decides whether the type system or a runtime check discharges one. Rust ctors stay hand-written (generating them forces `-> Result`, wrecking the prelude's composition for ~50 lines inside a shipping crate). `syn` is kept but re-scoped to a fail-closed coverage gate over `lib.rs`'s `pub use` lists — **not** `pub fn` in `src/ctor/`, since `mod ctor` is private and that would over-report ~40% — paired with a golden-IR behavioural gate.

Dependency cost independently rules out the attribute route: a new workspace member plus `inventory` inside `mvm-sdk`, which is `crate-type = ["lib", "cdylib"]` and `dlopen`ed by both SDKs. `syn` in `xtask` costs zero shipped-closure delta.

**Tier A is not descoped to a hand-port.**

## WS-2: Tier E, the proving ground

A `macro_rules!` registry in `crates/mvm-sdk/src/env.rs` emits both the `pub const`s and a table. `emit_sdk_env` writes `schema/sdk-env-v0.json`; `xtask/src/gen_sdk_surface.rs` renders `mvm/_env/vars.py` and `src/_env/vars.ts`. All three are drift-gated by `check-stubs` (which already runs twice — `lint-policy` and inside BDD).

`typescript_only_absent_from_python` is now `[]`.

**Five names, not the four scoped.** The fifth is `MVM_CLI_BIN_ENV`, and it's the best argument for the tier: it was written out **four** times — `machine.rs`, `facade.rs` (twice in one crate, the second commented "shared with `machine.rs`" when it was a copy), `_cli.py`, `_cli.ts` — and because all four agreed, **no gate in the repo could see it**. Counting divergence entries would never have found it.

### Two decisions worth reviewing

**The existing generators cannot emit a constant.** `json-schema-to-typescript` produces `export type` only, and `tsc` erases types — so a "generated" constant would be absent from the built ESM namespace and invisible to the s27 check that reads `Object.keys(mvm)`. The drift gate would have certified a binding that does not exist. Hence a small hand-written emitter in xtask; determinism is unaffected since output depends only on that function.

**Emitting every name into every language would have been dishonest.** Clearing the last two divergence entries was one line away — but TypeScript's `_machine.ts` reads no environment at all, so `MVM_MACHINE_*` would have been dead exports certifying a parity that doesn't exist. Each row instead declares the surfaces that *read* it, a new s27 step checks that claim **in both directions**, and a unit test pins the pair as not-TypeScript. They stay recorded as the **behaviour** gap they are.

## Also here

A pre-existing ~50% flake in the s27 TypeScript live fixture, in its own commit so it can be dropped: `process_.wait()` returns a Promise and spawns asynchronously while every other verb is `spawnSync`, so unawaited it raced the next call. Measured on the **unmodified base**: 9/20 correct; with the `await`: 30/30. Not introduced here — my change only perturbed timing — but the suite gates PRs now, so it had to be deterministic.

## Filed, not fixed

- #2558 — TS `_machine.ts` has no `timeout`/`maxBuffer`: unbounded hang, and 1 MiB `ENOBUFS` misreported as a spawn failure. A behaviour change to a shipped SDK, not env-name codegen.
- #2559 — Rust `host_port` accepts port `0` where Python rejects it. `u16` is not `1..=65535`. Invisible to any signature-level gate, which is the argument for the golden-IR gate.

## Verification

- `cargo +nightly fmt --all -- --check`, `clippy --workspace --all-targets -D warnings` — clean
- `cargo nextest run --workspace` — **11,721 passed**; doctests clean
- `check-stubs` — no drift, **and proven to exit 1 naming the file** on a hand-edited constant
- `check-honesty`, `check-no-overclaim`, `check-forbidden-deps`, `check-no-spec-refs-in-comments`, `check-workspace-dep-inheritance`, `check-sprint-append`, `check-file-size` — clean
- Python **212 passed, zero test-file edits** (the byte-compatibility bar); TypeScript 135 passed
- BDD 193 scenarios, including the new "every Rust-owned env-var name reaches the surfaces it claims"

## Not in scope

WS-3–WS-5, WS-7, WS-8, and explicitly **WS-6** (Tier C, the ~1,165-line remote-function surface).


auser:
Superseded by #2568.

This PR reads as MERGED but none of its content reached `main`. It targeted `feat/sdk-bdd-coverage` (#2501); that branch squash-merged to `main` first, so merging this one afterwards landed it on a branch that is now a dead end. Verified by content — `crates/mvm-sdk/src/env.rs` is absent from `main`.

#2568 is the same two commits rebased onto `main`.

PR #2579, opened 2026-08-16
fix(hostd): decode the admitted plan as its signed envelope in the wall-clock timer

### LINKED ISSUE #2555: libkrun/HVF supervisor refuses every admitted workload: wall-clock timer decodes a SignedExecutionPlan as an ExecutionPlan
## Symptom

Any `mvmctl machine run --image …` on the libkrun/HVF path fails:

```
supervisor: refusing to boot a bounded workload it cannot audit: decoding the admitted plan for the wall-clock timer
[mvm] Guest console at ~/.mvm/vms/<name>/console.log was empty before transient cleanup.
Error: guest agent did not become reachable within 30s
```

Reproduced on macOS 26.5.2 / arm64 against `docker.io/library/alpine:latest`, on both the warm-claim and cold paths, with `mvmctl` and `mvm-libkrun-supervisor` built from the same checkout (so this is not a stale-aux-binary case — `~/.mvm/bin` holds no supervisor copy).

## Cause

`plan_json` is the **signed** plan. All three producers build it the same way:

- `crates/mvm-cli/src/commands/vm/exec.rs:492` — `serde_json::to_string(c.admitted.signed())`
- `crates/mvm-cli/src/commands/vm/session.rs:1061`
- `crates/mvm-cli/src/commands/vm/invoke.rs:325`

`SignedExecutionPlan` is `pub struct SignedExecutionPlan(pub SignedPayload)` (`crates/mvm-core/src/plan/signing.rs:33`), so it serializes as the signed envelope, not as plan fields.

But `arm_for_supervisor` decodes that value directly as a bare plan
(`crates/mvm-hostd/src/supervisor/wall_clock.rs`):

```rust
let plan: ExecutionPlan = serde_json::from_value(value.clone())
    .context("decoding the admitted plan for the wall-clock timer")?;
if plan.resources.timeouts.exec_secs == 0 { return Ok(None); }
```

That can never succeed against an envelope.

The correct two-step decode already exists elsewhere — `crates/mvm-runtime/src/checkpoint/mod.rs:543-553`:

```rust
let signed: SignedExecutionPlan = serde_json::from_str(plan_json)?;
let plan: ExecutionPlan = serde_json::from_slice(&signed.0.payload)?;
```

## Why it takes the whole boot down

The decode happens **before** the `exec_secs == 0` early return, so it fails for *every* plan, bounded or not — not just workloads that requested a wall-clock bound. The caller then fails closed by design (`mvm-libkrun-supervisor.rs:285`, and the same shape in `mvm-hvf-supervisor.rs:337`), which is the right posture for an unauditable bound but here converts a decode bug into "no admitted workload boots at all".

`SupervisorConfig::plan`'s doc comment (`crates/deps/libkrun-sys/src/supervisor.rs:106`) says "JSON-encoded `ExecutionPlan`", so the two sides disagree about the field's contract; the comment matches the timer's expectation and not what is actually sent.

## Suggested fix

Decode via `SignedExecutionPlan` then the inner payload, reusing the checkpoint path's shape, and correct the `SupervisorConfig::plan` doc comment to say signed envelope. A regression test should feed `arm_for_supervisor` exactly what `exec.rs` produces — the bug is that no test does.

## Why CI is green

The live-boot lanes do not gate a PR: `Boot latency ceiling` is `if: github.event_name == 'workflow_dispatch'` in `ci.yml`, and `security.yml` runs on tags/nightly. Nothing on the PR path boots a libkrun/HVF workload end-to-end, so a total boot failure is invisible to it.

Found while attributing launch latency; unrelated to the change that surfaced it.

auser:
Closes #2555

## What was broken

`arm_for_supervisor` deserialized `SupervisorConfig.plan` as a bare `ExecutionPlan`. Every producer puts the **signed envelope** there — `plan_admission::thread_plan_onto_config` serializes `admitted.signed()` onto `VmStartConfig.plan_json`; `spec_map::workload_plan_binding` re-parses that string into `PlanBinding`; the libkrun and HVF relays hand the value to the supervisor untouched. `SignedExecutionPlan` is `#[serde(transparent)]` over `SignedPayload`, so it is `{payload, signature, signer_id}` — disjoint from every `ExecutionPlan` field. The decode could never succeed.

## Blast radius: every admitted workload, not only bounded ones

The decode runs *before* the `exec_secs == 0` early return, so it fails for any plan-bearing boot. Both supervisors fail closed on the error by design (`mvm-libkrun-supervisor.rs`, `mvm-hvf-supervisor.rs`), which is correct for an unauditable bound but converts a decode bug into "no admitted workload boots" on the two macOS backends.

The pair of new tests pins both halves. Against unfixed code, **both** fail — including the `exec_secs = 0` case, which is the common one:

```
        FAIL [   0.020s] (5/9) mvm-hostd supervisor::wall_clock::tests::an_unbounded_admitted_plan_boots_without_a_timer

    thread '...an_unbounded_admitted_plan_boots_without_a_timer' panicked at crates/mvm-hostd/src/supervisor/wall_clock.rs:467:10:
    an unbounded workload must boot: decoding the admitted plan for the wall-clock timer

    Caused by:
        unknown field `payload`, expected one of `schema_version`, `plan_id`, `plan_version`, `tenant`, `workload`, `runtime_profile`, `image`, `environment`, `resources`, `grants`, `admission_profile`, `network_policy`, `network_mode`, `l3_network`, `network_limits`, `snapshot_at`, `build_provenance`, `fs_policy`, `secrets`, `egress_policy`, `redaction`, `reversible_replacement`, `tool_policy`, `artifact_policy`, `audit_labels`, `key_rotation`, `attestation`, `release_pin`, `post_run`, `valid_from`, `valid_until`, `nonce`, `agent_verbs`, `bundle`, `deps_volume`, `shares`, `services`, `stream_edges`, `stream_retention`

        FAIL [   0.020s] (6/9) mvm-hostd supervisor::wall_clock::tests::a_bounded_admitted_plan_arms_the_timer

    thread '...a_bounded_admitted_plan_arms_the_timer' panicked at crates/mvm-hostd/src/supervisor/wall_clock.rs:443:10:
    the supervisor must be able to read the plan it was admitted under: decoding the admitted plan for the wall-clock timer

    Caused by:
        unknown field `payload`, expected one of `schema_version`, ...

     Summary [   0.066s] 9 tests run: 7 passed, 2 failed, 1739 skipped
```

After the fix, 9/9 pass. The assertions are on the specific outcome — `Ok(Some(guard))` for a 900s bound, `Ok(None)` for an unbounded plan — not on "an error occurred", which the unfixed path also satisfies.

## Shape of the fix

The bug is a shape guess at a boundary, so the producer is typed rather than the one call site corrected:

- **`PlanBinding.plan_json: serde_json::Value` → `PlanBinding.plan: SignedExecutionPlan`.** A binding carrying a bare plan is now unconstructable, and `workload_plan_binding` decodes `VmStartConfig.plan_json` as an envelope, so a non-envelope yields no binding instead of a boot-time refusal.
- **`SignedExecutionPlan::payload_plan()`** is the single envelope→plan decode. It replaces four hand-rolled `from_slice(&signed.0.payload)` copies in `mvm-core::plan::signing` and is what the timer now calls.
- **`PlanBinding::plan_carrier()`** owns the one hop where the type is erased for the supervisor's JSON config. `libkrun-sys` deliberately keeps that field a serde boundary with no typed coupling, so the erasure stays — but it happens in exactly one place, and the supervisor reconstitutes a `SignedExecutionPlan`, never a bare plan.
- **`SupervisorTimerInputs.plan_json` → `signed_plan_json`**, so the field name carries the shape, and `SupervisorConfig::plan`'s doc comment stops claiming it holds a bare `ExecutionPlan` (that comment matched the timer's wrong expectation, not the wire).

## Why nothing caught it

Every fixture on this path modelled the shape no producer emits — `serde_json::json!({"resources": {"timeouts": {"exec_secs": 30}}})` in the spec-map, libkrun-relay and HVF-relay tests. `PlanFixture::build_signed()` (plus a `.exec_secs()` override) now supplies the envelope the wire actually carries, and those tests assert through the same two-step the supervisor uses. A new `a_bare_plan_on_the_wire_yields_no_binding` pins the negative.

The live-boot lanes that would have caught this do not gate a PR: `Boot latency ceiling` is `workflow_dispatch`-only and `security.yml` runs on tags/nightly.

## Claim ledger

Preview claim 18's wall-clock witnesses (`an_expired_workload_is_killed_and_the_kill_is_audited`, `a_wall_clock_bound_needs_a_clock_that_can_stop_the_workload`) both drive the timer from an in-memory plan and passed throughout — neither covers the arming step against the wire carrier. `a_bounded_admitted_plan_arms_the_timer` is added to the ADR-001 ledger row and to `model/claims.toml`, with a note in the ADR saying what the other two do not cover. `CONFORMANCE.md` regenerated via `check-conformance --write`.

Left alone deliberately: the `MVM-SEC-18` comment in `model/claims.toml` says "Wall clock has no mechanism at all ... nothing stops the workload", which contradicts ADR-001's own §2 ("Wall clock is enforced only where a supervisor process holds the plan"). That divergence predates this change and reconciling the claim's prose is a maintainer call, not a drive-by edit.

## Verification

- `cargo +nightly fmt --all -- --check`
- `cargo nextest run --workspace` — 11,803 passed, 26 skipped
- `cargo test --workspace --doc`
- `just check-gated` (zigbuild `--target x86_64-unknown-linux-gnu --workspace --all-targets`, plus `mvm-conformance --features bdd`) — exit 0
- The `ci.yml` xtask gate list, including `check-claim-catalog`, `check-conformance`, `check-witness-citations`, `check-honesty`, `check-no-overclaim`, `check-no-spec-refs-in-comments`, `check-single-exec-secs-writer`

Not verified: a live `mvmctl machine run --image` boot on macOS 26 — that needs the hardware path the issue reproduced on, and the fix is pinned by the wire-shape test instead.


auser:
Superseded by #2564, which merged while this was in flight and closed #2555.

Main already carries the fix — `decode_admitted_plan` in `wall_clock.rs` — and the equivalent tests:

| this PR | #2564 |
|---|---|
| `a_bounded_admitted_plan_arms_the_timer` | `a_signed_plan_from_the_launch_path_arms_the_timer` |
| `an_unbounded_admitted_plan_boots_without_a_timer` | `a_signed_plan_without_a_bound_arms_no_timer` |
| (fail-closed on garbage) | `an_undecodable_plan_still_fails_closed` |

Both diagnoses agree, including the part that made it a total outage rather than a bounded-workload one: the decode sits ahead of the `exec_secs == 0` early return, so an unbounded plan fails identically.

**Why not rebase this on top.** #2564 decodes *either* shape on purpose:

> The bare form is still accepted. Not every producer goes through the admission path — the checkpoint restore path and the attach path build this value themselves.

This PR instead types `PlanBinding.plan` as `SignedExecutionPlan`, which is the stronger design in isolation but would refuse those two producers. Making it correct means converting them as well, and that is a larger change than the outage needed and not one to land as a conflict resolution.

**The one thing here that main lacks** is the claim-ledger entry. Claim 18's wall-clock witnesses all drive the timer from an in-memory plan; none covers arming against the wire carrier, which is precisely why a total launch failure was invisible to the ledger. Following up separately with that row, citing #2564's test rather than this branch's.

Branch untouched — `gh pr reopen 2579` if the typed-binding direction is wanted as its own change.

PR #2605, opened 2026-08-16
refactor(build): one resolver for mvm-network-endpoint, not two

auser:
`mvm_vmm::host::aux_bin::resolve` is the canonical search for a per-VM helper,
and its ordering is deliberate: `$<ENV_VAR>` → `$MVM_AUX_BIN_DIR` → beside
`current_exe` → workspace `target/`. The build script's dir comes first so a
freshly built helper is not shadowed by a stale copy sitting next to a dev
`cargo run` exe, which cargo never refreshes for a `[[bin]]` it did not build.

The builder path had its own 86-line copy that omitted `$MVM_AUX_BIN_DIR`
entirely, so it preferred exactly the stale copy the ordering exists to avoid.

That is not theoretical. Diagnosing the guest-egress failures this week, a fix
to `mvm-egress-client` appeared to do nothing three runs in a row: the embedded
binary was not being rebuilt (since fixed) *and* the builder was resolving a
stale endpoint. Each round cost a full Stage 0 boot to discover.

The duplicate is deleted and the builder delegates.
`resolve_network_endpoint_path` is now `pub` in `network_endpoint_spawn`, so the
binary's name and its override variable have one definition rather than two that
can drift.

This also drops the builder's run-time `cargo build -p mvm-hostd` fallback,
which is the intended behaviour: `aux_bin::resolve` deliberately never builds,
and errors with the `just build-supervisors` recovery hint instead. The workload
path has always behaved that way.

No new test. The property the builder was missing — aux dir before exe dir — is
already pinned by `candidate_order_is_aux_then_exe_then_targets`; a second copy
of that assertion in `mvm-build` would drift rather than protect.

Closes sections 1 and 2 of
`specs/plans/2026-08-15-aux-helper-binary-freshness.md`; section 2's watch-list
gaps were closed separately, and the plan is updated to say so.

## Verification

- `cargo nextest run -p mvm-build -p mvm-vmm` — 1332 passed
- `cargo clippy -p mvm-build -p mvm-vmm --all-targets -- -D warnings` — clean
- `xtask check-declared-backing`, `check-deferrals` — clean

### DIFF
--- a/crates/mvm-build/src/libkrun_builder.rs
+++ b/crates/mvm-build/src/libkrun_builder.rs
@@ -471,91 +471,14 @@ impl Drop for BuilderVsockEgressEndpoint {
     }
 }
 
+/// Locate the endpoint binary through the canonical resolver.
+///
+/// Delegates rather than re-deriving the search order: this path used to have
+/// its own copy that skipped `$MVM_AUX_BIN_DIR`, so a stale binary beside the
+/// dev exe shadowed the freshly built one and edits appeared to do nothing.
 fn resolve_network_endpoint_path() -> Result<PathBuf, BuilderVmError> {
-    if let Some(path) = std::env::var_os("MVM_SUBSTITUTION_ENDPOINT_PATH").map(PathBuf::from) {
-        if path.is_file() {
-            return Ok(path);
-        }
-        return Err(BuilderVmError::ExtractionFailed(format!(
-            "MVM_SUBSTITUTION_ENDPOINT_PATH points at {} which is not a file",
-            path.display()
-        )));
-    }
-
-    if let Ok(current_exe) = std::env::current_exe()
-        && let Some(dir) = current_exe.parent()
-    {
-        let candidate = dir.join("mvm-network-endpoint");
-        if candidate.is_file() {
-            return Ok(candidate);
-        }
-    }
-
-    let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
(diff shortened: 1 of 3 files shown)


auser:
Closing in favour of #2602, which landed first and fixes the harm this was aimed at.

We solved the same defect — the builder running a stale `mvm-network-endpoint` — from opposite ends:

- **#2602** keeps the builder's own resolver and refuses a next-to-exe copy older than the running `mvmctl` (`endpoint_predates_running_exe`).
- **This PR** deleted that resolver so the builder delegates to `aux_bin::resolve`, whose ordering puts `$MVM_AUX_BIN_DIR` — the build script's fresh output — first.

Rebasing this on top of #2602 would delete `endpoint_predates_running_exe`, and that check covers a case delegation does not: an endpoint that is stale *in the aux dir itself*. Trading a real safety net for tidiness is the wrong side of that deal, so I am not doing it silently.

The duplication is still there, and it is still worth removing — but the strictly-better version of this change now has to **move the mtime refusal into `aux_bin::resolve`** so every per-VM helper gets it, and only then delete the copy. That is a change to freshly-landed code in someone else's seam, so it belongs to whoever owns it rather than to a drive-by.

Leaving `specs/plans/2026-08-15-aux-helper-binary-freshness.md` §1 open on main, with that as the revised approach. Context for why it matters: during the guest-egress work a fix appeared to do nothing three runs running, because the embedded binary was not rebuilt *and* the builder preferred a stale copy — each round cost a full Stage 0 boot.

Issue #2543, opened 2026-08-15
Guest egress is dead on main: FlowMux guest adapter shipped without host identity provisioning


auser:
## Symptom

`mvmctl machine run --image python:3.12 -- python -c "print(2 + 2)"` fails during the Stage 0
workload-kernel build with a misleading network error:

```
warning: unable to download 'https://github.com/NixOS/nixpkgs/...': Failed to connect
         to github.com port 443 via 127.0.0.1 after 0 ms: Could not connect to server
Error: build the dm-verity-capable workload kernel
```

## Root cause

Three lines earlier in the Stage 0 console log:

```
stage0-init: forked mvm-egress-client pid=227 from /mvm-bins/mvm-egress-client
mvm-egress-client: failed to load guest signing key: failed to read guest signing key
                   from /run/mvm/flowmux-guest-signing-key: No such file or directory
stage0-init: mvm-egress-client exited before proxy became ready (exit code 1)
```

2821e2dd4 ("Feat/316 guest flowmux adapter", #2480) made the in-guest `mvm-egress-client`
FlowMux-only. It loads a per-boot guest signing key and a host-signer trust anchor **before**
binding its loopback listener (`crates/mvm-agentd/src/bin/mvm-egress-client.rs:52-90`; the
bind is the last step, inside `flowmux_egress::run`). Nothing in the workspace writes
`/run/mvm/flowmux-guest-signing-key`, so the client exits 1, nothing listens on
`127.0.0.1:1080`, and every proxied fetch gets an instant `ECONNREFUSED` — hence "after 0 ms".

The mismatch is symmetric: no host spawn site selects FlowMux either. All five production
constructions of `SubstitutionSpawnParams` hard-code `flowmux_identity: None`
(`workload_runner/runner/spawner.rs:51`, `builder_runner/runner.rs:159`,
`wasm_backend.rs:454`, `driver/libkrun_legacy.rs:84`, `driver/hvf_legacy.rs:247`), and Stage 0
hand-builds its own endpoint config with `"egress_mode": "raw"`
(`crates/mvm-build/src/libkrun_builder.rs:317`). The host serves Raw/Wire; the guest speaks
only FlowMux.

## Blast radius

All guest egress, not just Stage 0 — the same binary is the loopback proxy for workload
microVMs (`nix/lib/mk-guest.nix:786`) and the resident builder VM
(`crates/mvm-build/src/bin/mvm-host-vm-init.rs:1451`).

## Secondary defects found while diagnosing

[...]

PR #2609, opened 2026-08-17
feat(net): finish the one-transport cutover — WS4–WS7 and the gate that keeps it

auser:
Completes the FlowMux cutover from `specs/plans/2026-08-15-flowmux-single-transport-cutover.md` (#2543). WS0–WS3 landed in #2545; this is WS4–WS7 plus the BDD scenarios.

**Rebased onto main rather than continuing the old branch.** #2545 squash-merged at 16:46Z and these commits were pushed to `fix/one-transport-flowmux-cutover` afterwards, so they were stranded on a branch with no open PR. #2599's fix was stranded the same way — it merged into that branch at 21:37Z, after its base had already gone to main — so it rides along here. Verified by content, not by PR state.

## WS4 — substitution as a typed HTTP flow

`OpenHttp` → `HttpRequestHead` → `HttpRequestBody`*, answered by `HttpResponseHead` → `HttpResponseBody`* → `HttpComplete`. The head declares `body_len` and the host counts: treating a half-close as the end makes "finished sending" and "went away mid-body" the same event, and the host would forward a truncated request rather than refuse it.

Substitution itself did not move. The `Http` arm assembles frames into the `WireRequest` that `SubstitutionService::process` already takes and calls it, so placeholder resolution, destination binding, the claim-10 gate and payload-free audit stay where they are. `process` runs on the runtime rather than the session's blocking read loop — it is async and tokio's `block_on` panics inside `spawn_blocking`, and an inline forward would stall every other flow behind one slow upstream.

Two hard gates: an `OpenHttp` on an endpoint with no substitution service is refused rather than forwarded unsubstituted, and an endpoint carrying secrets with no substitution service refuses to boot. Placeholders are minted and injected regardless of both; without something to resolve them the guest sends `mvm-secret-<hex>` to a real upstream.

## WS5 — ICMP as a one-shot flow

`IcmpEcho` → `IcmpReply` | `IcmpRefused`, shaped like a DNS lookup: the reply confirms the flow and ends it. The decision stays `icmp_handler::serve_request` — parse, bounds, admission, rate, echo — so no transport can drift on what is allowed. `emit_icmp_echo_blocking` loses its Linux gate; it was gated because the blocking vsock transport was its only caller, and leaving it would have left an echo served on macOS unaudited.

## A blocking client, because the sealed agent has no tokio

`forward_proxy` and `icmp_client` live in `mvm-guest-agent`, whose closure is deliberately tokio-free. `flowmux_sync::SyncFlowMux` opens a connection, does one authenticated exchange and closes. `Session::guest` already takes `Read + Write` and the frame codec is `no_std`, so nothing new was needed. Several sessions is not several transports: one wire contract, one auth boundary, one gate.

## WS6 — the alternatives, and the gate

Removed: `EgressMode::Raw`, `serve_raw`, the guest-facing `raw_egress` dispatcher and its 14 tests, `substitution_client::substitute`, and the `can_skip_substitution_assembly` shortcut only `raw` used. `resolve_hostname_ips_pure` stays — `dns_handler` and `socks5_udp` pin names through it on the FlowMux path too.

**`EgressMode::Wire` and `substitution_client::relay` are deliberately kept.** The plan listed them for deletion, but their one consumer is the wasm tier's `mvm:egress` host import, which runs in the *host* process and connects to the endpoint's Unix socket: host-internal IPC between two host processes, not a guest speaking to its host. Moving it would mean giving a host process a guest identity — a different change with a different argument.

`xtask check-one-guest-protocol` is the recurrence gate. No retired line marker in guest or host sources, and every guest file that opens the egress port constructs a FlowMux client — checked against the file's own text rather than an allowlist of paths, because an allowlist is silenced by adding a name to it, which is exactly the move someone makes while introducing what the gate exists to catch. Both halves were confirmed red against a planted regression and green after.

## WS7 — readiness fails closed

An endpoint binds and prints its handshake line before the guest boots, so "ready" never meant a guest had reached it. The endpoint records its first authenticated session to `substitution.session`, and `refuse_launch_without_endpoint_session` fails the launch when it is absent, distinguishing "the endpoint exited" from "the endpoint is up and no guest authenticated" — the second is what a mismatched identity looks like.

Checked after activation, in `exec.rs` once the agent is reachable: a pre-boot wait deadlocks on an event the wait itself prevents. A boot with no endpoint at all — no secrets, deny-egress — is admitted; the pid file is the evidence one was spawned.

## Tests

The guest↔host substitution round-trip captures every frame the guest emits and asserts the real credential appears in none of them — claim 13 on this transport, checked on the bytes rather than assumed from the design. Five BDD scenarios state the rules the gate enforces; the secret-bearing one calls the real `refuse_secrets_without_substitution`, which is why that function moved from the endpoint binary into the library rather than being copied into a step.

## Verification

- `cargo nextest run --workspace` — 11969 passed
- `just bdd` — 204 passed, 1 pre-existing skip
- `just check-gated`, and the full `cargo-zigbuild --target x86_64-unknown-linux-gnu --workspace --all-targets --all-features` — clean
- xtask: `check-one-guest-protocol`, `check-uniform-vsock-egress`, `check-vsock-only-egress`, `check-no-spec-refs-in-comments`, `check-build-egress-callers`, `check-l3-expansion-freeze`, `check-no-gateway-names`, `check-claim-catalog`

## Not covered

The live `machine run --image python:3.12` still does not print `4`. It now fails past egress entirely, in the supervisor's wall-clock plan decode (`Preview` claim 18's fail-closed path), which nothing here touches — #2579 was the fix for that and is currently closed.


auser:
Pushed `632a69bf3` from another session to unblock the `Invariant` lane — flagging it here since this is your branch.

`check-mutation-witnesses` was failing with:

```
[error] surface lost crates/mvm-agentd/src/substitution_client.rs — claim(s) [13] no longer resolve there
```

WS4 deletes `substitution_client::substitute`, and the baseline had that file pinned as claim 13 surface because the function name matched.

I checked before re-pinning rather than just running `--write-baseline`, because a re-pin that quietly drops a claim's coverage is exactly what this gate exists to prevent. The pin was wrong to begin with: claim 13 is "no raw secret value crosses the broker channel", and the guest relay never held a secret — it carried an opaque placeholder, which is the design. The function that does hold raw credentials is `keyholder::substitution::substitute` on the host, and it stays pinned along with `substitution.rs`, `vm_backend.rs`, `network/stages.rs` and `network_endpoint_proxy.rs`. Six files to five, and the one that left was never evidence for the claim.

`check-claim-catalog` stayed green throughout, correctly — ADR-001's `substitute` witness resolves to the host-side function, not the deleted guest one.

One note on reproducing: this fails in a clean checkout but **passes** in `.worktrees/mvm-one-transport`, at the same SHA. If you gate there and see green, that is the anomaly, not CI.

PR #2615, opened 2026-08-17
feat(image): record where a boot image came from, and let it be inspected

auser:
Implements **WS3** of `specs/plans/2026-08-15-boot-image-lifecycle.md`.

Stacked behind #2612 (WS2) and #2614 (WS4) — it touches `config.rs`, `default_microvm.rs`, and `builder_vm.rs`, which both of those also touch. I'll rebase as they land.

## Provenance

`GuestSidecar` described *what* an image is and nothing about *which* image it is. Five fields close that — `image_tag`, `source`, `built_at`, `protocol_version`, `generator_rev` — every one `#[serde(default)]`, no schema bump. `source` is the field that turns "the split misfired" from a suspicion into a readout; `protocol_version` is the one that makes a future repo split checkable.

The compatibility guarantee is asserted against a **literal pre-provenance JSON blob**, because a round-trip of the new struct cannot catch a missing default. Deleting one `#[serde(default)]` makes it fail with `missing field 'imageTag'` — it is load-bearing, not decorative.

### Three honest gaps, not fabricated values

- **`built_at` is empty from Nix.** A hermetic build has no clock. The host stamps it; `status` falls back to the sidecar's mtime.
- **`generator_rev`** resolves for a git flake ref and is empty for the `path:` flake mvmctl uses locally. No ambient git in a Nix build, and nothing invented.
- **`image_tag`** comes from the environment the release supplies rather than being hardcoded. Hardcoding it would drift against the constant that names the line and keep reading as truth after the line moved.

### A bug the plan didn't know about

The builder-VM job script overwrote `/out/mvm-meta.json` by `nix eval`-ing `passthru.mvm`, which would have discarded the flake's sidecar entirely and made the emitter edit a silent no-op. It now prefers the sidecar the image published.

## Commands

`mvmctl image boot status | check | update`.

- **`status`** — cache only, no network. Never prints a blank; an unrecorded field says `(unrecorded)`.
- **`check`** — lists releases and filters the image namespace rather than asking for `/releases/latest`, which answers across *every* tag namespace and would happily return a CLI release. Only `Behind` exits nonzero, so it gates a script without output parsing. "No published line" is a clean exit 0 — which is today's state.
- **`update`** — stages beside the target on the same filesystem so the swap is a rename, keeps the previous entry until the swap completes, and restores it if the second rename fails.

`a_failed_update_leaves_the_previous_image_in_place` is the test that got written first and mutated hardest: serve assets whose bytes don't match the manifest, run update, assert the original bytes survive. Downloading straight into the live entry makes it fail.

## Reuse

`check`/`update` extend `update.rs` and share its `MVM_UPDATE_API_URL` / `MVM_UPDATE_DOWNLOAD_URL` overrides. Hashing calls the existing verified-manifest path. Asset names come from the existing table. The cache path is one new `mvm_core::config` helper that `default_microvm.rs` now also uses, so acquisition and inspection cannot disagree about where the image lives.

## ⚠️ Depends on #2612's trust change

`update` verifies against a signing identity, and a boot-image release signs under `release-boot-image.yml@refs/tags/boot-image/v*` — which nothing accepted. Today that means `update` **fails closed** rather than admitting unverified bytes, which is the right failure but a permanent one. #2612 adds `accepted_boot_image_identities` as a *separate* list from the CLI one (so a boot-image signature cannot validate a CLI tarball). Wiring this path onto it is part of the rebase.

Also: only the **prod** variant is fetchable. Dev is built locally by design — it carries an interactive shell.

## Verification

`cargo nextest run -p mvmctl -p mvm-cli -p mvm-build -p mvm-core` → **4951 passed, 0 failed**. Doctests, workspace `clippy --all-targets -D warnings` (no `#[allow]` added), nightly `fmt --all --check`, and `check-test-home-isolation`, `check-no-spec-refs-in-comments`, `check-cli-runtime-surface`, `check-file-size`, `check-single-home`, `check-two-surfaces`, `check-nextest-groups`.

Every test was proven red first; the report notes which mutation killed each. A smoke run of the built binary caught a real bug the suite missed — `ui::info` is verbosity-gated, so `status`'s header and `check`'s verdict were invisible by default.


auser:
Your `Invariant` failure is not your test — it is a gap in the gate. Fix is up as #2616.

`check-no-orchestration-server.sh` flags `crates/mvm-cli/src/commands/image/boot/tests.rs:70`:

```
let listener = TcpListener::bind("127.0.0.1:0").expect("bind loopback");
```

The script strips `#[cfg(test)] mod tests { .. }` by scanning a file for the attribute, but `boot/mod.rs:51` uses the out-of-line form — `#[cfg(test)]` / `mod tests;` — so the attribute stays in the parent and `tests.rs` carries no marker. Every line of it read as production code. The `!**/tests/**` glob does not cover it either, since that matches a `tests` directory rather than a `tests.rs` file under `src/`.

#2616 makes the gate resolve the declaration instead of trusting a filename, so a real server in a file called `tests.rs` is still caught when its parent ships it. Nothing to change on your side — re-run `Invariant` once #2616 lands and it should go green.

PR #2616, opened 2026-08-17
fix(ci): let the orchestration gate see an out-of-line test module

auser:
Unblocks #2615, whose `Invariant` lane is red on a legitimate test.

## The gap

`check-no-orchestration-server.sh` strips `#[cfg(test)] mod tests { .. }` by scanning the file for the attribute. It cannot see the out-of-line form:

```rust
// crates/mvm-cli/src/commands/image/boot/mod.rs
#[cfg(test)]
mod tests;
```

The attribute stays in the parent; the body lives in `boot/tests.rs`, which carries no marker of its own. Every line of it read as production code, so a test binding an ephemeral loopback port —

```rust
let listener = TcpListener::bind("127.0.0.1:0").expect("bind loopback");
```

— looked exactly like a new orchestration server. The `!**/tests/**` glob does not help: that matches a `tests` *directory*, not a file named `tests.rs` under `src/`.

## The fix

Resolve the declaration. For `foo/tests.rs` the gate now reads `foo.rs` / `foo/mod.rs` and asks whether the `mod tests;` there is gated out of the build.

Exempting by filename would have been a line shorter and wrong: it would let a real server hide in a file called `tests.rs`. Resolving the declaration keeps the question the gate is actually asking — *is this compiled into a shipped binary?*

## Witness

`scripts/check-no-orchestration-server.test.sh`, wired into the `Lint` job beside the existing release-asset gate tests, and shellchecked with it.

```
ok   — out-of-line cfg(test) module may bind a listener
ok   — a listener in tests.rs is caught when the parent ships it
ok   — pub mod form of the out-of-line cfg(test) declaration
ok   — a non-test cfg does not exempt the module
ok   — a listener in a plain production module is refused
ok   — the working tree passes the gate
```

Four of the six assert a **refusal**. A gate that only ever passes is indistinguishable from a gate that is not running, and this one's false negative is a server reaching the CLI.

Confirmed to bite: against the unfixed script, exactly the two acceptance cases fail and all four refusals still hold. So the tests fail for the reason claimed rather than passing for an unrelated one.


auser:
Pushed `9aebe8b5d`. CI proved the point these tests were written to make, on their own first run.

Three cases went red — all three refusal directions — because the gate accepted every probe. `ripgrep` is installed in the **Invariant** job, which runs the gate; it is not installed in **lint-core**, where I had put the test step. Without `rg` the single search the gate performs returns nothing, `|| true` makes that indistinguishable from a clean tree, and the script reports success having examined nothing.

So the gate has a second, larger hole than the one this PR started with: **it passes vacuously wherever ripgrep is absent.** A check that cannot run must not report success — the green tick gets read as evidence the thing was looked at. Fixed by requiring `rg` up front and exiting non-zero with the reason, and by moving the test step beside the gate it tests, in the job that has the tool.

One note on the seventh case, because its first draft was wrong in an instructive way. I originally ran the gate under `env -i PATH=/nonexistent`, and it passed against the *unfixed* script too — emptying PATH removes coreutils, so the gate died for an unrelated reason and my assertion went green while proving nothing about ripgrep. It now drops only ripgrep's own directory from PATH.

Against the unfixed script, exactly three cases fail: the two out-of-line acceptances and the ripgrep one. Against the fixed script, all seven pass.

PR #2619, opened 2026-08-17
perf(fc): stop narrating every boot across an emulated serial port

auser:
## The cost

Firecracker's `ttyS0` is an emulated 16550A — the guest takes **a VM exit per byte** written to it. A workload boot emitted ~19 KB of kernel log that way. On a host with speculative-execution mitigations active, each of those exits is expensive, and the two multiply.

## Measured

12 launches per arm on Linux/KVM, same binary, same rootfs, same guest kernel, **mitigations on both sides** — so this is the real-world figure:

| | before | after |
|---|---:|---:|
| dispatch p50 | 504.0 ms | **297.7 ms** |
| dispatch p95 | 916.8 ms | **328.2 ms** |
| `driver_boot` | 435.6 ms | **222.4 ms** |
| console volume | 19,095 B / 247 lines | **2,752 B / 24 lines** |

**206 ms off p50**, and the tail closed further than the median did.

For scale: a separate controlled test on the same box — toggling only the host's `mitigations=` parameter, with the host kernel pinned across the reboot so version could not confound it — moved dispatch 504.0 ms → 308.1 ms. Nearly all of that penalty was being paid *per console byte*, which is why removing the bytes recovers it while the mitigations stay on.

## What `quiet` does and does not do

It raises the console loglevel; it does not remove the console. Warnings and errors still print, and the guest agent's own messages are userspace writes rather than printk, so `console.log` keeps the lines an operator actually reads — 24 of them instead of 247.

`MVM_GUEST_CONSOLE_VERBOSE=1` restores full kernel output, because a hung guest is diagnosed from exactly that. The verbosity decision is a pure function over the env value, in the same shape as `tracing_enabled_from`, so it is tested without mutating process state.

## Why not virtio-console, like libkrun

**Firecracker has no virtio-console device.** I checked before proposing it as a follow-up:

| cmdline under FC 1.14.1 | output |
|---|---:|
| `console=hvc0` | 6,664 B — *identical to no console at all* |
| `console=ttyS0` | 20,520 B |

`hvc0` produces nothing because the device does not exist; FC's model is block/net/vsock/balloon/rng plus the 8250. libkrun and the mock boot `console=hvc0` and never pay per-byte exits. **The asymmetry between the backends is a capability Firecracker lacks, not a choice mvm made** — and there is no structural fix short of patching Firecracker upstream, so `quiet` is the fix rather than a stopgap.

## Tests

Two existing tests pinned the old cmdline and are updated. Four added:

- default cmdline is quiet across all three boot shapes (virtiofs-root, disk, initramfs)
- `quiet` is additive and leaves the boot-shape tokens other code parses alone
- the verbose escape hatch, over the raw env values including `0`/`no`/`TRUE`

## Verification

`cargo nextest run --workspace` — **11,994 passed, 0 failed**. `cargo +nightly fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/mvm-backends/src/driver/fc.rs
+++ b/crates/mvm-backends/src/driver/fc.rs
@@ -77,10 +77,47 @@ impl Default for FcDriver {
 /// UART), and carry NO `mvm.ip=` / `mvm.gw=` tokens — the converged path has no
 /// guest NIC. The root/init selection follows the boot shape; the shared cmdline
 /// assembler layers verity/grants/egress/uvols tokens on top of this.
+/// Set to `1`/`true` to restore full kernel console output on a Firecracker
+/// guest, at the cost documented on [`fc_console_verbosity`].
+pub const GUEST_CONSOLE_VERBOSE_ENV: &str = "MVM_GUEST_CONSOLE_VERBOSE";
+
+/// The console-verbosity token for a Firecracker boot, `" quiet"` by default.
+///
+/// Firecracker's `ttyS0` is an emulated 16550A: the guest takes a VM exit per
+/// byte written to it. Measured on a Linux/KVM host with speculative-execution
+/// mitigations active, a workload boot emits ~19 KB of kernel log that way and
+/// pays 148-176 ms for it — a third of the launch, spent narrating a boot
+/// nobody is reading. The other backends do not share the cost: libkrun and the
+/// mock boot `console=hvc0`, a virtio-console whose output rides a ring.
+///
+/// `quiet` raises the console loglevel so `KERN_INFO` chatter stops crossing
+/// the port. Warnings and errors still print, and the guest agent's own
+/// messages are userspace writes rather than printk, so they are unaffected —
+/// `console.log` keeps the lines an operator actually reads.
+///
+/// The escape hatch matters: kernel boot output is how a hung guest gets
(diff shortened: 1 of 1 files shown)


auser:
## Force-pushed: this branch was silently reverting #2595

Caught before merge. The branch carried an older copy of `fc.rs`, moved wholesale between worktrees, which undid #2595's one-line fix:

```
-        trace.write_driver_to(&state_dir);
+        trace.write_to(&state_dir);
```

That is the trace-clobber fix — reverting it would have re-armed the bug where the runner's trace write truncates the driver's, so `driver_boot` becomes unsplittable again. With auto-merge armed, it would have landed without anyone reading that line.

It surfaced by accident: I deployed this branch to a test host expecting `driver_boot` to decompose into `driver_boot.vmm_start` / `driver_boot.guest_boot`, and it came back opaque at 222.7 ms. Going after the missing instrumentation found that I had deleted it myself.

**Now fixed:** auto-merge disarmed, the line restored, and the branch rebased onto current `main` so the diff is exactly this change. Verified the diff touches one file and the only removed lines are the three `console=ttyS0` strings this PR intends to change.

`cargo nextest run --workspace` after the rebase: **12,016 passed, 0 failed.**

Re-arming auto-merge.

PR #2631, opened 2026-08-17
feat(cli): make `--help` describe the product the docs describe

auser:
## The problem

The published CLI reference documents `mvmctl run` as the one-shot flagship — *"like `docker run --rm` but with a Firecracker microVM as the sandbox"* — and `getting-started/happy-paths.md` puts it in the three-command path for the primary audience. In the code it carried `hide = true`, demoted in its own doc comment to an SDK transport.

It was not only `run`. **Eleven documented verbs were invisible in `--help`**, including:

- `secret` — the entry point to the whole host-side credential-substitution subsystem
- `trust` — which owns `trust receipt verify`, the verification half of the signed-receipt feature

and **seven visible verbs had no reference row at all**. A user typing `mvmctl --help` could not learn from the tool that either subsystem had a CLI. A verb you cannot discover is a verb you do not have.

## What changed

**ADR-027 amended.** `run` is a first-class visible top-level verb alongside `machine run`. The hidden/visible split is restated as three buckets:

| Bucket | Contents |
|---|---|
| Visible | anything a user is expected to invoke, grouped by `display_order` |
| Dev tooling | `seccomp-audit`, `storage`, `reconcile`, `dashboard`, `persistent-builder` |
| Internal transports | the `__`-prefixed subprocess plumbing |

Fifteen user-facing groups promoted out of `hide = true`; twelve missing reference rows written.

**A gate, so this cannot drift back.** `xtask check-cli-help-matches-docs` (Lint job) enforces both directions: a documented verb must be visible, a visible verb must be documented. Only *command evidence* counts as documentation — a table row or a fenced invocation — because the reference legitimately discusses retired verbs in prose and in one place explicitly negates one ("not by a public `mvmctl policy` command").

Hiding a verb to dodge writing its row also deletes it from `--help`. That price is what keeps the escape honest, and it is why the gate needs no allowlist.

**Also fixed:** the `mvmctl doctor` claim-10 failure string told the user to set `--network-preset`, a flag that does not exist, and cited ADR-002 for a claim that lives in ADR-001. Two comments on live paths named the same dead flag.

## The resulting surface

```
  machine     Beginner microVM workflows (run an OCI image and more)
  run         Run one command in a fresh transient microVM, then tear it down
  build       Build-time commands (image, compile, validate, kernel)
  ...
  secret      Manage local secret namespaces
  trust       Manage trusted bundle publishers
  deps        Inspect cached application dependencies
```

## Verification

- `cargo nextest run --workspace` — 12072 tests, green (one unrelated `mvm-http` parallelism flake, passes isolated)
- `cargo test --workspace --doc`, `just check-gated`, nightly `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`
- Doc/spec gates: `check-cli-help-matches-docs`, `check-machine-doc-guards`, `check-doc-claims`, `check-adr-coverage`, `check-claim-catalog`, `check-honesty`, `check-no-overclaim`, `check-plan-names`, `check-sprint-append`, `check-no-spec-refs-in-comments`, `check-workflow-paths`
- **Mutation-checked:** re-hiding `secret` and deleting the `pool` rows each produced exactly one finding; restoring them returned the gate to clean.

Three existing tests encoded the old policy and were rewritten to encode the new bucket rule rather than deleted — they are the in-crate counterpart to the gate.

## Scope

This is Phase A of the run-first CLI plan (`specs/plans/329-run-first-cli-and-upstream-adoption.md`, which also gains a corrections section: Phase 3's presets were already shipped, the two run surfaces diverged in *both* directions, Phase 2 must reuse the existing `mvm.toml` discovery rather than invent a config file, and Phase 6 is blocked behind a withdrawn ADR).

Consolidating `RunArgs` and `MachineRunArgs` onto one struct is the next phase and is deliberately not in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,6 +326,9 @@ jobs:
       - name: mvm-cli drives machines through the mvm-client facade
         run: cargo run -p xtask -- check-cli-runtime-surface
 
+      - name: "`mvmctl --help` and the CLI reference describe one surface"
+        run: cargo run -p xtask -- check-cli-help-matches-docs
+
       - name: repr(C) types carry a layout contract (check-abi-layout)
         run: cargo run -p xtask -- check-abi-layout
 
--- a/crates/mvm-cli/src/commands/mod.rs
+++ b/crates/mvm-cli/src/commands/mod.rs
@@ -151,11 +151,11 @@ pub(in crate::commands) enum Commands {
     /// Manage versioned packs (list/rollback/prune/download/update)
     #[command(display_order = 9)]
     Pack(pack::Args),
-    /// SDK transport surface (`run --mode live/plan`). Hidden: the user-facing
-    /// transient-run role folded into `machine run`; `run` survives only as the
-    /// SDK Sandbox launcher the Python/TS SDKs shell to, so it stays hidden
-    /// rather than break that transport.
-    #[command(hide = true)]
+    /// Run one command in a fresh transient microVM, then tear it down
+    ///
+    /// The argument surface still differs from `machine run` in both
+    /// directions; consolidating the two into one struct is the next step.
+    #[command(display_order = 2)]
     Run(vm::exec::RunArgs),
     /// Internal SDK host-dispatch transport for `MVM_NO_VM=1`.
     #[command(name = "__sdk-no-vm", hide = true)]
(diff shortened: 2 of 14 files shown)


auser:
Pushed `327d9ab49` from another session to clear the BDD lane — flagging since this is your branch.

`readme_contract.feature:31` asserted `run --help` contains the literal `run --mode live/plan`. That string came from the verb-level doc comment at `commands/mod.rs:154`, which this PR rewrites — correctly, since un-hiding `run` and describing it as the transient-run verb is the point of the change. The scenario was written when `run` was hidden and pinned the old prose verbatim, so it had an opinion about wording it should not have had.

What the scenario is genuinely for is untouched: the Python and TS SDKs shell to `run --mode live/plan`, so the flag and its meaning must survive any rewording of the verb above it. I kept both assertions that carry that guarantee — `--mode` and `SDK transport mode`, which still live on the `--mode` field's own doc comment in `exec.rs:195` — dropped the one that pinned the summary, and retitled the scenario to say what it protects.

Verified against the built binary rather than reasoned about: `run --help` matches `--mode` three times, `SDK transport mode` once, and `run --mode live/plan` zero times.

`Test` and `Test Linux and conformance` were red for the same assertion and should clear with it.

PR #2643, opened 2026-08-17
perf(launch): cut the sleep quantization out of teardown, and read the phases as a tree

auser:
## Why

`MVM_PHASE_TIMING=1 mvmctl machine run --image alpine` reported `total=287.2ms`. The half this repo actually gates — `dispatch_window`, 74.3ms — is 2.7x inside the published 200ms prepared-cold p50 and needed nothing. The other 213ms sits in teardown and admission, where no budget looks, and most of it turned out not to be work.

## What moved

**`stop_console_cleanup`: 57.0ms → 7-11ms** (measured, HVF/macOS, steady state). It was the largest single span in a transient teardown, and two more instances of the cadence bug Plan 299 catalogued as "the same cadence bug, three more times":

- The stream accept loop polled a *nonblocking* listener on a 25ms tick. Backoff — the fix used at the other four sites — cannot help here: the loop idles for the whole life of the VM, so it is always already sitting at the ceiling when the stop arrives. It needed an event, not a shorter tick. The listener now blocks in `accept` and `shutdown` wakes it by connecting to its own socket.
- `DurableSink::seal` polled `is_finished()` every 20ms because `std` has no timed join and the module cannot take a dependency to get one. The writer exits in microseconds once the job channel drops. It now waits on a channel whose sender the writer thread drops on exit — a timed join with no cadence.

Both swap a wait that could time out on its own for one that cannot, so both are bounded and both have a test for the pathological case. The accept wake-up cannot be delivered if something already unlinked the socket; that path detaches rather than hanging teardown.

**Admission.** The three pre-boot chain barriers now share one flush, closing the open acceptance item on #2293. `sync_data` is file-wide and none of the three acts on the one before it, so syncing each separately bought no ordering the single flush does not — and on macOS each is a real `F_FULLFSYNC` at **7.41ms p50** (measured; `fsync(2)` on the same host is 0.03ms, which is why this looked free in the original #2293 comparison).

`sync_policy_for`'s fail-closed default and `DEFERRABLE_EVENTS` are untouched. Batching is a scope a caller opts into, it fails closed, and it closes before the caller has an `AdmissionContext` to boot from.

The chain cursor also stops re-reading the whole segment on every append. The active segment grows to 4MiB before rotating and a launch appends ~6 entries, so admission was costing up to 24MiB of reads per launch — rising as the segment filled and dropping at each rotation, a sawtooth no sample would explain.

## Readability

`MVM_PHASE_TIMING=tree` renders the same spans nested by containment:

```
[mvm] launch cold — total 287.2ms, dispatch window 74.3ms (prepared-cold p50 budget 200ms across 20 samples)
  resolve                             8.7ms    3%
  drives                             13.4ms    5%
    artifact verify                   0.2ms
  admit                              47.4ms   17%
  ...
  teardown                           91.4ms   32%
    cleanup handoff                  91.3ms
      stop transient                 90.7ms
        stop driver kill             33.7ms
          stop pid disappearance     33.7ms
        stop console cleanup         57.0ms
```

The greppable pair is byte-identical and still the default — `scripts/check-hvf-warm-restore.sh` parses it, and a test pins that.

Nesting comes from a declared table asserted against the spans rather than read off the names, because the names mislead: `artifact_verify` is recorded during *drives*, not admit, and driver kill and console cleanup are *siblings*. Both are what a plausible guess gets wrong — the invariant test was verified to go red on exactly that error. The budget prints as context with no verdict: one launch is one sample, the published number is a p50 over twenty, and `bench/doc_table` already warns about conflating the two.

`warm-claim-phase` reported cumulative time from boot start under span-shaped names, so the reap looked free and the claim absorbed its cost. It reports deltas now.

## Not changed, and now named

**The receipt write is the dominant admit cost — ~36ms of ~45ms.** The emit trace shows it plainly. `emit_receipt_for_entry` is explicitly best-effort and its own doc calls receipts "a derived cache [that] must never block the primary audit emit", but `append_chained` writes the body through `write_atomic` → `sync_all` → `F_FULLFSYNC`, synchronously, before boot — while the head beside it already uses `write_atomic_unsynced`. Fixing it is a durability decision about claim-8 evidence export rather than redundancy removal, so it is written up in Plan 299 with three options and left for a maintainer.

Also left alone deliberately: the readiness poll cap. Plan 299 Phase 5 already took `guest_kernel_entry` 53.8 → 18.0ms with backoff and notes that tightening *widened the p99 tail* because the flat tick had been hiding real variance. Its remaining task there is event-driven readiness, not a smaller constant.

## Verification

- `cargo nextest run --workspace` — **12,093 passed**
- `cargo test --workspace --doc`, `cargo +nightly fmt --all -- --check`, `cargo clippy --workspace --all-targets -D warnings`, `just check-gated` — clean
- xtask: `check-claim-catalog`, `check-honesty`, `check-no-overclaim`, `check-doc-claims`, `check-witness-citations`, `check-nextest-groups`, `check-single-home`, `check-test-home-isolation`, `check-no-spec-refs-in-comments`, `check-plan-names`, `check-sprint-append` — all OK
- Live HVF runs before/after for the console-seal number; `mvmctl trust audit verify` clean across batched-emit runs; `a_batch_chain_still_verifies_end_to_end` pins the chain itself

### DIFF
--- a/crates/mvm-cli/src/commands/vm/phase_timing.rs
+++ b/crates/mvm-cli/src/commands/vm/phase_timing.rs
@@ -9,6 +9,8 @@
 
 use std::time::Instant;
 
+use crate::bench::cold_launch::MIN_MATRIX_SAMPLES;
+use mvm_core::launch_trace::PhaseTimingMode;
 use mvm_runtime::workload_runner::StopTiming;
 use serde::{Deserialize, Serialize};
 
@@ -318,6 +320,112 @@ impl RunPhaseMarks {
 /// as warm-launch latency.
 pub const WARM_START_MAX_MS: f64 = 300.0;
 
+/// Which span contains which, as `(span, parent)`.
+///
+/// Parents named here are either another sub-span or one of the coarse bucket
+/// labels [`RunPhaseTimings::coarse_rows`] emits. This table is the single
+/// place the report's shape is stated, and it is asserted against the spans
+/// themselves by `every_parent_covers_its_children` — a row placed under the
+/// wrong parent renders a plausible, wrong partition, and the arithmetic is
+/// the only thing that catches it.
+///
+/// The non-obvious entries: `artifact_verify` is recorded while drives are
+/// being prepared, not during admission; and `stop_driver_kill` and
+/// `stop_console_cleanup` are siblings under `stop_transient`, not nested —
+/// stopping the console happens after the driver is killed, not inside it.
+const SPAN_PARENTS: &[(&str, &str)] = &[
+    ("mount_fingerprint", "drives"),
+    ("mount_cache_lookup", "drives"),
+    ("mount_materialize", "drives"),
+    ("artifact_verify", "drives"),
+    ("vmm_create", "backend start"),
+    ("guest_kernel_entry", "guest ready"),
+    ("agent_auth", "guest ready"),
+    ("first_dispatch", "command"),
+    ("cleanup_handoff", "teardown"),
+    ("stop_transient", "cleanup_handoff"),
+    ("pool_replenish", "cleanup_handoff"),
+    ("state_remove", "cleanup_handoff"),
+    ("stop_attach", "stop_transient"),
+    ("stop_endpoint_reaping", "stop_transient"),
+    ("stop_driver_kill", "stop_transient"),
+    ("stop_console_cleanup", "stop_transient"),
+    ("stop_supervisor_signal", "stop_driver_kill"),
+    ("stop_pid_disappearance", "stop_driver_kill"),
+    ("stop_force_kill_wait", "stop_driver_kill"),
+    ("stop_state_cleanup", "stop_driver_kill"),
+];
+
+/// One rendered row, before the label column has been sized.
+struct TreeRow {
+    /// Indent already applied, so sizing is one `max` over these.
+    label: String,
+    ms: f64,
+    share: Option<f64>,
+}
+
+/// One coarse bucket, with its share of the launch.
+struct CoarseRow {
+    label: &'static str,
+    ms: f64,
+    share: f64,
+}
+
+impl CoarseRow {
+    fn new(label: &'static str, ms: f64, total_ms: f64) -> Self {
+        let share = if total_ms > 0.0 {
+            ms / total_ms * 100.0
+        } else {
+            0.0
+        };
+        Self { label, ms, share }
+    }
+}
+
+/// Size the label column to the widest row, then emit.
+///
+/// Two passes rather than a fixed column because the deepest labels
+/// (`stop supervisor signal` at four levels of indent) are more than twice the
+/// width of the shallowest, and any constant wide enough for them wastes a
+/// screen of space on every other row.
+fn render_rows(rows: &[TreeRow]) -> String {
+    let width = rows.iter().map(|r| r.label.len()).max().unwrap_or(0);
+    rows.iter()
+        .map(|row| {
+            let value = format!("{:.1}ms", row.ms);
+            let mut line = format!("{:<width$}  {:>9}", row.label, value);
+            if let Some(share) = row.share {
+                line.push_str(&format!("  {share:>3.0}%"));
+            }
+            line.trim_end().to_string()
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+/// Append every recorded descendant of `parent`, depth-first, in the order
+/// [`LaunchSubTimings::recorded`] declares them.
+fn push_children(
+    rows: &mut Vec<TreeRow>,
+    parent: &str,
+    depth: usize,
+    recorded: &[(&'static str, f64)],
+) {
+    for (name, ms) in recorded {
+        if SPAN_PARENTS
+            .iter()
+            .any(|(span, span_parent)| span == name && *span_parent == parent)
+        {
+            rows.push(TreeRow {
+                label: format!("{}{}", "  ".repeat(depth), name.replace('_', " ")),
+                ms: *ms,
+                share: None,
+            });
+            push_children(rows, name, depth + 1, recorded);
+        }
+    }
+}
+
 impl RunPhaseTimings {
     /// Admitted-plan to command-dispatch: the window the warm-start SLO is set
     /// against (backend claim/restore + boot-to-agent wait).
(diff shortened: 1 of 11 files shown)


auser:
Diagnosis for the BDD failure, from another session — I did the analysis before noticing you are active in this worktree, so I have not pushed anything.

## The failure

```
✘  And a workload plan was admitted in the audit chain
   Step panicked. Captured output: Cannot drop a runtime in a context where
   blocking is not allowed. This happens when a runtime is dropped from
   within an asynchronous context.
```

Two scenarios in `s6_admission_audit/signed_execution_plan.feature`, and `Test Linux and conformance` is red for the same reason.

## Cause

The `OnceLock<tokio::runtime::Runtime>` on `AuditEmitter` is the right call for the cost you are removing, but it moves *when the runtime is dropped*.

Before, `emit_entry_blocking` built and dropped a runtime inside its own synchronous call — wasteful, but the drop always landed in a sync context. Now the runtime lives on the emitter and dies whenever the emitter does. `admit_plan_to_audit_chain` (`crates/mvm-conformance/tests/steps/admission_audit.rs:29`) constructs an `AuditEmitter` in a cucumber step, and cucumber runs steps inside its async runtime — so the emitter drops there, and dropping a runtime from async is a panic by design.

Nothing in the production path changed behaviour; the drop site did.

## Suggested fix

`Drop` that shuts the runtime down without blocking:

```rust
impl Drop for AuditEmitter {
    fn drop(&mut self) {
        // The runtime outlives the calls that use it, so it dies with the
        // emitter — which may be inside someone else's async context, where
        // a blocking drop panics. `shutdown_background` never blocks.
        if let Some(rt) = self.runtime.take() {
            rt.shutdown_background();
        }
    }
}
```

`OnceLock::take` needs `&mut self`, which `Drop` has, and `shutdown_background` consumes the runtime and returns immediately.

Worth a test that constructs and drops an `AuditEmitter` inside a `#[tokio::test]`, since that is precisely the shape no unit test covered and the one the BDD suite exercises by accident.

I have not verified the fix — it is your branch and you are in it. Happy to take it if you would rather hand it over.

auser:
Pushed the fix I described above — the branch had not taken it and your last push was docs, so I applied it rather than leave four lanes red.

`impl Drop for AuditEmitter` calling `shutdown_background()`. All four failures were the same tokio line (`runtime/blocking/shutdown.rs:51`): the two `mvm-client` launch tests, the BDD admission scenarios, and `Test Linux and conformance`.

The witness is `an_emitter_dropped_inside_an_async_context_does_not_panic` — construct, emit and drop inside `#[tokio::test]`. It emits **first** deliberately: the runtime is built lazily, so an emitter that never emitted drops an unfilled `OnceLock` and passes no matter what `Drop` does. Verified it fails at the same tokio line without the impl, and that the two originally-red `launch::tests` now pass.

Two notes, both worth your eye since this is your change:

**The caching itself is right** — it removes a real per-entry cost on the admission path. The bug was only that it relocated the drop from inside the emit to wherever the emitter dies, and one of those places is async.

**A first `nextest` run showed 4 unrelated failures a re-run did not reproduce** (2108/2108 clean). This crate pair flakes under parallel load, so I am reporting the first run rather than only the green one — if you see the same four, they are probably not yours either.

auser:
Pushed a fix and the Phase 6 measurement follow-up.

**CI fix.** All four red lanes were one cause, mine: caching the audit emit runtime on `AuditEmitter` moved *where it is dropped*. Built per entry it was also dropped there, on a thread the caller has already guaranteed has no ambient runtime — which is what made the drop legal. Cached, it drops whenever the emitter does, and an emitter held across an async boundary drops it inside tokio's blocking-pool shutdown: `Cannot drop a runtime in a context where blocking is not allowed`. The claim-8 admission-audit BDD scenario is what caught it.

Reverted rather than worked around: the prize was 0.03 ms of construction per entry next to the ~7.4 ms `F_FULLFSYNC` on the same path, so it was never the cost. Verified by reproducing the CI test-support lane locally (3,593 + 63 tests, exit 0) and `just bdd` back to 205 passed / 0 failed from 203 / 2.

Worth noting for the next person: none of this appears in `cargo nextest run --workspace`. Those tests are behind the `test-support` feature and the BDD harness needs `--features bdd`.

**Phase 6 measurement.** `shutdown-timing.json` was supposed to decide between the 5 ms watchdog tick and the console rewrite. It excludes both — its spans sum to **2.9 ms** against a host-measured 33–37 ms, because `pid_disappearance` waits on kqueue *process exit* while the record stops being written just before the PID file is removed.

Varying only guest memory shows what the rest is:

| `--memory` | `stop_pid_disappearance` |
|---|---:|
| 512M (default) | 33.1, 37.1 ms |
| 1G | 76.1, 55.8 ms |
| 2G | 92.0, 92.7 ms |
| 4G | 193.0, 195.6 ms |

Linear at ~48 ms/GB, with `hv_vm_destroy` already accounted for at 346 µs — it is the guest address space being reclaimed at process exit. So the watchdog self-pipe is not worth building (at most 15% of the span at the default size), and a 4 GiB workload pays ~195 ms of foreground teardown that no `alpine`-sized benchmark can see.

Two instrumentation gaps recorded while doing it: `stop_observed_at` is stamped *after* the watchdog's sleep, so the tick latency the file exists to expose is the one thing it cannot measure; and `vm_destroy` is hardcoded `Duration::ZERO` at one construction site.

**Release before/after**, now that a like-for-like binary exists — `stop_pid_disappearance` is the control, unchanged by this PR:

| | before | after |
|---|---:|---:|
| `stop_console_cleanup` | 57.0 ms | **7.0–7.7 ms** |
| `stop_pid_disappearance` (control) | 33.7 ms | 33.4–35.5 ms |
| teardown | 91.4 ms | **41–45 ms** |
| total | 287.2 ms | **~233 ms** |

auser:
Correction to my previous comment: it was posted before I saw that a concurrent session had already fixed this differently, and its first paragraph is wrong.

I did **not** revert the runtime caching. `114996789` keeps it and adds a `Drop` impl calling `shutdown_background()`, which is the better fix — it keeps the change this branch is about and, unlike my revert, ships the regression witness: `an_emitter_dropped_inside_an_async_context_does_not_panic`, which constructs, emits and drops an emitter inside `#[tokio::test]`. I discarded my own commit in favour of it.

I verified that fix rather than assuming it: 35 `audit::emitter` tests and all 21 `mvm-client launch::tests` under `--features test-support` pass, and removing the `Drop` impl makes the new witness fail at exactly the `tokio/runtime/blocking/shutdown.rs:51` line CI reported. It emits before dropping on purpose — the runtime is built lazily, so an emitter that never emitted would drop an unfilled `OnceLock` and pass no matter what `Drop` does.

Everything else in that comment stands: the root-cause analysis, the guest-RAM teardown measurement, and the release before/after numbers.

PR #2644, opened 2026-08-17
perf(build): key mvm-cli's nested binaries by content, not by profile

auser:
`cargo build` sits on two progress lines, `mvm-cli(build)` and `mvmctl(bin)`, and the reason is one build script.

## Measured

Fresh worktree, aarch64 macOS, `[build] jobs = 6`.

| Workload | Before | After |
| --- | --- | --- |
| Cold build, fresh worktree | 359s | **45.5s** |
| ↳ `mvm-cli` build script within it | **332.7s (93%)** | **0.4s** |
| `--release`, empty nested target dir — script only | ≥163s musl leg, always | **0.6s** |
| Edit `mvm-cli/src` | re-ran both nested legs | **7.7s**, script not re-run |
| Edit `mvm-core`, warm worktree | ~18s (script 14.3s) | 15.5s |
| Workspace `clippy --all-targets` | spawned 4–6 nested `cargo build`s | 6.0s warm |

The script starts at t=16.4s and runs to t=349s; every other crate finishes underneath it. `mvm-cli` the crate is 9.1s and cannot start until it ends.

## Why it was slow

`crates/mvm-cli/build.rs` runs two nested builds and neither could tell a fresh artifact from a stale one:

- **musl leg** (~163s) reused when `PROFILE == "debug"` **and** `prebuilt.is_file()`. That proves nothing about the bytes, so it had to refuse every other profile — `--release`, `release-witness` and each `--target` paid full price into their own nested target dir.
- **per-VM helper leg** had no reuse at all, and cargo runs build scripts for `check` and `clippy`, so a lint pass spawned 4–6 real linking `cargo build`s of the `mvm-hostd` closure.

Nothing was shared across worktrees, profiles or triples — `shared_nested_target_dir` shares only across feature fingerprints inside one `target/<profile>/`.

## What changed

Key both legs on content: each binary's real dependency closure (walked from the manifest graph, not a hand list), plus `Cargo.lock`, `rust-toolchain.toml`, the pinned toolchain, target, features, and a flavour separating a musl static from a native host build. A hit proves the artifact is what this tree produces, which is what makes reuse sound outside dev and safe to share — so the store lives at `~/.cache/mvm/embed/`, outside `target/`, with atomic publish.

Deriving the closure also narrows the watch set: `mvm-cli`'s 251 files are downstream of these binaries and no longer re-run either leg.

**The dev trade is preserved on purpose.** A miss means the sources really changed and rebuilding costs ~163s, so dev still reuses a stale embedded binary rather than regress the inner loop. The difference is the key now *knows* it is stale and says so through `cargo:warning=` — the one channel cargo surfaces without `-vv`. (`MVM_EMBEDDED_BINS_REUSED`, which this replaces, has never had a reader.) `MVM_EMBED_NO_CACHE=1` opts out of both.

A store hit also seeds the nested target dir the fallback reads from; without it a store-seeded worktree had nothing to fall back to and its first edit paid a full cross-compile — moving the cost rather than removing it.

Also makes `scripts/dev-env.sh` authoritative. Its `${VAR:-default}` exports let a value inherited from another worktree win, so two source trees shared one `CARGO_TARGET_DIR`; cargo fingerprints embed absolute paths, so alternating recompiled the workspace and concurrent builds serialized on one lock, silently. `AGENTS.md` already documented the intended behaviour.

## Verification

- `cargo nextest run --workspace`: **12,100 passed**, 0 failed
- `cargo test --workspace --doc`: passed
- `cargo clippy --workspace --all-targets -- -D warnings`, nightly `cargo fmt --all --check`: clean
- Gates: `check-single-home`, `check-closure-budget`, `check-duplicate-majors`, `check-workspace-dep-inheritance`, `check-core-runtime-free`, `check-mvm-host-binaries-sync`, `check-no-spec-refs-in-comments`, `check-plan-names`, `check-sprint-append`, `check-file-size`
- 31 unit tests for the store, including that the key moves for every input, that field boundaries cannot be shifted to forge a collision, and that `mvm-cli` is absent from `mvm-build`'s closure when walked against the real manifests

## Known gaps, recorded in the plan

- The store is **unbounded** — 424 MB / 48 entries after a day, and `cargo clean`, `just clean-dev-state` and `mvmctl cache prune` all miss it by design. Needs an LRU bound before wide use.
- In a store-seeded fresh worktree the first `mvm-core` edit is ~52s: the aux leg has no stale fallback by design (a stale supervisor silently produces a guest that ignores your edit) and its nested target dir is cold there.

Does not re-litigate `specs/plans/334-build-critical-path.md` — its five refuted hypotheses stand. Its baseline was a warm `cargo build -p mvm-cli`; the cold and per-profile costs here were outside it. Supersedes the freshness half of `specs/plans/2026-08-15-aux-helper-binary-freshness.md`.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -219,9 +219,23 @@ test-doc:
 # scoped to this recipe instead of being set globally.
 #
 # A content cache, not a build lock, so parallel sessions don't serialize on it.
-# Needs `cargo install sccache`. For cross-worktree hits the host config also
-# needs `basedirs`; without it every worktree is a private cache and the
-# measured hit rate is 0%.
+# Needs `cargo install sccache`.
+#
+# Do not expect this to dedupe across worktrees. `basedirs` in the host config
+# is necessary for that but is NOT sufficient: sccache's key covers the
+# rustc command line, which carries the target directory's full path, and every
+# worktree has its own. Measured on this host with `basedirs` correctly set,
+# building one crate three ways:
+#
+#   cold, target dir A                       2 hits / 100 misses
+#   target dir A wiped and rebuilt          70 hits /  79 misses
+#   identical source, target dir B           0 hits / 152 misses
+#
+# So it pays for re-populating one checkout after `cargo clean`, and pays
+# nothing for the second checkout — which is why the machine-wide Rust hit rate
+# sits near 2.5% across ~35k compiles rather than the 84% a same-path
+# experiment suggests. Cargo already caches deps within a target dir, so the
+# remaining value here is narrow.
 test-cached:
     @command -v sccache >/dev/null || { echo "sccache not found — install with: cargo install sccache"; exit 1; }
     RUSTC_WRAPPER=sccache CARGO_INCREMENTAL=0 cargo nextest run --workspace
--- a/crates/mvm-cli/build.rs
+++ b/crates/mvm-cli/build.rs
@@ -1,5 +1,7 @@
 #[path = "build_aux_helpers.rs"]
 mod build_aux_helpers;
+#[path = "build_embed_cache.rs"]
+mod build_embed_cache;
 #[path = "src/host_binaries/toolchain.rs"]
 mod host_binaries_toolchain;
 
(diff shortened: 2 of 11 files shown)


auser:
Pushed `60a02f5a0` from another session — your `Invariant` lane was red on `check-declared-backing` and the worktree looked idle.

`specs/plans/2026-08-17-embedded-binary-content-store.md` had no `Backing:` / `Validation:` header, which the gate requires of claim-bearing prose — and this plan asserts measured timings and two completed phases.

I declared `shipped-source` rather than `preview`, because phases 1–2 are in this branch with tests behind them (the `closure_*` unit tests and `tests/dev_env_isolation.rs`), so the assertive vocabulary is earned rather than borrowed. The validation line names which phases that covers and says plainly that 3–5 claim nothing — one header has to speak for a document describing both shipped and unshipped work, and `preview` would have understated the first half while `shipped-source` alone overstates the second.

Change it if you would rather it read differently; it is your plan and the wording is a judgement call, not a gate requirement. `check-declared-backing` is clean either way as long as the header is present and the validation resolves — mine is a sentence, which the gate deliberately does not try to resolve.

PR #2646, opened 2026-08-17
refactor(deps): declare three crate edges where they are actually used

auser:
Three manifests declared a runtime dependency that no library code reaches.

**This is not a build-time change.** Measured closure is unchanged — 239 (linux) / 230 (darwin) — because each crate stays in the graph via other edges. I originally picked these up as a "dead edges" performance item and they are not; reporting them as hygiene instead.

What it does buy: the next person trying to remove one of those *other* edges won't find the crate still pulled in for a reason that turns out not to exist.

| Crate | Declared | Actually reached from |
| --- | --- | --- |
| `mvm-runtime` | `libkrun-sys` | `tests/phase3c_supervisor_dispatch.rs` only → dev-dependency |
| `mvm-cli` | `mvm-net` | `#[cfg(test)] mod tests` in `commands/machine` → dev-dependency |
| `mvm-hostd` | `mvm-fs` | `mvm-apfs-snapshot-helper` bin only, itself `required-features = ["trusted-apfs"]` + macOS-gated → optional, enabled by that feature |

## Two blind spots this walked into, both documented in CLAUDE.md

- The first attempt **deleted** `libkrun-sys` from `mvm-runtime` outright, on the strength of a grep over `src/`. `cargo check --workspace --all-targets` failed with `unresolved import libkrun_sys` — the use is in `tests/`. That is the case `--all-targets` exists to catch.
- `--all-targets` in turn **cannot** see `required-features` targets, so it never builds `mvm-apfs-snapshot-helper`. Verified separately with `cargo check -p mvm-hostd --features trusted-apfs --bin mvm-apfs-snapshot-helper`.

## Verification

- `cargo nextest run --workspace`: **12,069 passed**, 0 failed
- `cargo check --workspace --all-targets`: clean
- nightly `cargo fmt --all --check`: clean
- Gates: `check-closure-budget`, `check-feature-closure-budget`, `check-workspace-dep-inheritance`, `check-duplicate-majors`, `check-core-runtime-free`


auser:
`Test workspace` was not your change — re-run triggered.

The job ran 1h0m14s and was cancelled at the timeout, but it never reached the tests. Step-by-step status shows it hung on **step 5, `Install system build deps`** — the `apt-get update && apt-get install` — with everything after it skipped:

```
4  success   Install Rust toolchain
5  cancelled Install system build deps      <-- 56 minutes here
6  skipped   Run ./.github/actions/install-zigbuild
7  skipped   Run ./.github/actions/rust-cache
9  skipped   Build
12 skipped   Run tests
```

So it never built anything, let alone ran a test. An apt mirror stalling on a runner is a transient; there is nothing in `refactor/crate-edge-honesty` that could produce it, and the aggregate `Test` lane is red only because it reports this one.

Re-ran the whole run rather than the single job — a single-job rerun in this repo materializes the lanes the scope filter had skipped, which turns a clean transient into a confusing partial result.

PR #2649, opened 2026-08-17
perf(agent): stop charging every exec a 50ms poll tick to notice a child that already exited

auser:
Stacked on #2643. Base is that branch so this diff is exactly its own two commits; it will be rebased onto `main` and retargeted once #2643 squash-merges.

## The finding

`command` — the span from an authenticated guest agent to a finished guest command — was **52.5 ms** for `/bin/true`, and constant to a tenth of a millisecond across runs: 52.5, 52.7, 52.9. That constancy is the tell. A number that steady is a floor, not work.

`crates/mvm-agentd/src/exec_stream.rs:191` slept a flat 50 ms whenever `try_wait` returned `None`. A command that exits in microseconds always loses that first race — the child cannot have been reaped by the time the loop reaches it — so the second check found it already gone, having waited the entire interval to look. **Every exec paid one full tick, and nothing could finish this span in under 50 ms.**

This is the fifth site of the cadence bug this work has now found, and the first one *inside the guest* rather than on the host — which is why Plan 299's four fixes and the two seal fixes in #2643 never came near it.

## The change

It now shares `mvm_core::poll_backoff` with the other four sites, rather than being a fifth copy of a constant. The idle counter resets on any output, so a chatty command keeps draining at the fast end of the backoff instead of decaying to the ceiling and staying there — this loop drains stdout/stderr on each pass, so the ceiling bounds streaming latency as well as exit latency.

Measured, release, HVF/macOS: **`command` 52.5 → 4.6–5.4 ms.**

## Where prepared cold lands

Release, HVF/macOS, `machine run --image alpine -- true`, warm artifact cache, first (cold) sample discarded:

| span | before this work | after |
|---|---:|---:|
| `admit` | 47.4 | 36–40 |
| `vsock_wait` | 58.2 | 61–70 |
| `command` | 52.0 | **4.6–5.4** |
| `teardown` | 91.4 | **39–48** |
| **total** | **287.2** | **p50 191.2, max 196.2** |

**Read the caveats with the table.** This is **9 samples, not the 20** a publishable lane requires, so there is no p95/p99 and these are not publishable numbers — the host became too loaded to finish the set. `vsock_wait` drifted slightly *up* and is now the largest remaining span; it is Phase 5's event-driven-readiness task and is deliberately untouched, since Plan 299 records that tightening the poll cap widens the p99 tail. And the `teardown` figure is a 512 MiB best case, because ~33 ms of it is guest-RAM reclaim that scales at ~48 ms/GB.

`dispatch_window` is 79–85 ms against its 200 ms p50 budget and is not materially changed here — every cut is *outside* the window the contract measures, which is the point: the contract was already being met while the user-visible number was 287 ms.

## On test coverage

No unit test pins the latency. The honest guard would be a wall-clock budget, and this repo has been bitten by those failing for reasons unrelated to what they claim to measure. The 21 existing `exec_stream` tests cover the behaviour that must not change — ordering, tail bytes, timeout, output cap, exit codes — and all pass; the number itself is evidenced by the before/after recorded in Plan 299.

## Verification

- `cargo nextest run -p mvm-agentd` exec_stream + child_wait: 21 passed
- Full-workspace clippy via the pre-commit hook: clean
- `cargo +nightly fmt --all -- --check`: clean
- xtask `check-honesty`, `check-no-overclaim`, `check-doc-claims`, `check-plan-names`, `check-sprint-append`: all OK

### DIFF
--- a/crates/mvm-agentd/src/exec_stream.rs
+++ b/crates/mvm-agentd/src/exec_stream.rs
@@ -17,7 +17,6 @@ use std::time::{Duration, Instant};
 /// (was `MAX_EXEC_OUTPUT` in the agent bin).
 pub const MAX_EXEC_OUTPUT: usize = 1024 * 1024;
 const DRAIN_BUF: usize = 4096;
-const POLL_INTERVAL: Duration = Duration::from_millis(50);
 
 fn spawn_drain<R: Read + Send + 'static>(
     mut reader: R,
@@ -143,10 +142,18 @@ pub fn stream_exec<F: FnMut(ExecEvent)>(
     let mut sent_out = 0usize;
     let mut sent_err = 0usize;
     let deadline = timeout_secs.map(|s| Instant::now() + Duration::from_secs(s));
+    // Consecutive polls that found neither output nor an exit. Reset on any
+    // progress so a chatty command keeps draining at the fast end of the
+    // backoff instead of decaying to the ceiling and staying there.
+    let mut idle_polls = 0u32;
 
     loop {
+        let before = (sent_out, sent_err);
         let capped = drain_into(&out_buf, &mut sent_out, true, &mut emit)
             | drain_into(&err_buf, &mut sent_err, false, &mut emit);
+        if (sent_out, sent_err) != before {
+            idle_polls = 0;
+        }
         if capped {
             kill_pgroup(&child);
             let _ = crate::child_wait::wait(&mut child);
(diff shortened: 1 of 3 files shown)


auser:
Rebased onto `main` now that #2643 has landed (squash `8732259cb`), and retargeted from that branch to `main`.

Verified the base by content rather than by PR state, since a squash-merge leaves the original commits unreachable: `render_tree`, `AuditEmitter::batched` and `ACCEPT_JOIN_TIMEOUT` are all present on `main`, and `exec_stream.rs` on `main` still has no `poll_backoff` — so this PR is the only thing carrying the exec fix.

The rebase applied clean. Diff against `main` is now exactly the two commits and three files this PR is about:

```
crates/mvm-agentd/src/exec_stream.rs       | 25 ++++-
specs/REFACTOR-STATUS.md                   |  7 +-
specs/plans/299-cold-launch-performance.md | 54 +++++++++
```

Also checked that the Plan 299 doc kept both sides — the four sections that came in with #2643 (seal cadence, the five `F_FULLFSYNC`s, guest-RAM teardown, the receipt cost) and the two this PR adds are all present, none overwritten.

PR #2653, opened 2026-08-18
fix(release): make the boot gate assert what a bare image can actually prove

auser:
`boot-image/v0.1.1` refused the x86_64 image again. The refusal was mine, not the artifact's.

## What actually happened

The kernel loaded (the ELF fix works), reached userspace, exec'd `/init` — and then:

```
```

The guest agent ships in the **runtime overlay** — a separate verity-sealed disk mounted at `/mvm/runtime` per ADR-018. A bare kernel+rootfs boot has no overlay, so `/init` finds no agent and exits **by design**. The gate was asserting `MVM_RUNTIME_BOOT_READY=guest-agent` against an image that structurally cannot satisfy it.

**It was failing every correct image.** That is the worst kind of gate: it blocks releases and teaches everyone to distrust it.

## How it got there

I copied the readiness signal from the `boot-latency` lane. That lane is `workflow_dispatch`-only and pinned to `v0.17.0` — whose kernel was a bzImage that could not load — **so the assertion had never once been proven to hold.** Copying a witness that has never gone green is the same trap as trusting one that has never gone red, and I walked into it.

## What the gate proves now

`start-return` — Firecracker accepts the artifact and the kernel loads and starts.

| Historical defect | Caught by |
|---|---|
| bzImage published as `vmlinux` (fails `InstanceStart`) | **this gate** |
| `/init` shebang at byte 1 (ENOEXEC) | the static shebang assert, one step earlier |
| anything past kernel handoff | **nothing yet** — needs the overlay |

Stated in the workflow comment rather than left for someone to infer, including the part it does *not* prove.

## Not strengthening it here

Attaching the overlay is the real fix and the overlay is already built in the same run. Filed separately. The last two attempts at this gate were each wrong in a way only a real release surfaced; a third guess without evidence is not an improvement.

## Verification

16 tests in `release_assets.rs`. The new guard asserts `start-return` is set and `guest-agent` is not — restoring `guest-agent` fails it, verified then reverted byte-identically. `actionlint` clean.

### DIFF
--- a/.github/workflows/release-boot-image.yml
+++ b/.github/workflows/release-boot-image.yml
@@ -498,17 +498,30 @@ jobs:
         # Boots the artifact in `staging/`, never a published one — the whole
         # point is to refuse the upload rather than diagnose it afterwards.
         #
-        # This is not a latency assertion. The budget is an order of magnitude
-        # above the observed median and RUNS/CONCURRENT are 1, because the
-        # question here is "does it reach userspace". Latency belongs to the
-        # `boot-latency` lane, which has its own threshold and its own argument.
+        # `start-return`, not `guest-agent`, and the difference is the honest
+        # limit of this gate. The guest agent ships in the runtime overlay, a
+        # separate verity-sealed disk mounted at /mvm/runtime; a bare
+        # kernel+rootfs boot has no overlay, so /init finds no agent and exits
+        # by design. Asserting agent-readiness here would fail every correct
+        # image. Attaching the overlay — it is built in this same run — is the
+        # way to strengthen this, and is tracked separately.
+        #
+        # So what this proves is that Firecracker accepts the artifact and the
+        # kernel loads and starts. That is the class that shipped as a bzImage
+        # named vmlinux, which fails InstanceStart outright. What it does NOT
+        # prove is that userspace works: the /init defect that shipped
+        # alongside it is caught by the static shebang assert above, not here.
+        #
+        # Not a latency assertion either. The budget is an order of magnitude
+        # above the observed median and RUNS/CONCURRENT are 1. Latency belongs
+        # to the `boot-latency` lane, which has its own threshold and argument.
         env:
           ARCH: ${{ matrix.arch }}
           MVM_RUNTIME_BOOT_BENCH: "1"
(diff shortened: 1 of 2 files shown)


auser:
Closing as superseded by #2662, which solved the same problem from the other direction — and landing this now would regress the gate.

## The two answers

This PR's premise, stated in its own comment:

> The guest agent ships in the runtime overlay [...] a bare kernel+rootfs boot has no overlay, so `/init` finds no agent and exits by design. Asserting agent-readiness here would fail every correct image. **Attaching the overlay — it is built in this same run — is the way to strengthen this, and is tracked separately.**

#2662 did exactly that: it attaches the overlay in the boot gate and keeps `MVM_RUNTIME_BOOT_READY: guest-agent`. So the workaround is no longer needed, and the thing this PR named as the better fix is now on `main`.

## Why it cannot merge as-is

The conflict is not incidental. This branch sets `start-return`; `main` now sets `guest-agent` with an overlay attached. Taking this side would trade an assertion that userspace works for one that only proves Firecracker accepted the artifact — which this PR is explicit about:

> what this proves is that Firecracker accepts the artifact and the kernel loads and starts [...] What it does NOT prove is that userspace works

And  would *pin* that weaker signal, so a future attempt to strengthen the gate would fail a test asserting it must stay weak. A correct premise at the time it was written becomes a lock on the wrong behaviour once the premise changes.

## What was right here

The diagnosis. This PR identified that the agent lives in the overlay and that a bare kernel+rootfs boot cannot reach agent-readiness — the same root cause #2660 reached independently from the failing release run. The disagreement was only about which side to move: weaken the assertion, or supply the missing disk.

Branch left in place. If any of the surrounding comment prose is worth keeping — the bzImage-named-vmlinux distinction in particular is well put and is not on `main` — it would be welcome as a docs-only follow-up.

auser:
Correction to the comment above — a shell quoting slip on my end dropped a test name, leaving "And  would *pin* that weaker signal". The sentence should read:

And `the_boot_gate_must_not_assert_a_readiness_signal_the_artifact_cannot_give` would *pin* that weaker signal, so a future attempt to strengthen the gate would fail a test asserting it must stay weak.

That test is the specific reason this cannot simply be rebased: it does not merely permit `start-return`, it requires it.

PR #2666, opened 2026-08-18
fix(release): boot the image the way it ships — with its runtime overlay

### LINKED ISSUE #2660: The staged x86_64 default-microvm panics: no guest agent in the image
With the toolchain fixes in (#2641, #2645), the x86_64 boot gate finally *runs* — and the staged `default-microvm` image does not boot.

## Evidence

`release-boot-image` run `32085503849`, tag `boot-image/v0.1.0` at `b9b3d935c` (current `main`), job `default-microvm (x86_64, ubuntu-latest)`, step **Boot the staged image before it becomes a release asset**:

```
mount: mounting devtmpfs on /dev failed: Resource busy
[    0.077258] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
[    0.078504] CPU: 0 UID: 0 PID: 1 Comm: init Not tainted 6.12.103 #1-NixOS
[    0.097330] Rebooting in 1 seconds..
```

The kernel loads and reaches userspace — so this is **not** the `/init` ENOEXEC of #2539, and not the ELF-vs-bzImage problem of #2594. It is one step further in: `/init` runs, finds no guest agent, and exits, and PID 1 exiting is a panic.

## The two lines that matter

`mvm-init: no guest agent resolved from /mvm/runtime and no baked fallback` — the staged image carries neither a runtime-overlay agent nor a baked one. Whatever composes the x86_64 `default-microvm` is not putting an agent in it.

`mount: mounting devtmpfs on /dev failed: Resource busy` precedes it and may be incidental (devtmpfs already mounted by the initramfs) or may be the reason `/mvm/runtime` is not where the agent is looked for. I have not established which.

## What this changes

The gate is working as designed, and the design is right: `publish-boot-image` refused with **`Assert every arch produced its full asset set`**, so nothing was published. Before #2645 this would have shipped a partial release — which is exactly how `boot-image/v0.1.0` came to exist with 20 assets and two failed jobs.

So the release pipeline is now honest. What it is honestly reporting is that the x86_64 image is unbootable.

## Blocks

- any `boot-image/v*` release, and therefore #2542, which pins the boot-latency lane to a published image
- #2539's remaining half

## Not established

- whether aarch64 has the same defect — its job failed earlier in the same run on `builder-vm-image`, so its `default-microvm` never got to the gate
- whether the agent is missing from the image or merely from where `mvm-init` looks
- whether a locally built image boots, which would separate "composition is broken" from "the release path stages it differently"

auser:
Closes #2660. Unblocks every `boot-image/v*` release, and therefore #2542.

## The failure

The x86_64 boot gate reached userspace and died:

```
```

## Cause

`default-microvm` is a **lean** rootfs. `mk-guest.nix` resolves the agent from `/mvm/runtime` — the runtime overlay — and the rootfs carries no baked fallback. The gate passed only kernel and rootfs.

The overlay is built in the same release run (`runtime-overlay-image`) and was simply never attached. And no workflow-only fix was possible: `tests/runtime_boot_bench.rs` had no overlay knob at all, so the harness structurally could not boot a lean image.

## The choice

Two coherent fixes existed, and they are not equivalent:

1. **attach the overlay** — the gate boots the pair that actually ships
2. **bake an agent into the published rootfs** — the gate passes unchanged, the rootfs grows, and the overlay stops being the single source of the agent

This takes (1). A gate that boots a differently-composed image is precisely the shape of gate that stays green while the artifact is broken — which is how a partial `boot-image/v0.1.0` came to exist.

## The overlay is a triple, not a path

The backend attaches the overlay only when `runtime_overlay_path`, `runtime_overlay_verity_path` and `runtime_overlay_roothash` are all `Some`. A partial set silently becomes *no overlay*, and the run then fails inside the guest with a panic that never mentions the missing argument.

So `OverlaySpec::from_env` takes all three or none, and refuses anything between with a message naming the actual mistake. An empty roothash file is refused too — the backend needs the hex to thread `mvm.runtime_roothash` into the cmdline, and an empty string would produce a cmdline that looks right.

Four tests cover exactly that: absent, complete, each partial combination, and the empty roothash.

## Cost

`default-microvm` now `needs: [runtime-overlay-image]`, serialising two previously parallel jobs and adding wall clock to each release. The alternative is a gate that cannot boot what it gates.

fmt, clippy, `check-workflow-paths` clean; `overlay_spec_tests` 4/4.

### DIFF
--- a/.github/workflows/release-boot-image.yml
+++ b/.github/workflows/release-boot-image.yml
@@ -377,6 +377,12 @@ jobs:
   default-microvm:
     # Skipped on a dry-run dispatch (see builder-vm-image).
     if: ${{ github.event_name == 'push' || github.event.inputs.dry_run == 'false' }}
+    # The boot gate below attaches the runtime overlay, so this job now waits
+    # on the job that builds it. That serialises two previously parallel jobs
+    # and costs the release some wall clock — worth it, because the gate was
+    # otherwise booting a lean rootfs with no agent in it and calling the
+    # result a failure of the image.
+    needs: [runtime-overlay-image]
     # This job mints a keyless (OIDC/Fulcio) signature over the attested
     # runtime pack manifest under the same release.yml@<tag> identity class as
     # builder-vm-image — gating it behind the same protected environment
@@ -493,6 +499,13 @@ jobs:
       - uses: ./.github/actions/install-zigbuild
         if: matrix.arch == 'x86_64'
 
+      - name: Fetch the runtime overlay the boot gate attaches
+        if: matrix.arch == 'x86_64'
+        uses: actions/download-artifact@v6
+        with:
+          name: runtime-overlay-image-${{ matrix.arch }}
+          path: overlay
+
       - name: Boot the staged image before it becomes a release asset
         if: matrix.arch == 'x86_64'
         # Boots the artifact in `staging/`, never a published one — the whole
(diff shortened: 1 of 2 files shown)


auser:
Closing as redundant — #2662 landed the same fix while this was open, and did it at least as thoroughly.

`main` already carries the harness overlay knob (`MVM_RUNTIME_BOOT_OVERLAY{,_VERITY,_ROOTHASH}`), the `OverlaySpec` triple, the workflow wiring, and the validation I considered this PR's distinguishing contribution:

```
overlay_spec_refuses_a_half_configured_overlay_and_names_what_is_missing
overlay_spec_refuses_a_blank_roothash
```

Both of those cover the case that mattered — the backend attaches the overlay only when all three parts are present, so a partial set silently becomes *no overlay* and fails inside the guest with a panic that never mentions the missing argument. #2662 refuses it up front, which is the behaviour I wanted, so there is nothing here worth rebasing in.

Branch left in place rather than deleted, in case the wording of the diagnostics is worth cherry-picking.

Same root cause, same conclusion, arrived at independently: `default-microvm` is a lean rootfs whose agent lives in the overlay, and the boot gate was booting it without one.

Issue #2675, opened 2026-08-18
Published-kernel fetch is structurally broken: it derives its release tag from the crate version


auser:
## Summary

`mvmctl machine run --image` on a host that cannot compile kernels fails with a hard 404. This is not a one-off missing asset — the acquisition path is structurally unable to work for most of its lifetime.

Found by running a cross-built aarch64 `mvmctl` on a Raspberry Pi 4 against real KVM:

```
Error: no kernel asset vmlinux-aarch64-workload in release v0.18.0.
Build it locally with `--source compile`, or cut a release that publishes kernels.

Caused by:
    HTTP 404 Not Found downloading
    https://github.com/tinylabscom/mvm/releases/download/v0.18.0/vmlinux-aarch64-workload
```

## Three compounding causes

**1. The tag is derived from the crate version, which runs ahead of the last release.**

`crates/mvm-cli/src/update.rs:248`:

```rust
let tag = format!("v{}", current_version());
```

The workspace is at `0.18.0`. **`v0.18.0` has never been released** — `gh release view v0.18.0` returns `release not found`, and the latest release is `v0.17.0`. So every in-development build points its kernel download at a release that does not exist. The fetch can only work in the brief window where the crate version exactly equals a published release.

**2. `kernel-build.yml` has never run on a tag push.**

`gh run list --workflow=kernel-build.yml --event=push` returns **zero runs**. The workflow declares `on.push.tags: ["v*"]` and its upload step publishes `vmlinux-${ARCH}-builder` / `vmlinux-${ARCH}-workload`, but that step has never executed on a tag.

**3. So even the latest release lacks the assets.**

| release | `vmlinux-*` assets |
|---|---|
| v0.18.0 | does not exist |
| **v0.17.0 (latest)** | `default-microvm-vmlinux-{aarch64,x86_64}` only — **no `vmlinux-*-workload`** |
| v0.16.1 | `default-microvm-vmlinux-*` only |
| v0.16.0 | `vmlinux-aarch64-builder`, `vmlinux-aarch64-workload` ✅ |

Only `v0.16.0` carries them. A user on the current latest release 404s just as a dev build does.

## Why it matters

`download_kernel`'s own doc comment states the contract this breaks:

> Available without `builder-vm`: lean clients cannot compile kernels locally, so downloading the release-matched kernel is their supported acquisition path.

[...]

PR #2682, opened 2026-08-18
ci: run the workspace suite natively on aarch64

auser:
## The gap

Every lane in `ci.yml` runs on `ubuntu-latest` (x86_64). The two `Build kernels (aarch64)` jobs do use `ubuntu-24.04-arm`, but they compile a Linux kernel — they never build or test mvm's own Rust.

So **mvm's arch-gated code has never been compiled by CI, and its tests have never run.** `ci.yml` says as much in its own words elsewhere: *"This runner is x86_64, so only the x86_64-linux check variant builds here; the aarch64-linux variant stays defined for parity and for a future arm64 runner."*

## Why this is not theoretical

`crates/mvm-hostd/src/jailer/seccomp.rs` defines `CONFINED_ROLE_SYSCALLS` **twice** — a separate allowlist per architecture. The aarch64 table is the one no runner has ever seen, and the file's own comments describe exactly what a mistake in it costs:

> a missing `flock` SIGSYS-kills the endpoint at the moment it starts working, after the connect it was about to report succeeded.

That is a security-relevant, arch-specific, silent-until-load-bearing failure that x86_64 CI structurally cannot catch. Three existing tests — `syscall_name_to_nr_resolves_known_names`, `bridge_syscalls_has_no_duplicate_names`, and `crates/mvm-hostd/tests/seccomp_property.rs` — resolve against whichever table the target arch selects, so on an arm64 runner they finally exercise the aarch64 one.

`GuestArch::host()`'s non-x86_64 arm, the bundle arch gate, and the aarch64 `mvm-seccomp-apply` arm are all in the same position.

## Scope, stated honestly

The lane proves the suite passes **natively on aarch64**. It does **not** prove a guest boots.

GitHub's arm64 runners expose no nested KVM — `release-boot-image.yml` already records this: *"the aarch64 leg runs on ubuntu-24.04-arm, which has no nested KVM and cannot boot a guest at all."* So `backend.start()` is as unreachable here as in every other lane, and a green run is **not** a Tier-1 boot witness. That still needs real hardware. The job carries this caveat in a comment so a future reader can't mistake it.

## Cost

Workspace suite only. The extras `Test workspace` also runs — man pages, the `addons` feature, doctests — are arch-independent and already covered on x86_64, so duplicating them would be pure cost.

## Also fixed

A latent x86-ism the lane made obvious: `Platform::has_nested_kvm` probed only `/sys/module/kvm_intel/parameters/nested` and `/sys/module/kvm_amd/parameters/nested`. **arm64 has neither vendor module** — the kernel exposes the arch-neutral `/sys/module/kvm/parameters/nested` — so the probe returned "no nested KVM" on every ARM host regardless of the truth. Two new tests cover the neutral node enabled and disabled.

## Follow-up needed from a repo admin

Branch protection currently requires: `Lint (fmt + clippy + policy)`, `Test`, `Nix flake check (Linux eval)`, `Invariant`, `Build kernels (aarch64)`, `Build kernels (x86_64)`.

**`Test workspace (aarch64)` needs adding to that list to actually gate.** Until then it runs and reports but cannot block a merge — which is the same shape as a lane that exists and gates nothing.

## Verification

- `cargo nextest run --workspace` — **12,335 passed**, 0 failed
- `cargo test --workspace --doc` — pass
- `cargo +nightly fmt --all --check` — clean
- `just check-gated` — pass
- `actionlint .github/workflows/ci.yml` — pass
- `xtask check-workflow-paths` — pass

Note the lane itself is only truly exercised once this PR runs in CI, since no local host can stand in for the runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -769,6 +769,56 @@ jobs:
         # nextest does not run doctests; gate them separately.
         run: cargo test --workspace --doc
 
+  test-workspace-aarch64:
+    name: Test workspace (aarch64)
+    needs: [scope]
+    if: needs.scope.outputs.code == 'true'
+    # The only lane that executes mvm's aarch64 code paths. Everything else in
+    # this workflow runs on x86_64, so arch-gated code compiles here for the
+    # first time and its tests run for the first time. The sharpest example is
+    # `jailer::seccomp::CONFINED_ROLE_SYSCALLS`, which is a *separate table* per
+    # architecture: a wrong or missing entry in the aarch64 half SIGSYS-kills a
+    # confined role at the moment it starts working, and no x86_64 runner can
+    # see it. `GuestArch::host()`, the bundle arch gate, and the aarch64
+    # `mvm-seccomp-apply` arm are in the same position.
+    #
+    # SCOPE, honestly: this proves the suite passes natively on aarch64. It does
+    # NOT prove a guest boots — GitHub's arm64 runners expose no nested KVM, so
+    # `backend.start()` is unreachable here exactly as it is in every other lane.
+    # A green run here is not a Tier-1 boot witness; that still needs real
+    # hardware.
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/free-disk
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+
+      - uses: ./.github/actions/apt-deps
+        with:
+          packages: attr cryptsetup-bin libcap-ng-dev lld
+
+      - uses: ./.github/actions/install-zigbuild
+
+      - uses: ./.github/actions/rust-cache
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+
+      - name: Build
+        run: cargo build --all-targets
+
+      - name: Run tests
+        # The workspace suite only. The arch-independent extras Test workspace
+        # also runs (man pages, the addons feature, doctests) are covered on
+        # x86_64 and would only duplicate cost here.
+        run: cargo nextest run --workspace --all-targets
+
   test-release-witness:
     name: Test release profile (overflow + no debug assertions)
     needs: [scope]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -651,7 +651,7 @@ jobs:
   test:
     name: Test
     if: ${{ always() }}
-    needs: [scope, test-workspace, test-linux, test-release-witness, test-ebpf-telemetry, bdd-conformance, kernel]
+    needs: [scope, test-workspace, test-workspace-aarch64, test-linux, test-release-witness, test-ebpf-telemetry, bdd-conformance, kernel]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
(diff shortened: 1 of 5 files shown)


auser:
## First run: the lane immediately earned itself

It failed, and not on our code:

```
error[E0308]: mismatched types
  --> am-fs-ext4-0.4.0/src/capi.rs:1100:19
1100 |         name[i] = b as c_char;
     |         -------   ^^^^^^^^^^^ expected `i8`, found `u8`
```

`am-fs-ext4` assumes `c_char == i8`. That holds on **x86_64-linux** and on **aarch64-*darwin*** — but on **aarch64-linux** `c_char` is `u8`. So the crate compiles on every platform this project has ever built for, and fails on the one nobody had tried. That is the exact shape of bug a native ARM lane exists to find, surfaced on its first execution.

## Why it is excluded rather than fixed or dropped

- It is a **dev-dependency of `mvm-fs` only** — not in the shipped `mvmctl`. (Which is why a `cargo zigbuild --bin mvmctl` for `aarch64-unknown-linux-gnu` succeeds and runs fine on a Raspberry Pi: the crate is never in that graph.)
- It is **genuinely load-bearing on x86_64**. My first instinct was that it was an unused dep — `grep am_fs_ext4` finds nothing. That was wrong: the crate's lib name is `fs_ext4`, and `tests/oracle.rs` and `tests/xattr.rs` use it as the independent ext4 reader our writer's output is checked against. Removing it would delete a real correctness oracle.
- `0.4.0` is the newest published version, so there is nothing to bump to.

So the oracle stays an x86_64 oracle, and `mvm-fs`'s **test targets** are skipped on this lane only. Its library still builds here, as a dependency of everything else.

## Note on local verification

The seccomp justification cannot be checked from a macOS dev host: `jailer::seccomp` and `tests/seccomp_property.rs` are `cfg(target_os = "linux")`, so those tests do not exist locally at all. Confirming they run — against the aarch64 table specifically — is something only this lane can do, which is rather the point.

## Unrelated failure on this PR

`Boot latency ceiling` is also red, and it is **not** from this change. It fails at `curl: (22) ... 404` fetching assets from `IMAGE_TAG: boot-image/v0.1.0`, and `gh release view boot-image/v0.1.0` returns **release not found**. That lane is broken for every PR right now, and it is the same class of bug as #2675: code pointing at a release tag that does not exist.

auser:
## No branch-protection change needed after all

The PR description said `Test workspace (aarch64)` had to be added to branch protection by an admin. That was the wrong route — the repo already has a better one.

**`Test` is an aggregator, and it is already a required context.** It `needs:` every test lane and fails unless each matches the result its scope demands:

```yaml
needs: [scope, test-workspace, test-linux, test-release-witness, test-ebpf-telemetry, bdd-conformance, kernel]
```

```bash
for result in "$WORKSPACE_RESULT" "$LINUX_RESULT" "$RELEASE_WITNESS_RESULT" "$EBPF_RESULT"; do
  if [ "$result" != "$required" ]; then
    echo "A test lane did not match scope $SCOPE_CODE: $result"
    exit 1
```

So adding the aarch64 lane to that aggregator makes it blocking **with no branch-protection change and no seventh context to keep in sync**. Done here.

It joins the `SCOPE_CODE` loop rather than being matched separately, because it carries the same `if: needs.scope.outputs.code == 'true'` as the other four lanes in that loop — `success` in scope, `skipped` out of it.

This also avoids the check-name fragility the `kernel` job's comment documents at length: a required context that does not match a reported check name leaves every PR unmergeable.

## Rebased

Rebased onto main to pick up #2697 (seccomp `epoll_pwait`) and #2700 (doctor platform derivation), the two real failures this lane found. The third, `seccomp_audit_cli`, should clear with #2697.

Issue #2681, opened 2026-08-18
The boot gate cannot boot a sealed prod image: no initramfs, no rootfs verity, no runtime-source policy


auser:
The boot gate added to `release-boot-image.yml` cannot boot a **sealed prod
image**, and neither can `ci.yml`'s `boot-latency` lane. Both have now been
fixed twice and still fail, because the missing pieces were never one thing.
This issue records the full shape so the next attempt is not another guess.

## Where it stands

Latest run against `boot-image/v0.1.0` at `be3efc23e` — every job green except
one:

```
success  builder-vm-image (aarch64)      <- MVM_HOST_BIN_DIR + --impure fix works
success  builder-vm-image (x86_64)
success  default-microvm (aarch64)       <- no boot gate on this arch
failure  default-microvm (x86_64)        <- the boot gate
failure  publish-boot-image              <- completeness gate, correctly refusing
```

The x86_64 job fails only at `Boot the staged image before it becomes a release
asset`. Console:

```
mvm-init: no guest agent resolved from /mvm/runtime and no baked fallback
Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
```

`/init` runs — the byte-0 shebang fix holds. The agent is what is missing.

## Why the overlay fix was not enough

The gate now passes the overlay triple, and the env reaches the boot step
(`MVM_RUNTIME_BOOT_OVERLAY`, `_OVERLAY_VERITY`, `_OVERLAY_ROOTHASH=bbbaaeeb…`).
The drive attaches. The guest still cannot find the agent, because **attaching
the overlay is not the same as telling the guest where it is**, and a sealed
image has two ways to be told, neither of which the harness sets up:

- **Plain boot** — `mk-guest.nix`'s `/init` reads `mvm.runtime_data=/dev/vdN`
  off `/proc/cmdline` and mounts the overlay itself. Its own comment calls this
  the dev-boot path: *"a plain dev boot has no initramfs, so the overlay rides a
  read-only virtio-blk device the host names on the kernel cmdline"*.
- **Sealed boot** — `mvm-verity-init` runs from an **initramfs** and mounts the
  overlay at `/mvm/runtime` before `switch_root`. This is what production does,
  and what a `sealed:true` image expects.

`tests/runtime_boot_bench.rs` sets `kernel`, `rootfs`, and the overlay triple.
It sets no `initrd_path`, no `verity_path`, no `roothash`, and no
[...]

PR #2686, opened 2026-08-18
fix(ci): boot the sealed image the way production boots it

auser:
Closes the release-side half of #2681.

The boot gate handed Firecracker a kernel, a rootfs and a runtime overlay. That
is not how a sealed image boots.

`mvm-verity-init` runs from an **initramfs** and mounts the rootfs by roothash
and the overlay at `/mvm/runtime` before `switch_root`. With no initramfs the
kernel execs the rootfs `/init` directly, which walks its whole agent ladder and
finds nothing:

```
```

A prod rootfs carries no baked agent, so that ladder has no last resort. **The
image was fine every time.** The harness was booting a shape nothing ships.

## What changed

`tests/runtime_boot_bench.rs` now takes the verity sidecar and roothash, and
boots under `RuntimeSourcePolicy::RequiredOverlay` so a missing overlay refuses
the boot rather than falling through to a branch a prod image cannot satisfy.

The initramfs is resolved through `resolve_or_build_local_initramfs` — the same
function the CLI's boot path calls via `attach_universal_initramfs_if_cached` —
rather than a second acquisition path that could drift from it. On Linux with a
cold cache it builds one, which is what the runner does.
`MVM_RUNTIME_BOOT_INITRD` overrides it for a host that cannot.

`release-boot-image.yml` passes both sidecars. They were already staged and
already published; the gate simply never passed them — verity as a path,
roothash as file contents.

## Why the failures kept moving

Each attempt exposed exactly one layer, because the next only becomes reachable
once the previous is fixed:

| attempt | died on | fixed by |
|---|---|---|
| 1 | no pinned zig; the harness failed to compile | #2645 |
| 2 | overlay never attached — `VmStartConfig` triple unset | #2662 |
| 3 | `release-signing` refused a `boot-image/v*` tag | #2671 |
| 4 | overlay attached, guest never told where it is | **this** |

## Tests

Sealed-boot coverage mirrors the overlay triple's: absent, complete,
verity-without-roothash, blank roothash. A half-configured set is refused **by
name** — a silently-dropped sidecar boots the unsealed shape and fails as though
the image were broken, which is precisely the diagnosis that cost four release
runs.

13 pass. `actionlint` clean, including the shellcheck split of the roothash read
from its export.

## What this does not fix

`ci.yml`'s `boot-latency` lane. That one boots a **published** image, and no
release publishes an initramfs — `nix/images/default-tenant` emits
`{vmlinux, rootfs.ext4, rootfs.verity, rootfs.roothash, mvm-meta.json}` and
nothing else. That artifact gap is recorded in #2681 and still blocks #2542.

### DIFF
--- a/.github/workflows/release-boot-image.yml
+++ b/.github/workflows/release-boot-image.yml
@@ -564,6 +564,15 @@ jobs:
           MVM_RUNTIME_BOOT_ROOTFS: staging/default-microvm-rootfs-x86_64.ext4
           MVM_RUNTIME_BOOT_OVERLAY: overlay-staging/overlay.ext4
           MVM_RUNTIME_BOOT_OVERLAY_VERITY: overlay-staging/overlay.verity
+          # A sealed image boots through its initramfs: mvm-verity-init mounts
+          # the rootfs by roothash and the overlay at /mvm/runtime before
+          # switch_root. Handing over kernel + rootfs alone execs the rootfs
+          # /init directly, which finds no agent and panics — so the image looks
+          # broken when it is not. The sidecar is staged above; the roothash is
+          # exported below because it is file contents, not a path; and the
+          # harness resolves the initramfs itself through the same function the
+          # CLI boot path uses, building one here since the runner is Linux.
+          MVM_RUNTIME_BOOT_VERITY: staging/default-microvm-rootfs-x86_64.verity
           MVM_RUNTIME_BOOT_READY: guest-agent
           MVM_RUNTIME_BOOT_RUNS: "1"
           MVM_RUNTIME_BOOT_CONCURRENT: "1"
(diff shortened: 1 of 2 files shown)


auser:
Closing in favour of #2683, which fixes this correctly and more narrowly.

I had the symptom right and the mechanism wrong. My conclusion was that a prod image must boot the **sealed** path — initramfs, `mvm-verity-init`, rootfs verity + roothash — and this PR wired all of that up.

#2683 found the actual gate:

```rust
pub fn non_verity_overlay_ext4(config: &VmStartConfig) -> Option<&str> {
    if config.runtime_source_policy == RuntimeSourcePolicy::RootfsOnly {
        return None;
    }
```

`RootfsOnly` is the derive default, so `..Default::default()` in the bench suppressed the `mvm.runtime_data=` cmdline token while the host still attached the drive on the strength of the triple alone. The guest got the disk and no way to name it — indistinguishable at the console from no overlay at all, which is why it read as an image fault.

So no initramfs is needed: the **non-verity** path boots this image fine once the policy is declared. Everything this PR added on top of that — verity sidecar, roothash, initramfs resolution — forces a heavier boot shape the image does not require, and it collides with #2683 on the same file.

The half-configured-set refusals here were worth having, but #2683's `runtime_source_policy(overlay)` derivation makes the state unrepresentable instead, which is better than refusing it.

Superseded. #2681 stays open for the remaining artifact gap: no release publishes an initramfs, which still blocks `boot-latency` from ever booting a *published* prod image sealed.

PR #2711, opened 2026-08-18
feat(sdk): add opt-in Obscura browser provider

auser:
## Summary

- add an explicit, experimental `obscura` browser provider to the Python and TypeScript SDKs while keeping Chromium as the unchanged default
- distinguish manifest and digest-pinned OCI boot sources, lower representable live options, and reject unsupported or secret-bearing options before boot
- fix Obscura to guest loopback CDP plus the mvm egress proxy, with bounded readiness checks and teardown on failure
- add a pinned cross-architecture Nix guest example, regression coverage, and an owning implementation plan

## Safety and scope

- Obscura is opt-in only; there is no automatic fallback or default-provider change
- the provider uses the immutable upstream OCI index digest and rejects command overrides
- CDP binds `127.0.0.1:9222`; browser egress is routed through `127.0.0.1:1080` and mvm admission
- this remains a draft pilot until real-backend policy proof and compatibility work are complete

## Validation

- Python SDK: 229 passed, 7 skipped
- TypeScript SDK: 149 passed; build and typecheck passed
- Rust SDK: 282 unit tests plus integration and documentation tests passed
- Obscura example contract: 3 passed
- `cargo check --workspace --all-targets`: passed
- `cargo clippy -p mvm-sdk -p mvm-cli --all-targets -- -D warnings`: passed
- `just check-gated`: passed
- builder VM Nix evaluation, package build, and complete `aarch64-linux` guest build: passed
- plan-name, sprint-append, no-spec-reference, formatting, and diff checks: passed
- the full workspace run had one unrelated parallel environment race in `doctor::toolchain::tests::check_cmd_rustup_on_host`; its isolated rerun passed

## Remaining pilot gates

- real microVM CDP lifecycle and positive/negative egress policy witness
- Playwright/Puppeteer compatibility corpus and lifecycle benchmarks
- complete Nix architecture matrix and native Linux all-target Clippy
- clean one-shot full workspace test run


auser:
`check-l3-expansion-freeze` is failing this, and I think it is a false positive
— but resolving it is your call, because every fix touches either your error
messages or a deliberately-ratcheted gate.

## What the gate sees

Six new non-test references to `raw_ip_stack`, in
`sdks/python/mvm/_sandbox.py` and `sdks/typescript/src/_sandbox.ts`.

## What the code actually does

Refuses it:

```python
if encoded_network.get("raw_ip_stack") is True:
    _reject_live_option("network.raw_ip_stack", "Obscura must stay on the proxy path")
```

The other occurrences are the field appearing in the *known-fields* set so that
unknown-field detection keeps working. Nothing here reaches the retired stack —
this is the freeze being enforced, not extended. The gate matches on the symbol
and cannot tell those apart.

## Three ways out, none of which I want to pick for you

1. **Allowlist both files.** There is precedent: the gate's own allowlist
   carries `crates/mvm-core/src/plan/l3_retirement.rs` with the comment *"The
   refusal itself has to name what it refuses."* That is exactly this case.
   Against it: the gate's doc says *"Append nothing"* and *"it can only
   shrink"*, and the ratchet is most of its value.

2. **Drop the field from the known set and the explicit check.** All six
   references disappear and the behaviour is unchanged — an unrecognised field
   already gets rejected by the `unknown fields` path. The cost is the error
   message: `unknown fields: ['raw_ip_stack']` instead of
   `Obscura must stay on the proxy path`. I did not do this because calling a
   field the SDK type genuinely has "unknown" is a worse message, and your
   wording is better.

3. **Teach the gate that a refusal is not an expansion** — e.g. don't count an
   occurrence on a line that also rejects. Narrow and defensible, but it is a
   shared gate and softening its matcher deserves its own discussion.

Everything else on this PR is fixable and I have left it alone: the `Test`,
`Test workspace` and `Lint` failures were the apt-mirror stall, fixed on main in
#2693, so merging main should clear them. `Boot latency ceiling` is red
repo-wide on the boot-image 404 and is being fixed separately.

Issue #2635, opened 2026-08-17
Guest-image generator changes reach main without any lane building or booting them


auser:
Two related gaps, found while finishing #2542. Both are about the same thing:
**changes to the guest image generator reach `main` without any lane building or
booting what they produce.** This is the mechanism that let #2539 survive five
weeks.

Everything below was read off `main` at `94e4121c5`, not from memory.

## 1. The cheap guard that exists is wired to the wrong scope

`mk_guest_init_shebang_lands_at_byte_zero` (`tests/nix_flake_structure.rs:735`)
renders `initText` out of `nix/lib/mk-guest.nix` and asserts the first line is
`#!/bin/sh`. Pure Rust, no Nix, sub-second. It exists precisely to catch #2539.

It runs in `test-workspace`, which is gated `if: needs.scope.outputs.code == 'true'`
(`ci.yml:682`). And `scope.code` (`ci.yml:88`) matches `nix/` only via
`nix/packaging/release/[^/]*\.sh$`.

So a PR touching only `nix/lib/mk-guest.nix` resolves:

| output | value |
|---|---|
| `code` | **false** |
| `kernel` | **false** — `scope.kernel` (`ci.yml:105`) covers `nix/images/kernel/` but not `nix/lib/**`, not `nix/images/default-tenant/**` |
| `nix` | true — but its only consumer, `nix-flake-check`, is skipped on `pull_request` by design (`ci.yml:1105`: "run it in the merge queue and on manual dispatch, not on every PR update") |

That is exactly the shape of the change that broke `v0.17.0`. `b4a6fadd7`'s diff is
one line moving from three spaces to four in the `initText` block of
`nix/lib/mk-guest.nix` — nothing else. Such a PR gets no Nix build, no boot, and
does not run the one test written to catch it.

## 2. Nothing boots a tree-built image before merge

Exactly three jobs in `.github/workflows/` start a real VMM:

| Job | Image source | Trigger |
|---|---|---|
| `ci.yml::boot-latency` | downloaded published release (`IMAGE_TAG`) | `workflow_dispatch` today; #2542 restores PRs |
| `release-boot-image.yml::default-microvm` | **built from the tree** | `push: tags: boot-image/v*` |
| `ci-full.yml::workload-spawn-smoke-linux` | upstream Firecracker-CI kernel + a blank 64 MiB ext4 | `workflow_dispatch` only |

`ci.yml::kernel` does build both kernels from the tree via
[...]

PR #2716, opened 2026-08-18
fix(bench): mint the host signer identity the readiness probe signs with

auser:
The next — and, on the evidence, last — layer of the boot-image release chain.

## The guest now boots

This run got further than any before it. The VM started, ran, and the teardown
reported a guest-filesystem flush, which only a live guest produces. Then:

```
Error: read host signer key /home/runner/.mvm/keys/host-signer.ed25519
Caused by: No such file or directory (os error 2)
```

The readiness probe opens an **authenticated** session, and the host half of
that handshake signs with that key. A real `mvmctl` run reaches it through
`host_keypair::load_or_init`, which mints it on first use. The harness went
straight to the probe, so on a fresh runner — which is every CI runner — a
working guest was reported as a broken image.

Call the same initializer once, before the first boot.

## Why not make the read path mint it

`open_authenticated_session_stream` could create the key when it is missing, and
that would be the smaller diff. It would also be wrong.

That key is the host's signing identity and the anchor the guest pins in
`/run/mvm/host-signer.pub`. A read path that silently mints one makes "there is
no key yet" indistinguishable from "the key was replaced" — which is exactly the
distinction the pinning exists to detect. The caller that wants an identity asks
for one; the caller that wants to *use* an identity should fail when there
isn't one.

## Progress through this chain, for the record

Each fix exposed the next, and every one was a real defect:

| | died on |
|---|---|
| 1 | no pinned zig — `cargo test` would not compile |
| 2 | overlay never attached |
| 3 | `release-signing` environment refused a `boot-image/v*` tag |
| 4 | overlay attached, guest never told where it is (#2683) |
| 5 | `runtimeLean` dropped by the sidecar serializer (#2699) |
| 6 | host signer key never minted — **this PR** |

Layers 4 and 5 both required a real boot to observe, which is what #2635 is
about.

`cargo test --test runtime_boot_bench` 13/13.

### DIFF
--- a/tests/runtime_boot_bench.rs
+++ b/tests/runtime_boot_bench.rs
@@ -66,6 +66,8 @@ fn prebuilt_runtime_image_boots_within_budget() -> Result<()> {
         return Ok(());
     };
 
+    ensure_host_signer()?;
+
     let serial = measure_serial(&spec)?;
     let serial_summary = summarize(&serial);
     eprintln!(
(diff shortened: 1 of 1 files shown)


auser:
Closing in favour of #2714, which fixes the same thing better and landed first.

We diagnosed this independently and reached the same conclusion — including the
important half, that the read path must *not* mint a key on demand, because
that would make "no key yet" indistinguishable from "the key was replaced".
#2714 goes further on both counts:

- it calls `load_or_init_at(keys_dir)` rather than the ambient-`MVM_HOME`
  variant, so the behaviour is exercisable without touching the real home;
- it identifies the precise reader as the per-tenant host-agent daemon spawn
  path, where mine said only "the readiness probe";
- it notes that `load_or_init` is idempotent and why that matters — rotating
  would orphan any audit chain already signed under the old key, which is a
  consequence I had not written down.

My fault for not checking open PRs before writing it. The one thing worth
carrying over is the chain summary from #2716's description, which records that
this is layer 6 and that layers 4 (#2683) and 5 (#2699) were both only
observable from a real boot — the argument #2635 is making.

PR #2725, opened 2026-08-19
fix(checkpoint): measure a fork child's grants against the tier that boots it

### LINKED ISSUE #2723: A checkpoint whose parent bounded CPU or wall clock cannot be forked at all
## What

`admit_forked_child` passes `backend_kind: None` to `admit_plan_for_boot`, so
the grant gate has no backend to measure against and takes its fail-closed arm:

```
refusing an unenforceable grant: this run was admitted without the backend that
will boot it, so no mechanism can be named for the grants it declares
```

A fork child inherits `parent_meta.grants` and is admitted prod-profile
(`grant_eligible(false, false, false)` is unconditionally true, so
`restrict_agent_verbs` is true, so `Variant::Prod`). Under Prod that arm is a
hard refusal rather than a warning.

**Net effect: any checkpoint whose parent carried a CPU or wall-clock grant is
unforkable**, on every backend, and the error names a missing backend rather
than the grant — so it reads as an internal fault, not a policy decision.

Reproduced by admitting a fork child whose parent meta carries
`CpuGrant::Share { millicores: 1500 }`.

## Why the obvious fix is wrong

`AdmitPlanForBootParams::backend_kind` documents itself:

> the backend object, never parsed from `backend_name` — a name is a label, and
> a grant checked against a label is checked against whatever the caller typed.

So `BackendKind::from_label(backend_name)` is exactly what that field exists to
prevent, and is not the fix.

## The fix

Each fork arm is one backend and knows it as a constant, so pass the tier
**typed** rather than as a label: `AdmitForkedChildParams` takes a
`BackendKind`, the arms pass `BackendKind::Firecracker` / `BackendKind::Hvf`,
and the plan's `backend_name` is derived from it. Two values from one typed
source cannot drift into disagreeing about what is about to boot.

The safety property is preserved: a grant the real tier cannot enforce still
refuses under a prod posture. What changes is that the gate now has a mechanism
to measure against instead of failing closed for lack of information.

## Related, not fixed here

The child's prod posture comes from a hardcoded `is_dev_profile: false` in
`admit_forked_child`'s `grant_eligible(false, false, false)` call. The first two
literals are facts about forks (no PTY, no ad-hoc argv); the third is not. With
the backend passed, it matters much less, but it is still a hardcoded constant
where every other caller derives the value.

auser:
Closes #2723.

## The bug

`admit_forked_child` passed `backend_kind: None`, so the grant gate took its
fail-closed "no backend object, no answer" arm. A fork child inherits
`parent_meta.grants` and is admitted prod-profile, where that arm is a hard
refusal:

```
refusing an unenforceable grant: this run was admitted without the backend that
will boot it, so no mechanism can be named for the grants it declares
```

**Any checkpoint whose parent bounded CPU or wall clock was unforkable**, on
every backend — and the error named a missing backend rather than the grant, so
it read as an internal fault rather than a policy decision.

## The fix

Each fork arm is one backend and knows it as a constant, so pass the tier
**typed**: `AdmitForkedChildParams` takes a `BackendKind`, the arms pass
`BackendKind::Firecracker` / `::Hvf`, and the plan's `backend_name` is derived
from it. Two values from one typed source cannot drift into disagreeing about
what is about to boot — a disagreement the gate refuses outright.

**Not parsed from the label.** `AdmitPlanForBootParams::backend_kind` documents
that it comes from the backend object and never from `backend_name`, "because a
grant checked against a label is checked against whatever the caller typed".
`BackendKind::from_label(backend_name)` is the fix this field exists to prevent;
threading the typed kind is the one that respects it. This also removes a
stringly-typed field from the params struct.

## The safety property is unchanged

A share the tier genuinely cannot serve is still refused, and now says why:

```
refusing an unenforceable grant: the hvf tier cannot enforce cpu.share:
CPU share must be between 1 and 1000 millicores (1.0 core) on this tier; got 1500
```

That is asserted by its own test, including that the message is no longer the
no-backend arm.

## Verification

- 3 tests: an enforceable grant forks; an unenforceable share still refuses and
  names the tier; the plan records the tier the gate measured.
- Mutation-checked: restoring `backend_kind: None` turns both grant tests red
  (the refusal test too, since the refusal reverts to the wrong reason) while
  the tier-agreement test correctly stays green.
- Full `mvm-cli` suite 1924 passed, workspace clippy clean, `fmt --all` clean.

## Related, not fixed here

The child's prod posture comes from a hardcoded `is_dev_profile: false` in
`grant_eligible(false, false, false)`. The first two literals are facts about
forks; the third is a constant where every other caller derives the value. With
the backend passed it matters much less, but it is still worth revisiting.

### DIFF
--- a/crates/mvm-cli/src/commands/vm/checkpoint/fork_vm_full.rs
+++ b/crates/mvm-cli/src/commands/vm/checkpoint/fork_vm_full.rs
@@ -10,6 +10,7 @@
 //! guest has no NIC to inherit an address on.
 
 use anyhow::{Context, Result};
+use mvm_contract::protocol::vm_backend::BackendKind;
 use mvm_core::checkpoint::{CheckpointId, VmFullOrigin, vm_full_origin};
 use mvm_core::config::vm_state_dir;
 use mvm_runtime::checkpoint::{CheckpointStore, ForkParams, fork_vm_full};
@@ -205,7 +206,7 @@ pub(in crate::commands) fn fork_vm_full_arm_fc(
         checkpoint: p.checkpoint,
         parent_meta: &p.parent_meta,
         child_vm_name: &p.child_vm_name,
-        backend_name: "firecracker",
+        backend_kind: BackendKind::Firecracker,
         declared_secrets: p.declared_secrets,
     })?;
 
@@ -298,7 +299,11 @@ struct AdmitForkedChildParams<'a> {
     checkpoint: &'a CheckpointId,
     parent_meta: &'a mvm_core::checkpoint::CheckpointMeta,
     child_vm_name: &'a str,
-    backend_name: &'a str,
+    /// The tier that will boot this child. Typed, not a label: the grant gate
+    /// measures against it, and a grant checked against a string is checked
+    /// against whatever the caller typed. The plan's `backend_name` is derived
+    /// from this, so the two cannot disagree.
+    backend_kind: BackendKind,
     /// Secret bindings the caller declares for this child.
     ///
     /// Declared, never inherited: the parent's set is not consulted, so the
@@ -354,7 +359,7 @@ fn admit_forked_child(p: &AdmitForkedChildParams<'_>) -> Result<AdmittedForkChil
             network_mode: super::parent_network_mode(p.checkpoint, p.store),
             tenant: &tenant,
             vm_name: p.child_vm_name,
-            backend_name: p.backend_name,
+            backend_name: p.backend_kind.as_str(),
             rootfs_path: &rootfs_blob,
             precomputed_image_sha256: recorded_sha,
             boot_artifact_identity: None,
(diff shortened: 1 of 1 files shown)


auser:
Pushed a fix for the failing test, verified on both platforms.

## What was wrong

`a_parent_that_bounded_cpu_can_still_be_forked` hardcoded `BackendKind::Hvf`,
which meters CPU **only on macOS**:

```rust
BackendKind::Hvf | BackendKind::AppleContainer => Self {
    cpu: if cfg!(target_os = "macos") { CpuControl::HvfVcpuQuota }
         else { CpuControl::None },
```

There is no in-process vCPU scheduler to enforce a share on Linux, so on Linux
the refusal the test was written to disprove is the **correct** answer:

```
refusing an unenforceable grant: the hvf tier cannot enforce every declared
grant: cpu.share — this tier has no CPU quota mechanism
```

The test asserted the opposite of what the tier can do there. It passed on a Mac
and failed in CI for a real reason, not a flake.

## The fix

Firecracker is the mirror image — `CgroupShare` on Linux, nothing on macOS — so
no single kind is right on both. The test now selects the tier that actually
meters CPU on the host running it, with the reasoning in a comment so the next
person does not re-hardcode one.

The companion `a_share_the_tier_cannot_serve_is_still_refused` already covers
the other direction, so the pair now reads "admitted where a mechanism exists,
refused where it does not" on either platform, instead of only on one.

## Verified on both

- macOS: `cargo test -p mvm-cli --lib checkpoint::fork_vm_full` → 7/7
- Linux (x86_64 KVM host, the platform it was failing on) → **7/7**

I also merged main in. One process note: my first push carried only the merge —
I had patched the file but not committed it, so the Linux run I did to check
still exercised the old code and "failed" misleadingly. Second push has the
actual change; the 7/7 above is from `b4a4149e6`.

PR #2739, opened 2026-08-19
fix(update): fetch published images from the boot image tag, not the CLI version

### LINKED ISSUE #2675: Published-kernel fetch is structurally broken: it derives its release tag from the crate version

auser:
Closes #2675 — which was **closed as completed on 2026-08-18 with the code unchanged**. `download_kernel` still built its URL from `CARGO_PKG_VERSION`, so the bug is live.

## The bug, still reproducible

The CLI ships from `v<crate version>`; the guest images ship from `boot-image/vN`, on a counter that moves independently so a kernel fix does not wait for a CLI release. Deriving an image URL from the crate version points at a tag nobody has published for most of a release cycle:

| | |
|---|---|
| workspace version | `0.18.0` |
| `v0.18.0` release | **does not exist** |
| newest release with kernel assets | `v0.16.0` |

So a lean client — the one that *cannot* compile kernels, whose "supported acquisition path" this is — 404s.

## The switch was scaffolded and waiting

Three separate places in the tree pre-marked this, each naming the same unblock condition:

- `DEFAULT_BOOT_IMAGE_TAG`: *"**Nothing reads this yet.** ... Publishing that tag is the condition that unblocks the switch."*
- `boot/update.rs`: *"the boot-image release train has not published its first tag yet ... **when the image train ships, its own identity belongs here.**"*
- `release_trust`'s boot image identity templates: **no caller outside their own tests.**

`boot-image/v0.1.0` published on 2026-08-19 with 42 assets. The condition is met.

## What moves

Four fetch paths: the kernel, the default microVM image, the Stage 0 builder image, and `image boot update`.

Each now also declares **which release train** signed what it fetched. The two trains sign different artifact classes under different refs, and `release_trust` keeps their identity sets deliberately disjoint so a boot image signature cannot validate a CLI tarball or the reverse. A `ReleaseTrain` on the request makes that a decision at the fetch site rather than an accident of which helper the caller reached for.

## The kernel is a rename, not a substitution

The trains publish the same bytes under different names. `default-tenant`'s flake settles it:

> Workload kernel — the single shared definition in `nix/images/kernel/`, **identical to the one builder-vm builds**.

| variant | boot image asset | checksums |
|---|---|---|
| workload | `default-microvm-vmlinux-<arch>` | `default-microvm-<arch>-checksums-sha256.txt` |
| builder | `builder-vm-vmlinux-<arch>` | `builder-vm-<arch>-checksums-sha256.txt` |

Verified against the published manifests — both arches, both variants, each checksum file containing its kernel.

## Deliberately not moved

- **`runtime_overlay.rs` / `sdk_sidecar.rs`** — same root cause, but their version is *also* the on-disk cache key, so splitting download tag from cache identity is a separate change. They keep `ReleaseTrain::Cli` explicitly, so the choice is recorded rather than defaulted.
- **`bootstrap.rs`** — fetches `*.pack-manifest.json`, which the release job's `nullglob` silently dropped. Repointing it trades one 404 for another.

## Guard

Over the class, not the instance: no image fetch may contain `releases/download/v{version}`. Mutation-tested — restoring it in Stage 0 fails with the file named:

```
these build an image release URL from the CLI version, which 404s whenever the
crate version is ahead of the last CLI tag: ["commands/env/builder_vm/stage0_cache.rs"].
Use `update::boot_image_release()`.
```

Plus unit tests that the image tag is not the CLI version, that the bare semver matches what the signing-identity template interpolates, that each variant maps to its published asset, and that an unknown variant is refused rather than guessed.

## Verification

`cargo nextest run --workspace` — 12,456 passed. Three failures were attributed, not waved through:

- `check-l3-expansion-freeze` and `check-no-gateway-names` fail **identically on unmodified main** in this checkout: two worktrees are nested at `.claude/worktrees/`, and the workspace-scanning gates walk into them and count duplicated source. Environmental; CI checks out clean.
- `run_detached_spawns_redirects_console_and_reports_exit` passes in isolation; it failed only while two suites ran concurrently.

`clippy --workspace --all-targets -D warnings`, `fmt --all --check`, and `actionlint` clean.

## Not claimed

That a fresh install now boots end to end. That needs a real download against a live release with cosign, which no local run can stand in for.

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -29,11 +29,9 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@nightly
 
-      - name: Install system build dependencies
-        run: |
-          sudo rm -f /etc/apt/sources.list.d/microsoft*.list /etc/apt/sources.list.d/azure*.list
-          sudo apt-get update
-          sudo apt-get install -y libcap-ng-dev lld
+      - uses: ./.github/actions/apt-deps
+        with:
+          packages: libcap-ng-dev lld
 
       - uses: ./.github/actions/install-zigbuild
 
--- a/crates/mvm-build/src/release_signature.rs
+++ b/crates/mvm-build/src/release_signature.rs
@@ -57,6 +57,26 @@ pub struct ReleaseSignatureRequest<'a> {
     /// Version whose signing identity is accepted. The release identity is
     /// tag-bound, so this pins *which* release may have signed the archive.
     pub version: &'a str,
+    /// Which release train published the asset, and therefore which workflow
+    /// identity may have signed it.
+    pub train: ReleaseTrain,
+}
+
+/// The release train an asset was published on.
+///
+/// The two trains sign different artifact classes under different refs, and
+/// `mvm_core::release_trust` keeps their identity sets deliberately disjoint so
+/// a signature minted over a boot image cannot validate a CLI tarball or the
+/// reverse. Making the caller name its train keeps that separation a decision
+/// at the fetch site rather than an accident of which helper it reached for.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum ReleaseTrain {
+    /// `release.yml`, signing CLI tarballs under `refs/tags/v<version>`.
+    #[default]
+    Cli,
+    /// `release-boot-image.yml`, signing images under
+    /// `refs/tags/boot-image/v<version>`.
+    BootImage,
 }
 
 /// Verify `request`'s archive against the release workflow's keyless signing
@@ -106,11 +126,17 @@ fn verify_against_release_identities(
     bundle: &[u8],
     request: &ReleaseSignatureRequest<'_>,
 ) -> Result<(), RuntimeOverlayError> {
+    let identities = match request.train {
+        ReleaseTrain::Cli => mvm_core::release_trust::accepted_release_identities(request.version),
+        ReleaseTrain::BootImage => {
+            mvm_core::release_trust::accepted_boot_image_identities(request.version)
+        }
+    };
     verify_release_archive_bytes(
         archive,
         bundle,
         request.asset,
-        &mvm_core::release_trust::accepted_release_identities(request.version),
+        &identities,
         mvm_core::release_trust::RELEASE_OIDC_ISSUER,
     )
 }
@@ -196,6 +222,7 @@ mod tests {
             asset: ASSET,
             archive_path: &path,
             version: FIXTURE_VERSION,
+            train: ReleaseTrain::Cli,
         })
         .map_err(|e| e.to_string())
     }
--- a/crates/mvm-build/src/runtime_overlay.rs
+++ b/crates/mvm-build/src/runtime_overlay.rs
@@ -935,6 +935,10 @@ pub fn download_runtime_overlay(
             asset: &names.archive,
             archive_path: &archive_local,
             version,
+            // Unchanged: this path still resolves its base URL from the CLI
+            // version, so its signature train must stay the CLI one. Splitting
+            // its download tag from its cache key is a separate change.
+            train: crate::release_signature::ReleaseTrain::Cli,
         },
     )?;
     extract_release_archive(&archive_local, stage, &OVERLAY_ARCHIVE_MEMBERS)?;
--- a/crates/mvm-build/src/sdk_sidecar.rs
+++ b/crates/mvm-build/src/sdk_sidecar.rs
@@ -148,6 +148,9 @@ pub fn download_sdk_sidecar(
             asset: &names.archive,
             archive_path: &archive_local,
             version,
+            // Unchanged for the same reason as the runtime overlay: the base
+            // URL is still CLI-version-derived, so the train must match it.
+            train: crate::release_signature::ReleaseTrain::Cli,
         },
     )?;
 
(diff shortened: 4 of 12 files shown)


auser:
## Verified end to end on a real KVM box

The PR said this could not be locally verified — "that needs a real download against a live release with cosign, which no local run can stand in for." It has now been done, on x86_64 Linux with `/dev/kvm`, against the live `boot-image/v0.1.0`, with the **release feature set** (`host,user,template-registry-s3,release-artifact-bootstrap`) so the cosign rung is the one a shipped binary runs.

```
$ mvmctl --version
mvmctl 0.18.0                          # the crate version whose tag does not exist

$ MVM_HOME=/root/mvm-2675-fresh mvmctl kernel build --which workload --kernel-source download
######################################################################## 100.0%     # kernel
######################################################################## 100.0%     # signed checksum manifest
EXIT=0

$ find $MVM_HOME -name 'vmlinux*'
.../cache/builder-vm/x86_64/kernels/workload/vmlinux         20021680 bytes
.../cache/builder-vm/x86_64/kernels/workload/vmlinux.sha256

$ sha256sum .../vmlinux
cb2fd28aba160c4e4db26df315423f5259679146c3d2da1bcdce7ba15bcef0e9
```

That digest is byte-identical to the entry in the published `default-microvm-x86_64-checksums-sha256.txt`. So the whole chain holds: correct tag → correct asset name → redirect followed → signed manifest fetched → **cosign verified under the boot-image identity train** → SHA-256 matched → cached with its digest sidecar.

## It also found a second bug, which is why the run mattered

The first attempt, with only the tag fix, failed:

```
Caused by: HTTP 302 Found downloading
  .../boot-image/v0.1.0/default-microvm-vmlinux-x86_64
```

The URL was already right. GitHub answers every release asset with a 302 to blob storage, and `mvm_http` **does not follow redirects** — deliberately and documented ("no HTTP/2, no redirect following..."), which is true of all its other callers. So this path could never have fetched a release asset, at any URL. The tag fix alone would have shipped and still failed.

All four release fetches in `update.rs` now go through one helper reusing the curl downloader the working release paths already use (`-fSL`). Two of those four are the self-update archive and its cosign bundle — same defect, same file.

## One thing worth knowing for anyone reproducing this

A plain `cargo build --release --bin mvmctl` produces a binary that **cannot** verify release manifests:

```
manifest signature is invalid: manifest-verify feature is disabled in this build
```

The root `mvmctl` package strips sub-crate default features, so the cosign rung only exists when the release feature set is passed explicitly. Verifying this path with a default build tests the wrong binary and gets a confusing refusal.

PR #2748, opened 2026-08-19
fix(stage0): find labelled disks wherever they are attached, and attach each once

auser:
**Every Stage 0 kernel build on the QEMU path failed.** Found by running `machine run --image` on a real KVM box; reproduced on a fresh `MVM_HOME` every time.

```
stage0-init: no disk carries the ext4 label "mvm-nix-store";
             falling back to /dev/vde by enumeration order
stage0-init: FATAL: /dev/vde ... is 32768 bytes, below the 67108864-byte floor
             for a store disk
```

## The store was there the whole time

```
$ blkid /root/.../cache/builder-vm/nix-store-stage0-x86_64.img
  LABEL="mvm-nix-store" UUID="7532f94a-..." TYPE="ext4"      # 68 GB
$ ls -l .../flowmux-identity.ext4
  32768                                                       # the disk it picked
```

Correctly labelled, correctly sized, correctly attached. The guest never looked at it.

## Cause

```rust
const STAGE0_DISK_CANDIDATES: &[&str] = &["/dev/vda", "/dev/vdb", "/dev/vdc", "/dev/vdd"];
```

The QEMU path fills exactly those four with seed, work, out and mvm-bins, then attaches the FlowMux identity image, then the store. So the by-label lookup **stopped two devices short of the disk it was searching for** and could never succeed there.

A fixed list of device letters is precisely the positional assumption the label was introduced to remove — the code's own comment says so:

> The device letter is a function of how many block devices the backend attached ahead of this one, so adding a drive silently re-letters every disk behind it — which is how this guest came to mount a 32 KiB identity image as its Nix store.

That comment describes this bug, written while adding the label that was supposed to fix it.

Both lookups now enumerate every attached virtio disk via `virtio_block_devices()` — which `find_labeled_ext4_disk_among` already documents itself as taking. The `/work` lookup had the identical latent fault and is fixed with it.

**Second, smaller:** the identity image was attached **twice** in `qemu_builder.rs`, consuming another letter and pushing the store one slot further still. `/dev/vde` — the 32 KiB image the guest reported as its store — is that duplicate.

The fail-closed size check behaved correctly throughout. It was reporting a symptom two layers below the cause, which is why it read as a store-sizing problem rather than a lookup one.

## Verified on real KVM

Same box, same command, after the fix:

```
stage0-init: Stage 0 Nix store is /dev/vdf (label "mvm-nix-store")
```

`/dev/vdf` — two past the end of the old list. Stage 0 then proceeds into the real kernel build instead of refusing.

## Tests

`no_disk_is_attached_to_stage0_twice` asserts each disk operand appears in exactly one `-drive`, and explains why a duplicate is not merely wasteful: it re-letters every disk behind it.

`clippy --workspace --all-targets -D warnings` and `fmt --all --check` clean. The pre-commit hook caught a second live use of the constant I had missed on the first attempt, which is why the const is now gone rather than left `#[cfg(test)]`.

### DIFF
--- a/crates/mvm-build/src/bin/stage0-init.rs
+++ b/crates/mvm-build/src/bin/stage0-init.rs
@@ -55,7 +55,6 @@ mod linux {
     /// explicit rather than a `/sys/class/block` walk — a RootDir guest never
     /// has more than a handful of disks, and a missing candidate is just a
     /// failed `File::open` (cheap, non-fatal).
-    const STAGE0_DISK_CANDIDATES: &[&str] = &["/dev/vda", "/dev/vdb", "/dev/vdc", "/dev/vdd"];
 
     const VSOCK_EGRESS_PROXY_URL: &str = mvm_core::guest_netd::DEFAULT_EGRESS_PROXY_URL;
     const VSOCK_EGRESS_NO_PROXY: &str = "127.0.0.1,localhost";
@@ -594,7 +593,12 @@ mod linux {
         // letter stays as a fallback for a store image formatted before it
         // carried a label.
         let label = mvm_build::rootfs::STAGE0_NIX_STORE_EXT4_LABEL;
-        let by_label = find_labeled_ext4_disk(STAGE0_DISK_CANDIDATES, label)
+        // Every attached virtio disk, not a fixed prefix of them. The QEMU path
+        // attaches seed, work, out, mvm-bins and the FlowMux identity ahead of
+        // the store, so a four-entry list stopped two devices short of the disk
+        // it was looking for — the lookup could never succeed there, fell back
+        // to enumeration order, and picked the 32 KiB identity image.
+        let by_label = find_labeled_ext4_disk_among(virtio_block_devices(), label)
             .map(|d| d.to_string_lossy().into_owned());
         let device: &str = match by_label.as_deref() {
             Some(dev) => {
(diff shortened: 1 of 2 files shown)


auser:
## Verified end to end on real KVM

Same box, same command, fresh `MVM_HOME`. Stage 0 now runs to completion instead of refusing:

```
stage0-init: Stage 0 Nix store is /dev/vdf (label "mvm-nix-store")
stage0-init: adopting host-prepopulated Stage 0 Nix store at /dev/vdf
stage0-init: done; halting
[  571.789346] reboot: Power down
```

`/dev/vdf` — two devices past where the old four-entry list stopped looking.

The kernel it was refusing to build now exists:

```
20021680  /root/mvm-boot-fetch/cache/builder-vm/x86_64/kernels/workload/vmlinux
```

No error lines anywhere in the run. Note `done; halting` specifically: the earlier failure was *"QEMU Stage 0 exited before reaching a clean halt"*, and that clean halt is exactly what the store refusal was preventing.

Before the fix, on the identical setup:

```
stage0-init: no disk carries the ext4 label "mvm-nix-store";
             falling back to /dev/vde by enumeration order
stage0-init: FATAL: /dev/vde ... is 32768 bytes, below the 67108864-byte floor
```

PR #2755, opened 2026-08-20
fix(ci): compile the release bootstrap feature

auser:
## Summary

- remove the stale CLI-version binding from the builder boot-image downloader
- compile the dormant release-artifact bootstrap feature with warnings denied in pull-request CI
- record Security-lane recovery as pending a clean final workflow run

## Verification

- release feature check reproduced the unused-variable failure before the fix
- `RUSTFLAGS="-D warnings" ./scripts/cargo-stable.sh check -p mvm-cli --features release-artifact-bootstrap --lib`
- `./scripts/cargo-stable.sh clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` completed all test targets; its mvm-cli doctest collided with the suite's nested Cargo process and passed on the immediate isolated rerun
- `cargo fmt --all --check`
- `actionlint .github/workflows/ci.yml`
- `cargo run -p xtask -- check-sprint-append`

Refs #2736

PR #2762, opened 2026-08-20
fix(security): contain arrayref registry incident

auser:
## Summary

- pin `arrayref` to the exact official upstream revision that produced reviewed release 0.3.9
- apply the pin to the root and every standalone fuzz workspace
- refresh fuzz lockfiles and require CI to consume them with `--locked`
- admit only the exact upstream repository in the source policy and add synchronization regressions

The registry yanked the established release line and published 0.3.10 with an unavailable `proc-macro1` dependency. This caused both cargo-deny and the merge-queue invariant job to fail. No 0.3.10 code enters this graph.

## Validation

- `cargo deny check`
- `cargo fmt --all --check`
- `cargo test --test supply_chain_dependency_pin` (3 passed)
- `cargo check --locked --manifest-path <each fuzz workspace> --all-targets` (10/10 passed)
- `cargo test --workspace` (all substantive groups passed; final nested-Cargo sync test hit contention, then passed in isolation)
- `cargo test -p xtask --test check_sync -- --nocapture` (passed)


auser:
Superseded by #2755 plus #2763. This branch duplicates the fuzz lock/workflow work already present in #2755 and adds source patches to standalone graphs that do not contain arrayref. #2763 is stacked on the current #2755 head, patches only the root and four actually affected graphs, and adds a discovery-based regression test that fails if a future affected graph is missed. Closing this duplicate PR; issue #2736 remains open until the complete Security workflow is green on current main.

auser:
Reopened to keep this work active for explicit verification and resolution.

PR #2788, opened 2026-08-21
fix(oci): resolve a workload's environment once, for both ways in

auser:
## The report

```
$ mvmctl machine run --image rust:latest --env "NAME=ari" -it -- /bin/bash
I have no name!@(none):~$ which rustc
I have no name!@(none):~$ which ld
/usr/bin/ld
```

The image *was* being used — `/usr/bin` carried the full Debian toolchain `rust:latest` ships. What was not being used was its **declared environment**. `rust:latest` installs its toolchain under `/usr/local/cargo/bin` and puts it on `PATH` through the image config alone (`PATH`, `CARGO_HOME`, `RUSTUP_HOME`).

Three defects stacked into that one prompt.

## 1. Two paths into a workload, one of which resolved the environment

Materialization parsed the image's `Env`/`WorkingDir` and wrote them into the rootfs. Exactly one binary read them back — `mvm-oci-entrypoint`, on the `--entrypoint` path. `-it` takes `ConsoleOpen` instead, where `console.rs::build_shell_env_from` assembled its own block from the guest agent's process vars plus `--env` pairs. It never opened the file.

New `mvm-agentd::workload_env` is the single resolver. `ImageRuntimeConfig` is now defined by the crate that *reads* it, so writer and reader cannot drift. A `WorkloadEnvironment` builder fixes precedence: default `PATH` floor → inherited vars → image declaration → `--env` overrides → forced `HOME`, with `TERM` only when interactive.

Layers **replace in place**. `execve` accepts a block with a repeated name and `getenv` answers with the first, so the old append made an override of an already-present name a silent no-op — the old code worked around this for `HOME`/`TERM` by dropping inherited copies by name, and nothing else was covered.

The console now also honours the image's `WorkingDir`, matching where its own entrypoint would start.

## 2. An image declaring `Env` and no argv lost both

`oci_entrypoint_from_config_bytes` returned `None` on an empty argv; `inject_mvm_runtime` gated the write on `!argv.is_empty()`. The config is now written whenever the image declares anything. The entrypoint *marker* stays gated on there being something to run.

## 3. `I have no name!` and an empty `~`

`provision_workload_identity` sits behind `optional_image_writes_allowed(rootfs_is_read_only())`, and **every** workload rootfs is mounted `MS_RDONLY` (`guest_mount::mount_rootfs`, plus `root=/dev/vda ro` on every backend's cmdline). So on the OCI path both writes always no-opped and `workload_home()` fell back to `/tmp` — which is why the shell started in an empty directory.

`mk-guest.nix` already bakes `/etc/passwd`, `/etc/group`, and `/home/mvm-worker` at build time for the images mvm builds itself (`nix/lib/mk-guest.nix:1232-1248`). The OCI path never got the equivalent. It does now, reusing `workload_identity::provision_in` — the same function, at the one point in the lifecycle where the rootfs is still writable. Appends only; never over an entry the image already claims.

For the home, `mount_workload_home` lays a tmpfs (`uid=901,gid=901`) over a baked-in mount point after the pivot and before the privilege drop. Mounting over a directory needs no write to the underlying filesystem, so this holds on a sealed dm-verity root exactly as on a plain one. `workload_home()` now reports what boot actually secured: on a failed mount it returns the `/tmp` fallback rather than the read-only mount point the `is_dir` probe would have named.

## Renamed

`/etc/mvm/oci-entrypoint.json` → `/etc/mvm/image-runtime.json`; the file is no longer an entrypoint concern. It is folded into `INJECT_DESTS`, so the content digest changes and already-materialized images re-materialize rather than booting against a name nothing writes.

## The gate

`xtask check-single-workload-env` pins the invariant the bug violated: the materialized config has exactly one reader, and every declared consumer resolves through `WorkloadEnvironment::builder()`. Wired into `ci.yml`. Both arms were confirmed **red** against planted violations before being called done.

## Evidence

- `cargo nextest run --workspace --exclude mvmctl`: 12635 passed, 3 failed. Doctests clean. Nightly `fmt --all --check` clean.
- Those 3, plus 24 in the root package, are one pre-existing **local-environment** fault — `CARGO_BIN_EXE_*` unset, so `assert_cmd` cannot find the binary under test. Every one reproduces on clean `main` at `ee6b98d050`; verified directly rather than assumed. Nothing here is implicated, and nothing here is covered by them either.
- The pre-commit hook ran `cargo clippy --workspace --all-targets -- -D warnings` on the repo's **pinned stable** toolchain and passed clean.
- `just check-gated` (zigbuild, `x86_64-unknown-linux-gnu`, `--all-targets`) clean — the Linux-gated arms of `mount_workload_home` and `guest_bootstrap` are invisible to a macOS check.
- Mutation-checked, not just green: reverting the builder to append reddens `later_layers_replace_earlier_ones_rather_than_shadowing_them` and `the_console_and_the_entrypoint_agree_on_what_the_image_declared`; dropping `.image()` from the console reddens `console_takes_path_from_the_image_over_the_agents_own` and `console_starts_in_the_images_working_dir_when_it_declares_one`.

## Deliberately not done

The `(none)` hostname. The machine name never reaches the guest, so setting it needs a new `mvm.hostname=` kernel-cmdline parameter across every backend's boot-args construction plus the validator. `(none)` is what Nix-built images report too, so this is not an OCI-path regression. Filed separately.

Not verified on real hardware: the end-to-end proof is `mvmctl machine run --image rust:latest -it -- /bin/bash` reporting a `rustc` on `PATH` and a resolvable user, which needs a host with a working backend.


auser:

PR #2791, opened 2026-08-21
fix(security): pack mutation shards by cost, and let the watcher report a timeout

### LINKED ISSUE #2736: Security lane is red
The `Security` workflow (schedule) failed on `49053826bac5b2c349a73159a8ffad802c10638a`.

Run: https://github.com/tinylabscom/mvm/actions/runs/32217229444

Failing or unfinished jobs:

- Supply chain — cargo-audit (ADR-001 §W5.2) — [log](https://github.com/tinylabscom/mvm/actions/runs/32217229444/job/95960907611)
- Claim witnesses — mutation-tested (mvm-hostd/2of4) — [log](https://github.com/tinylabscom/mvm/actions/runs/32217229444/job/95960907651)

These lanes do not run on pull requests, so a failure here is
invisible until someone reads run history. Several of them are the
only witness for a numbered claim in `specs/adrs/001-microvm-security-posture.md`
— while a lane is red, that claim has no live evidence behind it.

auser:
This closes two gaps that made the scheduled Security lane incomplete or silent, and now also includes the mutation witnesses exposed by the first fresh scheduled run.

## Cost-aware mutation shards

The old path-stride assignment split files by count rather than mutation cost. Measured mvm-hostd shards were 90/323/24/324 minutes, with the two expensive shards repeatedly timing out. The new longest-first packing uses source size as the stable cost proxy and expands mvm-hostd to six shards. Replaying the measured costs produces a near-even 188/189/191/193 split before the additional width.

The balance test asserts the ratio rather than a wall-clock threshold and fails against the old assignment.

## Timeout-aware watcher verdict

The watcher treated a timed-out job as a failure, but an earlier run-level cancellation guard returned before inspecting jobs. The verdict now skips only when no job reached a verdict at all. A shell fixture suite covers success, failure, timeout, and operator-cancel shapes and is wired into the always-on lint-policy lane.

## Scheduled-run mutation repair

Security run 32552650847 then exposed seven surviving authorization mutants in extension verification, budget narrowing, guest attachment, artifact limits, and assurance proxy collision handling. Focused one-field-at-a-time and boundary tests now pin every fail-closed predicate. The obsolete accepted miss for host-budget multiplication was removed from the mutation baseline.

A bounded cargo-mutants proof over the affected predicates caught 20 of 20 generated mutants, including all seven scheduled-run survivors and adjacent provenance-deletion mutations.

## Verification

- cargo test --workspace: green on a complete retry; the first attempt had one transient nested compiler process failure, and cargo test --workspace --no-run plus the complete retry both passed
- cargo clippy --workspace --all-targets -- -D warnings: green
- cargo check --workspace: green
- just check-gated: green for Linux-gated and BDD targets
- cargo fmt --all -- --check: green
- cargo run -p xtask -- check-mutation-witnesses: 42 files, 176 accepted misses, clean
- cargo run -p xtask -- check-sprint-append, check-plan-names, and check-no-spec-refs-in-comments: clean
- bounded cargo-mutants proof: 20 caught, 0 missed

Closes #2736


auser:

PR #2793, opened 2026-08-21
ci: run the security lanes that never ran, delete the workflow they hid in

auser:
> **Stacked on #2791.** Base is `fix/security-lane-coverage`; retarget to `main` once that merges. Review the second commit only.

`ci-full.yml` had **never been run** — `total_count: 0` against the Actions API — and eleven of its eighteen jobs carried `if: github.event_name == 'workflow_dispatch'`, a redundant guard on its only trigger.

That mattered more than dead YAML usually does: ADR-001's ledger cites `ci:app-deps-audit` as claim 11's witness, and that job lived there. It is the only `ci:` witness in the table not already in `security.yml`. The row has been green for as long as it has existed, backed by a lane that never once executed.

## Revived

| lane | was | now |
| --- | --- | --- |
| `app-deps-audit` | `ci-full.yml`, dispatch-only, **0 runs** | `security.yml`, nightly |
| six `oci-*` lanes | `ci-full.yml`, dispatch-only, **0 runs** | `security.yml` `oci-hardening` matrix, nightly |
| `pack-signing-smoke.yml` | dispatch-only, **0 runs** | own nightly cron |

`app-deps-audit` keeps its exact job name — ADR-001 and `model/claims.toml` both match on it.

The six OCI lanes differed only in the test selector, so they are one matrix. Their `dorny/paths-filter` steps are dropped rather than ported: each gated its work on `steps.filter.outputs.oci == 'true' || github.ref == 'refs/heads/main'`, and the second arm is always true on these triggers, so the filter decided nothing. ADR-001 cites none of the six — CLAUDE.md's prose named them, and now says explicitly that the ledger does not.

`pack-signing-smoke.yml` gets its own trigger rather than a `workflow_call` from `security.yml`: it reconstructs its cosign OIDC identity from `.../pack-signing-smoke.yml@${REF}`, and the SAN semantics under `workflow_call` are not obviously identical. Not worth guessing at on a lane about to run for the first time. `security-lane-watch.yml` now watches it, keyed on workflow name so two lanes cannot overwrite each other's tracking issue.

## Deleted

- **`ci-full.yml`** — what remained after the move duplicated `ci.yml` (`lint`, `test`, `nix-flake-check`, 17 xtask gates).
- **`windows.yml`** — the only workflow in the tree with no consumer at all.

Two consumers had to move first, and neither fails loudly: `tests/github_actions_fuzz_gate.rs` `include_str!`s `ci-full.yml`, so deleting it is a **compile** error, and `scripts/check-fast-cargo.sh` required two exact strings from it.

## Kept, despite never running

My first pass called six workflows dead on run count alone. Five were load-bearing, and acting on that would have deleted release machinery and a security lane:

- `publish-crates.yml`, `kernel-build.yml` — dispatched by `release.yml` (lines 657, 678). Zero runs because `release.yml` has itself run once.
- `pack-signing-smoke.yml` — a security lane; revived above.
- `merge-queue-requeue.yml` — disabled at the GitHub level, not in YAML; has a test.
- `architecture.yml` — read by `xtask/src/check_workflow_paths.rs`.

Run count is evidence a lane is not running. It is not evidence nothing depends on it.

## Duplicate required-check name

`ci.yml:kernel` and `kernel-build.yml:kernel` both declared `name: Build kernels (${{ matrix.arch }})` — a required status check — from the same build script. I called this a merge-queue hazard when I first found it; more precisely it is **latent**, resolving unambiguously only because the triggers are disjoint (`pull_request`/`merge_group` vs tags/dispatch), which nothing in the name states. The tag-time publisher is now `Publish kernels (<arch>)`, pinned by `only_one_workflow_claims_the_required_kernel_check_name` beside the existing test that pins the triggers apart.

## Still open

`security.yml` runs only on tags and the nightly, never on a PR, so none of its 21 jobs gate a merge — the watcher and its tracking issue are the whole enforcement surface. Deliberate, but it is why the two defects in #2791 went unnoticed for a fortnight. Worth revisiting separately.

## Verification

- `cargo test -p xtask` (647), plus `github_actions_fuzz_gate`, `github_actions_bdd_gate`, `ci_scope_aggregate`, `release_assets`, `merge_queue_requeue_workflow` — all pass
- New name-collision gate verified red against the old name, then green
- 12 xtask gates, `bash scripts/check-fast-cargo.sh`, `actionlint`, `shellcheck`, `cargo fmt --all --check`, workspace `clippy -D warnings` — clean


auser:
## Correction pushed — I had deleted two security lanes

The first commit on this branch deleted `ci-full.yml` outright, having moved only the seven claim-adjacent lanes. That dropped:

- **`sealed-prod-allowlist`** — the SealedProd vsock allowlist (plan 76 Phase 1), via `scripts/check-sealed-prod-allowlist.sh`
- **`jailer-property`** — the seccomp and Landlock property suites, the confinement primitives behind claims 1 and 2

and, incidentally, **every macOS lane in the repository** — `apple` and `libkrun-macos` were the only two, for a tool whose primary host is macOS.

The reasoning was "this workflow has never run, therefore it is dead." The run count was real (`total_count: 0`), but the conclusion doesn't follow for a security lane. Never having run is a reason to *start* running one.

### What changed

Both security lanes now live in `security.yml`, beside the other claim witnesses and inside what `security-lane-watch.yml` reports on. `jailer-property` keeps its Landlock probe rather than asserting the LSM — both suites share `confine_self` and Landlock applies first — but the probe now writes to the step summary, and the silent-skip risk is named in the comment instead of left implicit.

`ci-full.yml` survives as **`Extended CI`** on a nightly cron. Only `lint`, `test` and `nix-flake-check` are gone, each duplicated by `ci.yml` on every PR. Everything else had no other home:

| lane | why it stays |
| --- | --- |
| `apple`, `libkrun-macos` | the only macOS coverage in the repo |
| `workload-spawn-smoke-linux` | the only live Firecracker boot smoke |
| `builder-vm-image-linux` | builder-VM image build |
| `e2e` | `smoke_run_json_receipt`, `runtime_boot` |
| `sdk-release-dry-run` | dry-run of the SDK publish path |

All were guarded on `event_name == 'workflow_dispatch'`, and `apple` additionally required a release tag, so none could fire on a cron. Those guards are removed. The watcher covers `Extended CI` too — a nightly nobody reads is the defect this branch exists to fix, and it shouldn't be reintroduced one workflow to the left.

### Full disposition — zero lanes lost

```
lint                          duplicate of ci.yml (every PR) — dropped
test                          duplicate of ci.yml (every PR) — dropped
nix-flake-check               duplicate of ci.yml (every PR) — dropped
sealed-prod-allowlist         MOVED -> security.yml (nightly, watched)
jailer-property               MOVED -> security.yml (nightly, watched)
app-deps-audit                MOVED -> security.yml (nightly, watched)
oci-* (x6)                    MOVED -> security.yml oci-hardening matrix
e2e                           kept in Extended CI (now nightly)
sdk-release-dry-run           kept in Extended CI (now nightly)
builder-vm-image-linux        kept in Extended CI (now nightly)
apple                         kept in Extended CI (now nightly)
libkrun-macos                 kept in Extended CI (now nightly)
workload-spawn-smoke-linux    kept in Extended CI (now nightly)
```

`windows.yml` is still deleted — it remains the only workflow in the tree with no consumer of any kind.

### Gate

`security_lanes_live_where_the_watcher_reads_them` pins all four security lanes to `security.yml` and asserts every watched workflow has a `schedule:`. Verified red against the exact deletion it describes:

```
sealed-prod-allowlist: must live in security.yml — that is the workflow
security-lane-watch.yml watches, and these lanes back numbered claims in ADR-001
```

Net across the branch: **zero lanes lost, eighteen that had never run now run nightly.**

PR #2794, opened 2026-08-21
perf(vsock): send sealed control frames as binary instead of JSON

### LINKED ISSUE #2790: vsock: control plane JSON-wraps every sealed frame, costing ~705us and 3.6x bytes per frame
`AuthenticatedSession::write` wraps every sealed control frame in JSON, re-encoding the ciphertext `Vec<u8>` as an array of decimal numbers. `SealedFrame::encode` (`crates/mvm-core/src/net/session.rs:478`) already defines a compact binary layout for the same struct, and the FlowMux egress plane already uses it through `write_sealed_frame` / `read_sealed_frame`. Only the control plane pays the JSON tax.

## Measured

One 15872-byte stdout chunk (the current `MAX_DATA_CHUNK_SIZE`), release build, aarch64 macOS — same sealed frame, two encodings:

| | wire bytes | encode | decode |
|---|---|---|---|
| JSON envelope | 184050 | 212 us | 492 us |
| `SealedFrame::encode` | 51664 | 0.9 us | 0.9 us |
| ratio | 3.6x | **244x** | **540x** |

Bytes are the small half. The cost is ~705 us of serde_json per frame, so 900 KB of guest stdout burns roughly **40 ms of CPU across 57 frames** and puts 10.5 MB on the wire instead of 2.9 MB. Against a dispatch window targeted under 200 ms that is worth reclaiming.

## Scope

- `AuthenticatedSession::write` / `read` move to `write_sealed_frame` / `read_sealed_frame`, which are already generic over `Write` / `Read` and already carry the `[u32 BE len][payload]` framing.
- `MAX_DATA_CHUNK_SIZE` re-derives against one JSON hop instead of two. The expansion constant drops from 16 to 4, so the cap rises from 15.5 KiB toward ~62 KiB.
- The claim-5 fuzz surface moves with the ingress. `fuzz_authenticated_frame.rs` currently fuzzes `serde_json::from_slice::<AuthenticatedFrame>`, which stops being what the control plane parses; it has to fuzz `SealedFrame::decode` or the claim quietly loses its witness. `fuzz_authed_path.rs` needs the same treatment.
- A committed benchmark keeps the regression from returning silently.

## What this does not fix

Hop 1 survives. The inner `GuestResponse` JSON still encodes `chunk: Vec<u8>` as a number array — 51.5 KB for 15.9 KB of content, about 3.3x. Removing that means changing the body representation, which is a larger and separate change. Worth re-measuring after this lands rather than assuming.

## Compatibility

This changes the host<->guest control wire. Host and guest agent are built together and the OCI rootfs cache is keyed by guest-agent source hash, so a stale guest cannot pair with a new host. The repo's stated position is no backwards compatibility ahead of a first stable version, so no shim.

Follow-up to #2780, which capped the chunk against the wire this issue removes.

auser:
Closes #2790. **Stacked on #2784** — branched from it because both change the same lines. Review the top three commits; rebase onto `main` follows once #2784 lands.

`AuthenticatedSession` wrapped every sealed control frame in JSON, re-encoding the ciphertext `Vec<u8>` as an array of decimal numbers. The egress plane already used the compact `SealedFrame::encode` for the same struct; only the control plane paid for a second encoding.

## The number, and the guess it replaced

I first argued for this on bandwidth — ~10x amplification, so maybe 5–10 ms on a large transfer. That was wrong by two orders of magnitude, and wrong in the direction that would have made the change look not worth doing.

One 15872-byte stdout chunk, release, aarch64 macOS:

| | wire bytes | encode | decode |
|---|---|---|---|
| JSON envelope | 184050 | 212 us | 492 us |
| `SealedFrame::encode` | 51664 | 0.9 us | 0.9 us |
| ratio | 3.6x | **244x** | **540x** |

End-to-end on HVF, 900 KB of guest stdout, guest command phase, two runs each:

| | run 1 | run 2 |
|---|---|---|
| JSON envelope, 15.5 KiB chunks | 69.6 ms | 75.4 ms |
| binary envelope, 62 KiB chunks | 34.4 ms | 33.3 ms |

Both variables moved together — the cap rises *because* the envelope changed — so ~38 ms is the combined effect, not the envelope alone.

## What changed

`write`/`read` use `write_sealed_frame`/`read_sealed_frame`, already generic over `Write`/`Read` and already carrying the `[u32 BE len]` framing. No new framing code. Content now crosses one JSON hop instead of two, so the chunk cap re-derives from 15.5 KiB to 62 KiB and the `const` assertion moves with it.

**The outbound size check moved earlier and is stricter for it.** `write_frame` enforced the cap *after* sealing, so an oversize frame also spent a sequence number and poisoned the session on its way to being rejected. The check now runs on the plaintext before `seal`, so an oversize payload is refused with the session still usable — the test asserts the peer is still in step afterwards rather than asserting the session died.

That was a self-inflicted regression first: `write_sealed_frame` has no size check, so my first version silently dropped the outbound cap. It was caught by a test failing for the opposite reason it was written for.

## The claim-5 fuzz witness moved with the ingress

`fuzz_authenticated_frame` fuzzed `serde_json::from_slice::<AuthenticatedFrame>`, which stopped being what the control plane parses. It is now `fuzz_sealed_frame` against `SealedFrame::decode`, the parser an unauthenticated peer actually reaches.

That rename touched **seven** coupled files — the fuzz `Cargo.toml`, `security.yml`, the fuzz README, ADR-001's ledger row *and* its prose, `CONFORMANCE.md`, and `model/claims.toml`. The last was found by sweeping for the old name, not by knowing about it.

## Deliberately not done — and it leaves a real gap

`fuzz_authed_path` and the legacy `write_authenticated_frame` / `read_authenticated_frame` / `verify_authenticated_frame` trio are untouched. They have **no production callers** — only their own tests and that fuzz target. Before this change that mattered less, because `AuthenticatedFrame` was the shape the control plane really used. It is not now, so **claim 5 currently has a witness aimed at a path production does not take.**

Deleting the trio takes the target with it, and rebuilding it around `Session::open` — version, session id, replay, signer, signature, decrypt — deserves its own change. Deleting without rebuilding would trade a stale witness for a missing one. Recorded in the plan and the delivery note; happy to do it next.

## Verification

Gates: nightly `cargo fmt --all --check`, `clippy --workspace --all-targets -D warnings`, `nextest run --workspace`, `test --workspace --doc`, `check-gated` (zigbuild `x86_64-unknown-linux-gnu --all-targets` + `mvm-conformance --features bdd`), all 63 `xtask` gates.

End-to-end on HVF with a rebuilt guest agent (`MVM_EMBED_NO_CACHE=1`, since the embedded binaries were stale and would otherwise have booted the old agent and made a pass meaningless):

| single-burst stdout | result |
|---|---|
| 28000 B | byte-exact |
| 100000 B | byte-exact |
| 900000 B | byte-exact |

Plan: `specs/plans/2026-08-21-binary-sealed-frame-envelope.md`.


auser:

PR #2795, opened 2026-08-21
fix(mount): close the guest-mount shadowing hole and de-duplicate the policy

auser:
## The bug

The mount allow-list admits `/mnt` while reserving `/mnt/config` and `/mnt/secrets`. Those two checks disagree:

- the reserved check never fires, because a share at `/mnt` doesn't name a reserved path;
- the allow-root check passes, because `/mnt` *is* an allow-root.

So a share mounts over the parent and hides both drives. Depending on mount order it's worse than hiding: if the share mounts first, the secret drive lands *inside* a host-shared directory.

## The fix

Containment was checked in one direction only — a mountpoint was refused when it sat *under* a protected path. Shadowing is the other direction: a protected path sitting *under* the mountpoint. Nothing checked it.

Any reserved path beneath an allow-root has this shape, so checking both directions closes the class rather than the instance.

Behaviour:

| Path | Before | After |
|---|---|---|
| `/mnt` | admitted | refused, naming the drive it would hide |
| `/mnt/config`, `/mnt/secrets` | refused | refused |
| `/mnt/config/nested` | refused | refused |
| `/mnt/configx` | allowed | allowed |
| `/mnt/sub`, `/data`, `/work` | allowed | allowed |

**The allow-list is unchanged** — `/data`, `/work`, `/mnt` and nothing else. This is not a widening.

## Second copy of the policy removed

`mvm-cli` carried its own `ALLOWED_GUEST_MOUNT_ROOTS` + `RESERVED_UNDER_ALLOWED`, mirroring `mvm-core`'s `MountPathPolicy`, with a comment saying they had to be kept in sync by hand. That is how the CLI and the agent came to disagree about `/mnt` in the first place. The CLI now calls the one policy in `mvm-core`; `parse.rs` loses ~60 lines of duplicated constants and their tests.

## Witnesses

Five tests across `mvm-core` and `mvm-cli` fail if `shadowed_protected` is removed — I checked by deleting it, not by inspection. `bare_mnt_is_refused_because_it_shadows_the_config_drive` joins claim 1's witness set in `model/claims.toml` and ADR-001's ledger. `the_ancestor_rule_is_segment_aware` pins the boundary behaviour so the new direction can't over-refuse siblings.

## Verification

- `cargo nextest run --workspace` — 12766 passed. 108 failures are pre-existing and reproduce identically on this branch with the changes stashed: 105 `CARGO_BIN_EXE_mvmctl is unset` (harness/env) and 3 `mvm-hostd::extension_provider_process`.
- `cargo test --workspace --doc` — clean.
- `just check-gated` — clean (covers the `required-features` and Linux-gated targets `--all-targets` misses).
- All 62 workflow-derived xtask gates — clean. `check-claim-catalog` 18 claims / 92 witnesses; `check-mutation-witnesses` re-pinned for the moved surface.
- Stable `clippy --workspace --all-targets -- -D warnings` — clean.

## History

This PR previously proposed replacing the allow-list with a per-tier protected-path policy that widened mountable paths for OCI images. That was dropped in favour of keeping the strict allow-list; what remains is the shadowing fix and the de-duplication, both of which apply to the allow-list as it stands. No tiers, no sidecar field, no wire-format change — `schema/protocol-v0.json` and the SDK stubs are untouched.


auser:

Issue #2792, opened 2026-08-21
Security lane watch did not fire for a completed nightly run


auser:
`security-lane-watch.yml` triggers on `workflow_run: [Security] types: [completed]`. Security run [32448509693](https://github.com/tinylabscom/mvm/actions/runs/32448509693) (event `schedule`) completed at `2026-08-21T11:36:49Z` with conclusion `cancelled`, and **no watcher run was created for it**.

Evidence:

- `gh run list --workflow=security-lane-watch.yml` shows nothing after `2026-08-21T04:53:22Z`.
- Querying every repo workflow run with `updated_at` between 11:00 and 13:00 on 2026-08-21 returns exactly one row: the Security run itself.

Cancellation does not explain it. Several earlier Security runs concluded `cancelled` (cancel-in-progress evicting a dispatch when the nightly started) and each *did* produce a watcher run — those show as `skipped`, correctly, because the job-level `if:` only processes `schedule` and `push`. The 04:53:22 watcher run is one of them, fired by the 04:16 dispatch run being evicted. So the event is normally delivered for cancelled runs; this one specifically was not.

Consequence: the watcher is the sole enforcement surface for lanes that never run on a PR. A nightly that finishes red and delivers no event is unreported, and #2736 stayed stale as a result.

#2791 fixes a separate, provable defect in the same workflow — a guard that discarded the per-job verdict whenever the run concluded `cancelled` — so a *delivered* event now reports correctly. It does not explain an event that never arrived, and the two should not be conflated.

Worth checking:

- whether the watcher's `concurrency: {group: ${{ github.workflow }}, cancel-in-progress: false}` can drop a delivery rather than queue it, given several Security runs completed in quick succession that day
- whether GitHub suppressed the event, in which case the watcher cannot be the only backstop and `claim-witness-freshness.yml` should also assert the last Security run *concluded*, not merely that it ran

Do not close this on the strength of #2791 merging.

PR #2796, opened 2026-08-21
ci: pin every third-party action, and run the policy gates from one table

auser:
> **Stacked on #2793**, which is stacked on #2791. Base is `ci/workflow-consolidation`; retarget down the chain as each merges. Review the third commit only.

The remaining consolidation items from the workflow survey.

## Every third-party action floated on a branch

All **18** `nix-installer-action` uses were `@main` — including `release.yml` and `release-boot-image.yml`, the lanes that build and sign what we publish. At the time of the change `main` was **43 commits ahead of the newest release**: 43 commits nobody here had read, executing on the signing path, re-resolved on every run.

All 18 now pin `ef8a148…` (v22) by sha, with the version in a trailing comment.

`dtolnay/rust-toolchain@master` stays as the one named exception. That action publishes no version tags — the channel *is* the branch name — and `@master` is its documented entry point when the toolchain arrives as an input, which is how the lanes pinning a nightly *date* have to call it. Floating the action to pin the compiler is the trade its design forces.

`check-workflow-paths` now rejects a third-party `@main`/`@master`, and rejects one action pinned to two versions.

**Writing that check surfaced a bug in its own first draft.** The parser matched `uses:` but not the more common `- uses:` form, so it saw 134 refs where there were 206 — and would have read a floating ref as clean. The unit test caught it before the gate shipped; the count going 134 → 206 is the tell.

The two-version half then found a real split: `publish-npm.yml` and `publish-pypi.yml` paired `upload-artifact@v4` with `download-artifact@v4` against eleven lanes on v7 — one major bump from a handoff that fails only on the release path. Unified to v7/v8.

## Sixty gate steps became one

`ci.yml`'s lint-policy ran 60 consecutive `cargo run -p xtask -- check-<name>` steps. They ran serially and **stopped at the first failure**, so a branch breaking four gates was found, fixed and re-pushed four times, each round trip paying a full lane. And the list lived only in the YAML — which is why more than one session's notes carry the instruction "derive the gate list by grepping the workflows".

`xtask check-all` owns the list now, `main.rs` dispatches individual gates from the same table, and it reports **every** failure. Its first run against this branch reported three at once, which is the behaviour change demonstrated rather than asserted. `lint-policy` goes 75 steps → 17.

### Three gates were running nowhere

`check-backend-resource-controls`, `check-single-fixture-corpus`, `check-single-grants-projection` were reachable as subcommands and invoked by **no workflow at all**. All three pass on a clean tree — they were never wired in, not excluded for failing. They are in the table now, and `every_dispatched_gate_is_in_the_lane_or_excluded_with_a_reason` makes the omission unrepeatable: a gate must be in the table or on an exclusion list stating why. `check-kernel-config-budget` is excluded because it takes a path to a resolved kernel `.config`; the other six name the lane they run in.

### The ledger had to learn about the driver

Collapsing named steps broke `ci:` witness resolution — ADR-001 cites `ci:check-abi-layout`, which anchored on the parenthesised token in that step's name. `claims_ledger::ci_anchors` now anchors the whole table when it sees the driver invoked, so the witness resolves for the reason it always did: the gate runs in CI. **Removing a gate from the table immediately unbacks its claims**, verified by doing it.

That also retired a duplicate: `check_conformance` had its own `ci:` resolver doing `text.contains(name)` over workflow text — the weak form `ci_anchors` was explicitly written to replace, which matched a name in a comment as readily as a live job. It routes through the shared one now.

## Deliberately not changed

The raw `actions/cache@v5` blocks in `security.yml` and `ci-full.yml` are **not** drift against `./.github/actions/rust-cache`, and I listed them as such in the original survey. That was wrong. The composite is restore-only with a single writer on `main`, by a reasoned policy about GitHub's per-ref cache scoping and the 10 GB LRU cap; the nightly lanes want the opposite — their own per-package caches, written where they run. Two strategies for two kinds of lane, not one applied inconsistently.

## Verification

- `cargo run -p xtask -- check-all` — 63 gates clean
- `cargo test -p xtask` — 653 pass, incl. 5 new tests
- Both new gates verified red against the defect each describes, then green
- `github_actions_fuzz_gate`, `github_actions_bdd_gate`, `ci_scope_aggregate`, `release_assets`, `merge_queue_requeue_workflow` — pass
- `actionlint`, `cargo fmt --all --check`, workspace `clippy -D warnings` — clean


auser:
## Added: a backstop for the verdict that never arrives (#2792)

`security-lane-watch.yml` fires on `workflow_run: [Security] completed`. On 2026-08-21 the nightly completed at 11:36:49Z and **no watcher run was created for it** — no repo workflow run exists in that window at all. Other completions, cancelled ones included, did trigger it, so cancellation is not the cause. #2792 tracks the mechanism, still unknown.

An undelivered event produces no run, no verdict and no trace — which reads exactly like a green night. That is the same silence the five-week Security blackout produced, reached from the other direction, and it means the enforcement surface cannot be one event-driven watcher.

`check-claim-witness-freshness` already runs on **its own schedule** rather than on `workflow_run`, precisely so it can notice absence. It asked "did the lane run?"; it now also asks "did the thing that reports on the lane run?".

That is deliberately *not* the duplication its module header declines. The header refuses to re-check **conclusions**, because the watcher owns that verdict and two gates on one property can disagree. Whether a verdict was produced *at all* is a different property with no other owner.

Run against the live repository it reproduces the incident:

```
security.yml completed at 2026-08-21T11:36:49Z and security-lane-watch.yml
never ran for it — claim(s) [1, 3, 4, 5, 6, 7, 10, 11, 15] had no verdict
reported either way
```

### Off for pull requests, on purpose

Behind `--check-reporting`, which the workflow passes on every trigger except `pull_request`. The two questions have different audiences:

- **A lane stopped firing** can be caused by the diff under review — a cron edit dropping a lane out of scope is what the PR trigger exists to catch. Still fails a PR.
- **A completed run nobody reported on** is history. No PR author can fix it, and reddening their branch for it is how a gate teaches people to ignore it.

Without that split this would have turned every `security.yml`-touching PR red until a healthy nightly, for a historical incident.

### A note on how it was caught

The pre-commit hook rejected my first version: I anchored a test insertion on a `fn` line, so the *existing* `#[test]` bound to my new block and `first_created_at_reads_the_newest_entry` silently stopped being a test. `cargo test` reported 656 passing either way — the count went up, so nothing looked wrong. Only `-D warnings` (`duplicate-macro-attributes` + `dead-code`) surfaced it. Both tests run now.

### Limits

It detects that no watcher run followed a completion. It cannot explain why the event was not delivered, and cannot force delivery. If #2792 proves recurring rather than a one-off, the watcher stops being a sound design and the verdict wants pulling into the lane itself — a bigger change that should wait for a second occurrence.

PR #2798, opened 2026-08-22
test(claim5): fuzz the post-auth path production actually takes

### LINKED ISSUE #2797: claim 5: fuzz_authed_path witnesses a path production no longer takes
#2794 moved the control plane to the binary sealed envelope. A consequence it recorded but did not fix: `fuzz_authed_path` — a claim-5 witness — drives `verify_authenticated_frame`, a signature-only path over the *old* JSON envelope with no production callers. The claim keeps a witness aimed at code nothing takes.

## Scope

- Retarget `fuzz_authed_path` to `Session::open`, the real post-auth path: protocol version, session id, sequence (gap and replay), signer identity, Ed25519 signature, AES-GCM decryption.
- Delete `write_authenticated_frame` / `read_authenticated_frame` / `verify_authenticated_frame` and the `AuthenticatedFrame` type once nothing takes them.
- Fix the doc comments that describe the control plane as JSON-enveloped. There are six, across `mvm-core`, `mvm-hostd`, `mvm-build` and two fuzz targets.

## The part worth care

`Session::open`'s checks are not independent. The Ed25519 signature covers the frame context — session id, sequence, timestamp, signer id — so tampering any of those is caught by the signature, not by the check named for that field. A target that perturbs a sealed frame's `sequence` and asserts `Err` therefore passes whether or not the replay check exists.

Isolating a check means producing a frame that is *validly signed* and wrong only in that one respect: seal twice and deliver the second frame for a gap, deliver the same frame twice for a replay, seal in a second session for a session-id mismatch.

`Session::open` also has no test for its version / `sig_alg` rejection. That property was previously covered only by the JSON envelope's serde tests, which go with the type — so it needs direct coverage on the sealed path before those are removed, or the deletion widens a gap rather than closing one.

Follow-up to #2790 / #2794, where this was recorded as deferred.

auser:
Closes #2797. **Stacked on #2794** — review the top commit.

#2794 moved the control plane to the binary sealed envelope and recorded, but did not fix, the consequence: `fuzz_authed_path` drove `verify_authenticated_frame`, a signature-only path over the *old* JSON envelope that nothing called any more. Claim 5 kept a witness aimed at dead code.

## The obvious design doesn't work, and the mutation check is what said so

Perturb a sealed frame's `sequence`, assert `Err`. That's what my first draft did — and it passes for the wrong reason. The Ed25519 signature covers the frame context, so tampering the sequence, session id or timestamp is refused by the **signature** check. I deleted the sequence check entirely and the target stayed green through **100,994 runs**.

Isolating a check means a frame that is validly signed and wrong only in that one respect:

| scenario | how | isolates |
|---|---|---|
| gap | host seals twice, peer offered the second | sequence check |
| replay | same frame delivered twice | sequence check |
| ciphertext / signature tampering | as before | signature check |
| other session | sealed in a second session, same identities | **nothing new** |

Rebuilt, deleting the sequence check is caught in under 40 seconds: `a validly signed frame that skipped a sequence was accepted`.

The cross-session scenario deliberately claims less than it looks like it should. `derive_session_key` mixes the session id into the key, so such a frame fails AES-GCM decryption whether or not the session-id check exists — I confirmed that by deleting the check and watching the target stay green. That's defence in depth working, not a witness for the session check, and the target's comment says so rather than implying otherwise.

## A gap that had to close before the deletion, not after

`Session::open` refuses a frame whose version or `sig_alg` it doesn't speak, and **nothing tested it**. That property had been covered only by the JSON envelope's serde tests in `mvm-contract` — which were about to be deleted with the type. Removing them first would have widened a gap while looking like tidying.

`an_unsupported_version_or_sig_alg_is_refused` covers it directly on the sealed path, mutation-checked (`if false` on the version/alg guard → red) before the deletion went ahead.

## Deleted

`write_authenticated_frame`, `read_authenticated_frame`, `verify_authenticated_frame`, their seven tests, the `AuthenticatedFrame` type and its three tests.

Verified unused first — including in the **`mvmd` sibling repo**, which consumes `mvm-contract` as a path dependency and has zero references. My earlier "no production callers" grep had been too aggressive and missed doc-comment hits in `mvm-build` and `mvm-hostd`; the wider sweep found them, and they turned out to be comments rather than code, but the sweep is the reason I know that rather than assume it.

Six doc comments described the control plane as JSON-enveloped. All six were false before this branch started; they now name `SealedFrame`.

## Test-support seam

`paired_sessions_for_test` is the unit tests' own pairing helper promoted behind the existing `test-support` feature, so the fuzz target can build a session without a handshake per iteration. It is *handed* a shared secret rather than agreeing one — which is exactly why it stays feature-gated and cannot reach a production build.

## Verification

`cargo fuzz run fuzz_authed_path` — 75,260 runs clean on the baseline; caught in <40s with the sequence check removed.

Gates: nightly fmt, `clippy --workspace --all-targets -D warnings`, workspace nextest (12870 passed; the 3 `extension_provider_process` failures are #2787, which this branch predates), all 63 `xtask` gates including `check-claim-catalog`, `check-conformance` and `check-dormant-controls`.

Test count moved 12882 → 12873: ten deleted, one added.


auser:

PR #2801, opened 2026-08-22
fix(guest): set hostname from machine name

### LINKED ISSUE #2789: Guest hostname is always `(none)`
Split out of #2788, which fixed the other oddities visible in the same session:

```
$ mvmctl machine run --image rust:latest -it -- /bin/bash
I have no name!@(none):~$
```

`#2788` fixed the `I have no name!` (uid 901 now named in the image's account databases at materialization) and the empty `~` (a tmpfs home over a baked-in mount point). The `(none)` is the remainder.

## Why it is separate

Nothing in the guest calls `sethostname`, and the machine name never reaches the guest at all — `rg` for a hostname-bearing cmdline parameter or vsock field comes back empty. So this is not a one-line fix in the agent; it needs a new host→guest channel:

1. An `mvm.hostname=` kernel-cmdline parameter, carrying the machine name.
2. Plumbed through every backend's boot-args construction — `mvm-backends/src/driver/{libkrun,qemu}.rs`, `mvm-runtime/src/backends/hvf/kernel_boot.rs`, the Firecracker driver.
3. Accepted by the boot-args validator (`mvm-runtime/src/artifacts/validate.rs::boot_args_problems`).
4. Applied with `sethostname()` in `guest_bootstrap::provision_guest_environment`, before the privilege drop.

## Not an OCI-path regression

Nix-built images report `(none)` too — `nix/lib/mk-guest.nix` sets no hostname either. This has always been true for every backend and every image type, which is why it was scoped out of #2788 rather than folded in.

## Open question

What the hostname should be. Docker uses a truncated container ID; the natural analogue here is the machine name `mvmctl machine ls` shows. Worth deciding before implementing, since it is user-visible and awkward to change later.

auser:
## Summary

- set cold-boot guest hostnames from the validated machine name through the shared workload cmdline
- keep warm-pool parents identity-free and deliver each child hostname through the existing post-restore handshake
- fail closed on malformed names or sethostname errors while preserving compatibility with older peers
- add focused unit, wire, warm/cold parity, and live BDD coverage

## Validation

- cargo test --workspace
- cargo check --workspace
- just check-gated
- repository plan, sprint, and source-comment policy gates

Closes #2789


auser:

PR #2802, opened 2026-08-22
fix(ci): reconcile scheduled security outcomes

### LINKED ISSUE #2792: Security lane watch did not fire for a completed nightly run

auser:
Closes #2792

## Summary
- query only real scheduled runs for claim-bearing workflows
- require the newest scheduled run to be fresh, completed, and successful
- keep the scheduled reconciliation independent from workflow_run delivery
- update the tracking issue lifecycle and delivery records

## Validation
- cargo test -p xtask check_claim_witness_freshness
- cargo test --workspace
- cargo fmt --all -- --check
- actionlint .github/workflows/*.yml
- check-declared-backing, check-sprint-append, check-plan-names, check-no-spec-refs-in-comments
- live negative witness: correctly reports the latest scheduled Security run as completed/cancelled


auser:

PR #2840, opened 2026-08-25
test(docs): verify every documented mvmctl command against the real CLI

auser:
## Why

The README and the website docs print ~1,185 `mvmctl` invocations. The existing
`s8_readme_contract` suite checked a slice of them: README only, and only by
asking whether a flag *string* appeared somewhere in `--help` output. Nothing
checked the website docs at all, which is where the drift had gone.

A reader following the docs hit a broken command about one time in five.

## What this adds

`features/suites/s29_doc_examples/` extracts every invocation with `file:line`
provenance and verifies each at one of three tiers:

| Tier | Proves | Commands |
| --- | --- | --- |
| `parse` | the real clap tree parses it, arguments included | 86 |
| `exec` | additionally executed against an isolated `MVM_HOME` | 16 |
| `live` | additionally boots a real microVM (`MVM_BDD_LIVE=1`) | 9 |

Parsing goes through `try_get_matches_from`, so the resolved command path comes
from clap itself rather than a token heuristic — a flag *value* that happens to
spell a subcommand (`--image run`) cannot be mistaken for one.

The tier assignment is keyed by **command path, not by example**. The CLI
surface is finite and reviewable; the example set is neither. So the assignment
can be total: an unclassified path fails the suite by name, which is what stops
a newly documented verb from shipping with no decision about how it is proven.

Inline code spans are covered too, at verb level. The CLI reference documents
most of its surface in Markdown tables, so a fenced-block-only gate stays green
while an entire reference page rots — that surface alone held 35 stale
spellings. A fenced block is a recipe and must parse completely; a table cell or
a sentence names a verb without its arguments, so only the verb is checked.

### Ways the gate could have been green while lying, closed

- **Placeholder evasion** — templates (`mvmctl <verb> …`) skip parsing, so
  `<angle brackets>` would be an escape hatch for documenting a command that
  never existed. Their concrete verb prefix is resolved anyway.
- **`live` as a label** — the suite reads the `@live` scenarios back and fails
  if nothing actually runs a `live`-tier command. This caught `machine build`
  marked `live` with no witness; `documented_build_live.feature` now provides
  one.
- **Stranded commands** — a fence-balance check cannot see a fence closed one
  block early. Such a command renders as broken prose *and* drops out of
  extraction at the same moment, which is exactly how the README one survived.
- **Extractor blindness** — a corpus-size floor fails if extraction ever returns
  nothing, which would make every downstream assertion vacuously true.

Docs that correctly describe a command's *absence* ("the former `mvmctl dev
import-image`", "not exposed by a public `mvmctl policy` command") are declared
under `[[absent]]` with a reason rather than rewritten — rewriting them would
have made the docs wrong. The same check applies in reverse: if one ships, the
suite fails so the docs get updated.

## Doc fixes this found

- **~90 occurrences of verbs re-parented under `machine`** — `volume`, `cp`,
  `pause`, `resume`, `checkpoint`, `wait`, `boot-report`, `proc`, `diff`,
  `snapshot`, `sandbox`. All eleven are `hide = true`, so no help-text check
  could ever have caught them. Left hidden deliberately, to keep the advertised
  CLI surface small; only the docs changed.
- **`build` became a command group** — `mvmctl build --flake .` and
  `mvmctl build ./my-app` were dead (40+ uses); image builds are `machine build`.
- **Namespace moves** — `audit` → `trust audit`, `receipt verify` →
  `trust receipt verify`, `cleanup` → `env cleanup`, `metrics` → `ops metrics`,
  `compile` → `build compile`, `update` → `env update`.
- **Removed flags** — `machine exec --timeout`, `manifest ls --legacy`,
  `build --vcpus/--mem/--data-disk/--snapshot`, `checkpoint restore --name`,
  `pack download --kind` (now a positional).
- **`mvmctl up`** in both language-SDK READMEs — `up::Args` is not a `Commands`
  variant, so the verb has no dispatch path at all.
- **A prematurely closed README fence** that stranded an `mvmctl machine run`
  example in the prose.
- A lifecycle table had the attach/detach default **backwards** and named two
  commands (`run --attached`, `detach`) that do not exist.

## Evidence

`just bdd`, macOS 26 / Apple Silicon, no `/dev/kvm`:

- Baseline `origin/main`: 197 scenarios, 188 passed, 8 failed.
- This branch: 206 scenarios, 198 passed, 7 failed.

The failing scenarios are **identical on both sides** — pre-existing, from an
unbuilt TypeScript SDK `dist/` and an absent SDK codegen binary. I ran the
baseline on a pristine `origin/main` worktree rather than assume this. 9 new
scenarios, all passing, no regressions.

`fmt`, `clippy -D warnings` (including the `bdd` target), `nextest -p
mvm-conformance`, and `xtask check-all / check-conformance /
check-declared-backing / check-honesty / check-nextest-groups` all pass, before
and after rebasing onto current main.

## Limits worth knowing before release

- **The `live` tier is not exercised here.** 28 live scenarios skipped for want
  of a hypervisor. Parse and exec are green; the live tier needs
  `MVM_BDD_LIVE=1 just bdd` on a KVM/HVF host.
- `build compile --dev`'s own help text still says "Refused on `mvmctl
  compile`", naming the pre-`build` spelling. That string is CLI source, not
  documentation, so this gate does not see it — noted as a follow-up sweep of
  user-facing strings in `crates/mvm-cli`.


auser:
Superseded by #2846, which contains this PR's commit plus the Rust, Python, TypeScript and Nix example verification built on top of it.

Closing rather than leaving both open: the same commit in two open PRs means whichever merges second conflicts. #2846 carries everything.

For the record on why this one never landed despite being green for hours — it was enqueued 11+ times and every merge-group run was **cancelled** after 2–9 minutes, then dequeued. The ruleset has `min_entries_to_merge: 2` with `max_entries_to_build: 2`, so a lone PR waits for a partner and the arriving partner re-forms the batch, cancelling the in-flight run. Any PR whose CI takes ~28 minutes is structurally starved. Worth fixing independently of this work.

PR #2846, opened 2026-08-25
test(docs): verify every documented example — CLI, Rust, Python, TypeScript, Nix

auser:
Supersedes #2840, which contains this PR's first commit and has been starved by
the merge queue for ~6 hours (see "Why this is one PR" below).

## Why

The README and the website docs printed ~1,185 `mvmctl` invocations and 249
code blocks. The only thing checking any of it was a README-only suite that
asked whether a flag *string* appeared somewhere in `--help` output, and
whether a Nix block contained the substring `mkGuest`. Nothing checked the
website docs at all, and nothing compiled or executed a single code example.

A reader following the docs hit a broken command about one time in five.

## What now happens on every PR

`features/suites/s29_doc_examples/` extracts every example with `file:line`
provenance and verifies it against the real thing:

| Surface | Blocks | How it is proven |
| --- | --- | --- |
| CLI (`bash`/`sh`) | 421 | parsed by the real clap tree (arguments included); side-effect-free ones executed against an isolated `MVM_HOME`; 9 boot a real guest in the `@live` lane |
| `rust` | 18 | **compiled** against the workspace crates via `build.rs` → `OUT_DIR` → a test target |
| `python` | 53 | resolved against the **real installed SDK** — attributes, `from mvm import`, keyword names, and missing required arguments |
| `ts` | 50 | resolved against the SDK's `src/` **export graph** |
| `nix` | 44 | `mkGuest` attributes read from its actual argument set |
| `toml`/`json` | 38 | parsed |

The tier assignment is keyed by **command path, not example**: the CLI surface
is finite and reviewable, the example set is not. An unclassified path fails by
name, so a newly documented verb cannot ship without someone deciding how it is
proven.

Inline code spans are covered too. The CLI reference documents most of its
surface in **tables**, so a fenced-block-only gate stays green while an entire
reference page rots — that surface alone held 35 stale spellings.

## What it found

Roughly 240 defects. The ones that mattered most:

- **~90 CLI occurrences of verbs re-parented under `machine`** (`volume`, `cp`,
  `pause`, `checkpoint`, `wait`, `snapshot`, …). All are `hide = true`, so no
  help-text check could ever have caught them.
- **The README's flagship Python example could not run.** `mvm.resources(cpu=…)`
  is the wrong keyword and omits two required arguments, and `@mvm.app(...)`
  omitted the required `name=` and `source=`. Verified by executing it:
  `TypeError: app() missing 2 required keyword-only arguments`.
- **13 TypeScript blocks import `"mvm-sdk"`. The package is `@runmvm/mvm`.**
  `npm install mvm-sdk` does not get you this SDK.
- **`mvm_client_local` is not a crate**; `println!("{}", m.id)` on a type with no
  `Display`; `resources(1, 256)` against a 3-argument signature;
  `WorkloadBuilder::app` taking an `App`, not a closure. None of these are
  greppable — only a compiler finds them.
- `healthChecks`, `volumeMounts`, `serviceGroup` — taught by the mkGuest guide,
  present nowhere in `nix/`. Declared, and the pages now carry a
  "not implemented yet" notice.
- A prematurely closed README fence that stranded a command in the prose, and a
  reference table row broken by an unescaped `|` inside `<put|get|set|ls|rm>`.

## Ways the gate could have been green while lying, closed

- **Placeholder evasion** — a template skips parsing, so `<angle brackets>`
  would be a way to document a command that never existed. Its verb prefix is
  resolved anyway.
- **Unexplained opt-outs** — a `rust,ignore` block must state why on its first
  line. A marker with no reason is how a wrong example survives.
- **`live` as a label** — the suite reads the `@live` scenarios back and fails
  if nothing runs a `live`-tier command. That caught `machine build` marked
  `live` with no witness.
- **Extractor blindness** — a corpus-size floor fails if extraction returns
  nothing, which would make every assertion vacuously true.
- **Docs that correctly describe absence** ("the former `mvmctl dev
  import-image`") are declared under `[[absent]]`, and the suite fails if one
  ships — so a "removed" note cannot outlive its truth.

Every checker was **mutation-tested red-first**: renaming a symbol in the
README fails the suite naming that line, and restoring returns to green. Two
parser bugs were found this way — the Nix parser was reading words out of `#`
comments, and the extractor was treating shell redirects as placeholders.

## Evidence

- Baseline `origin/main`: 197 scenarios, 188 passed, 8 failed.
- This branch: 211 scenarios, 203 passed, 7 failed.

The failures are **identical on both sides** and pre-existing — the TypeScript
SDK `dist/` is unbuilt in a fresh worktree and the SDK codegen binary is absent.
I ran the baseline on a pristine `origin/main` worktree rather than assume it.
14 new scenarios, all passing, no regressions.

`fmt`, `clippy -D warnings` (including the `bdd` target), `nextest`, and
`xtask check-all / check-conformance / check-declared-backing / check-honesty /
check-nextest-groups / check-sprint-append` all pass, before and after rebasing
onto current main.

## Why this is one PR

#2840 carried the first commit alone and was green for ~6 hours without
merging. It was enqueued 11+ times; every merge-group run was **cancelled**
after 2–9 minutes and the PR dequeued. The ruleset explains it:
`min_entries_to_merge: 2` with `max_entries_to_build: 2` means a lone PR waits
for a partner, and the arriving partner re-forms the batch and cancels the
in-flight run. A PR whose CI takes ~28 minutes never survives a re-form. That
starves any long-CI PR, not just this one — worth fixing separately
(`min_entries_to_merge: 1`).

Rather than stack a second PR on an unmerged base that will be squash-merged —
the reliable way to strand work here — everything is in this one.

## Not covered

The `live` tier is asserted but not executed here: 9 commands need
`MVM_BDD_LIVE=1 just bdd` on a KVM/HVF host. The suite fails if a `live` label
has no scenario behind it, but nothing can make a skipped scenario prove
anything.


auser:
Rebased onto `f7dcff61ec` and re-verified on the new base: 212 scenarios, 204 passing, same 7 pre-existing failures, zero new. `fmt`, workspace `clippy -D warnings`, the `bdd` target, `nix_flake_structure`, and six xtask gates all pass.

Two commits added since the PR was opened:

**`fix(nix)`** — mkGuest accepts `healthChecks`, `volumeMounts` and `serviceGroup`, shape-checked and recorded in `passthru.mvm.unenforced`, with an eval-time trace saying they are declared but not enforced. The multi-service supervisor they depend on is still a stub, so accepting them silently would have been worse than rejecting them.

**`fix(cli)`** — every scaffolded flake now declares an `entrypoint`. All five bundled presets declared only a top-level `services` block; mkGuest requires `entrypoint` and never reads top-level `services`, so **every generated project was unbuildable**:

```
error: function 'anonymous lambda' called without required argument 'entrypoint'
```

Verified by generating all five scaffolds and applying each one's own mkGuest arguments to the real function — all five now get past the argument check, where before every one failed. Two tests pin it, both checked red-first.

Neither of these came from reading the docs; they came from the docs gate refusing to accept a claim it could not verify.

PR #2859, opened 2026-08-25
fix(builder): make copied rootfs writable before sealing

auser:
Follow-up to #2852.

The self-contained rootfs journal seal added in #2852 runs `e2fsck -p -f` on the copied rootfs artifact. Some copy paths leave the file read-only (e.g. mode 0444), which causes `e2fsck` to fail because it cannot open the filesystem read-write for repair.

This change normalizes the copied rootfs to mode 0644 immediately before the offline journal check, and removes the artifact if normalization fails so a broken file is never published.

## Changes

- `crates/mvm-build/src/builder_vm_runtime.rs`
  - `seal_rootfs_journal_sh`: add `chmod 0644` before `e2fsck`; delete and exit on failure
- `crates/mvm-build/src/persistent_builder.rs`
  - Updated existing tests to assert the new ordering and failure path

## Verification

- `cargo test -p mvm-build` for the affected modules
- Workspace clippy via pre-commit hook

### DIFF
--- a/crates/mvm-build/src/builder_vm_runtime.rs
+++ b/crates/mvm-build/src/builder_vm_runtime.rs
@@ -138,6 +138,11 @@ pub const MVM_BUILDER_STORE_GC_GIB_ENV: &str = "MVM_BUILDER_STORE_GC_GIB";
 pub(crate) fn seal_rootfs_journal_sh(rootfs_path: &str) -> String {
     format!(
         r#"echo "mvm-builder-vm: sealing rootfs journal" >&2
+if ! chmod 0644 {rootfs_path}; then
+    echo "mvm-builder-vm: rootfs permission normalization failed" >&2
+    rm -f {rootfs_path}
+    exit 1
+fi
 set +e
 /sbin/e2fsck -p -f {rootfs_path} >&2
 fsck_rc=$?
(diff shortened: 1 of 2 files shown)


auser:
Closing as a duplicate of #2858 — the diffs are byte-identical (`chmod 0644` on the copied rootfs before the journal seal, plus the ordering assertion and the removal-on-failure check).

#2858 is the one to keep: it is a single commit on a branch cut from main, and reads `BLOCKED` (awaiting checks). This one reads `DIRTY` because it also carries `da9ec54b0a`, whose content already landed through #2852 as a squash — so the branch conflicts with main on exactly the change that is already there.

Nothing is lost by closing this: the fix itself lives on in #2858.

For context, this fix and the one in #2856 are **complementary, not competing** — two distinct causes of the same `e2fsck`-before-seal symptom, in different files:

* **#2858** — the copied rootfs is read-only, so `e2fsck` cannot write repairs. Fixed by normalising permissions before the seal.
* **#2856** — the rootfs is still *mounted*, because a recursive bind left submounts the teardown never unmounted, so `e2fsck` exits 8 ("operational error", not a corrupt journal). Fixed by unmounting the subtree deepest-first.

Both are needed for `mvmctl machine build --flake` to work on a KVM host; I found the second one by running that command live for the first time.

PR #2883, opened 2026-08-26
perf(build): take the per-VM aux-helper leg off the inner loop

auser:
## What

`mvm-cli`'s build script was **17.8s of a 20.9s rebuild (85%)** after a one-line
`mvm-core` edit, and nothing else could start until it finished — the
`mvm-cli(build)` progress line everyone sits watching.

Plan 334 fixed the musl leg and named this as its own residual: *"Native per-VM
helpers still always rebuild (13s)."* It has since grown to 17.8s. This closes it.

## Why it could not just reuse

The content store (#2644) already keys both legs on their real dependency
closure. The aux leg still **cannot hit on the inner loop by construction**:
every per-VM helper links `mvm-hostd → mvm-core`, so any edit under those trees
misses by definition and paid a nested `cargo build` of a 100K-LOC crate in a
second target dir.

Reusing on a miss the way the musl leg does was not available on its own. A
stale *embedded* binary only sits inside `mvmctl`; a stale *supervisor* is
spawned, and a guest that silently ignores your edit is worse than a slow build
— which is exactly why #2058 made the rebuild unconditional.

So the check moves from build time to use time:

- `crates/mvm-cli/build.rs` — reuse on a key miss under the dev profile, and
  write a `<bin>.mvm-stale` marker beside the binary. Cleared on any path that
  proves freshness. A cold worktree still builds; `--release` /
  `release-witness` still always build, matching the musl leg's existing rule.
- `crates/mvm-vmm/src/host/aux_bin.rs` — `resolve` refuses a marked helper with
  an error naming `just embed-refresh`, on both the directory-search and
  explicit-override paths.

The silent ~30s hang becomes an instant, actionable error, **and** the 17.8s
comes off every build that does not boot a VM — including every `cargo check`
and `clippy`. Markers only exist in a source checkout's build-script directory,
so a downloaded release is untouched.

## Measured (aarch64 macOS, 16 cores, `jobs=6`)

`cargo build -p mvm-cli --timings` after a one-line `mvm-core` edit:

| Unit | Before | After |
|---|---|---|
| total wall | 20.9s | **8.5s** |
| `mvm-cli run-custom-build` | **17.8s** (0.8→18.6s) | not in top 8 |
| `mvm-cli` (the crate) | 2.3s, blocked until 18.6s | 2.2s, starts at 6.3s |

## Escape hatches

`just embed-refresh` · `MVM_EMBED_NO_CACHE=1` · `MVM_ALLOW_STALE_AUX=1`

## Tests

10 new cases in `aux_bin.rs`. Two drive `resolve_from` through both resolution
paths, so the gate is proven **wired**, not merely defined — mutating
`refuse_if_stale` to always admit turns exactly those red.

Gate run: nightly `fmt --all`, stable `clippy --workspace --all-targets -D
warnings`, `nextest --workspace`, `--doc`, `just check-gated`, and
`xtask check-all` (61 gates) + `check-declared-backing` +
`check-nextest-groups` + `check-single-workload-env`.

`nextest --workspace` failure sets are **byte-identical** to unmodified `main`
(111 pre-existing `mvmctl::*` failures, unrelated: `CARGO_BIN_EXE_mvmctl` is
unset under nextest while `cargo test` passes the same tests — reported
separately).

### DIFF
--- a/crates/mvm-cli/build.rs
+++ b/crates/mvm-cli/build.rs
@@ -338,7 +338,7 @@ fn main() {
         );
     }
 
-    build_native_aux_helpers(&workspace_root, &nested_target_dir, &mut cache);
+    build_native_aux_helpers(&workspace_root, &nested_target_dir, &mut cache, dev_profile);
 
     let embedded_rs = render_embedded_rs(&entries);
     std::fs::write(out_dir.join("embedded.rs"), embedded_rs).unwrap();
@@ -400,6 +400,7 @@ fn build_native_aux_helpers(
     workspace_root: &Path,
     nested_target_dir: &Path,
     cache: &mut EmbedCache,
+    dev_profile: bool,
 ) {
     let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
     let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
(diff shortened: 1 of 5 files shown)


auser:
reopening to fire CI (no check runs registered)

Issue #2888, opened 2026-08-26
HVF has no SMP: --cpus>1 exits 0 but the guest gets one vCPU (FDT, PSCI CPU_ON, supervisor config all single-CPU)


auser:
`mvmctl machine run --cpus N` has no effect on the HVF workload path. The guest
always boots with a single vCPU.

## Reproduction

```
$ mvmctl machine run --image alpine --cpus 4 -- sh -c 'nproc; grep -c ^processor /proc/cpuinfo'
1
1
```

`/proc/cpuinfo` agrees with `nproc`, so this is the guest genuinely having one
CPU, not a scheduling-affinity artifact. Same result with `--cpus 2`, and the
same with `MVM_RESIDENCY=cold`, so it is not the warm pool handing back a
differently-shaped guest.

## Pre-existing

Reproduces identically on a build predating #2885, so it is not caused by the
HVF stderr or `--up-json` fixes. It was unreachable before #2884, when no guest
booted on this host at all.

## Why it matters

`--cpus` is documented in the README's quick start:

```
mvmctl machine run --image alpine --cpus 2 --memory 512M \
  --allow-host api.example.com:443 -- ./fetch
```

A workload asking for parallelism silently gets none. Worth checking whether
`--memory` is honoured on the same path, and whether the persistent
(`machine create --cpus`) path differs from the transient one — this reproduction
is transient only.

## Test coverage

`features/suites/s31_launch_e2e/cli_launch_modes.feature`'s "--cpus and --memory
are honoured on a real boot" is tagged `@wip` pointing here, asserting what the
README documents. Un-tag when this is fixed.

Note the scenario passed in earlier runs of that suite **for the wrong reason**:
its assertion matched a substring of the combined stdout+stderr, and the
`MVM_PHASE_TIMING` diagnostic table supplied a "2". The step now reads the
guest's own stdout and compares the whole line, which is what made the defect
visible.

auser:
## Root cause: HVF is single-vCPU by construction, not dropping a value

Traced end to end. `--cpus` reaches `VmStartConfig.cpus` and then
`VmmSpec.vcpus` (`spec_map.rs:333`) correctly. It stops there, and three
independent things would each have to change for it to mean anything:

1. **The supervisor config cannot carry it.**
   `HvfSupervisorConfig` (`crates/mvm-vmm/src/host/hvf_supervisor.rs`) has
   `memory_mib` and no `vcpus` field at all, so the count is never piped to the
   process that creates the VM.

[...]

Issue #2887, opened 2026-08-26
fs/proc verb refusals were reported as protocol mismatches; dev-mode launches still cannot grant DevOnly verbs


auser:
`mvmctl machine fs write` fails against a running HVF guest, regardless of destination.

## Reproduction

```
$ mvmctl machine run -d --name fsw --image alpine --ttl 300s
$ echo hi | mvmctl machine fs write fsw /app/main.sh --mode 420
Error: Unexpected response to FS RPC verb
$ echo hi | mvmctl machine fs write fsw /tmp/main.sh --mode 420
Error: Unexpected response to FS RPC verb
```

Both a read-only image path and a writable tmpfs path fail identically, so this
is not the guest-mount allow-list or a permissions issue — the agent returns a
response shape the host does not expect.

## Pre-existing

Reproduces identically on a build that predates #2885's changes, so it is not
caused by the HVF stderr fix or the `--up-json` banner fix. It was simply
unreachable before #2884: no guest booted on this host at all.

## Impact

The runtime SDK's `Sandbox.files.write` shells this verb
(`crates/mvm-sdk/sdks/python/mvm/_sandbox.py:1085` → `_run_bytes`), so the
README's documented runtime-SDK example

```python
with mvm.Sandbox.create(image="python-3.12") as sb:
    sb.files.write("/app/main.py", "print('hi from mvm')")
```

cannot complete on HVF. `Sandbox.create` and `commands.start` both work — the
break is specific to the FS verbs.

## Where to start

- The error text is the host's, on receiving a reply it could not match to the
  verb it sent. Compare what the agent's FS handler returns against the response
  variant the host matches on; a `#[serde(deny_unknown_fields)]` type on either
  side will surface a shape mismatch rather than a transport failure.
- Worth checking whether any FS verb works (`fs read`, `cp`) or whether the whole
  family is affected — that distinguishes a wire-format drift from one handler.

## Test coverage

`features/suites/s31_launch_e2e/sdk_and_library_modes.feature`'s live-mode
scenario exercised this incidentally. Its fixture no longer calls `files.write`,
with a comment pointing here, so the scenario keeps witnessing the live
transport rather than going red on an unrelated defect. Restore the
`files.write` call when this is fixed.

auser:
## Wider than `fs write` — the whole `fs`/`proc` verb family

Re-tested against a build predating #2885, on a guest booted with
`machine run -d --name pv --image alpine`:

```
[...]

PR #2889, opened 2026-08-26
fix(hvf): stop the per-VM supervisor holding its caller's stderr

### LINKED ISSUE #2885: HVF supervisor inherits the caller's stderr, so any captured-output caller hangs
Found by the new end-to-end launch suite (`features/suites/s31_launch_e2e/`) on
macOS 26 / Apple Silicon with the HVF backend.

**Not** the launch regression fixed in #2884. That one — a cold universal-initramfs
cache silently producing a guest with no runtime overlay — is fixed and witnessed.
It had masked this defect completely: before that fix no guest booted at all on
this host, so nothing ever reached the persistent path's post-boot steps.

## Symptom

The README's documented persistent flow starts a guest that really is running,
but the CLI never completes the step:

```
mvmctl machine create e2e-web --image alpine --cpus 2 --memory 512M   # ok
mvmctl machine start  e2e-web                                          # never returns
```

`mvmctl machine ls` meanwhile reports the machine `running` on backend `hvf`, so
the VM itself came up. Against an already-running machine:

```
$ mvmctl machine exec e2e-web -- uname -s
failed to spawn: I/O error (os error 5)
```

The SDK's live transport rides the same path — `_LiveTransport.for_source` shells
`machine run -d --up-json --name <id> --image <ref> --ttl <n>s` — so `mvmctl run
--mode live` fails with it, likewise leaving a running VM behind while the command
hangs.

## What is unaffected

Every transient shape works and is witnessed live by the same suite: `machine run
--image`, multi-word argv after `--`, `--env`, `--mount`, `--allow-host` (egress
reaches the target), default-deny without it, guest exit-code propagation, and
`--cpus`/`--memory`. Warm dispatch measures 179ms, inside the 200ms budget.
`mvmctl run --mode plan` admits a signed plan without booting.

So the break is specific to the persistent/named path's post-boot handshake, not
to booting.

## Scenarios currently pending

Tagged `@wip` rather than deleted — the harness names them on every run
(`2 scenario(s) did NOT run`). Deleting the scenario you cannot pass is how the
regression in #2884 survived to a release.

- `s31_launch_e2e/cli_launch_modes.feature` — "the documented persistent machine
  lifecycle operates one guest"
- `s31_launch_e2e/sdk_and_library_modes.feature` — "a runtime-SDK script boots a
  real guest in live mode"

## Where to start

- `os error 5` is `EIO`, surfacing from the spawn attempt against the running
  guest — the agent channel is reachable enough to try and not enough to carry a
  command. Look at the persistent path's agent socket handoff, not at boot.
- The transient path reaches its agent fine on the same host and backend, so diff
  the two: transient holds the supervisor for the command's lifetime, persistent
  detaches and reconnects.
- `machine ls` reading `running` means the liveness probe and the agent channel
  disagree; the shared five-marker probe is what says `running`.

Full write-up: `specs/plans/2026-08-26-persistent-machine-path-on-hvf.md`

auser:
Closes #2885. **Stacked on #2884** — base is `fix/egress-client-launch-gate`, retarget to `main` once that merges.

## The defect

`mvmctl machine start` spawned the detached `mvm-hvf-supervisor` with
`stderr(Stdio::inherit())`. The supervisor owns its guest for the VM's whole
life, so it held the **spawning process's stderr file descriptor** for that whole
life. `mvmctl` returns in 0.4s, but a caller that captures stderr through a pipe
and reads to EOF never sees EOF:

```
mvmctl machine start X > file 2>&1     # 0.426s, exits 0
mvmctl machine start X 2>&1 | cat      # never terminates
```

That reaches `Command::output()` (the BDD steps), Python's
`subprocess.run(capture_output=True)` — the SDK's live transport, at every call
site — and any `2>&1 |` pipeline. It is invisible interactively at a TTY,
because there is no pipe to hold open. Only automation sees it.

## The original diagnosis was wrong

#2885 was filed as "`machine start` hangs and `machine exec` returns `os error
5`". Both halves are wrong: `start` returns promptly and `exec` works. The EIO
came from a machine left half-started by a SIGKILLed debug supervisor and does
not reproduce on a cleanly started machine. The correction is on the issue and
in the plan doc rather than edited away.

## Fix

libkrun already routed supervisor stderr to a per-VM log file; the two HVF sites
were never converted. Rather than copy its inline block a third time, the routing
is now one helper — `mvm_vmm::host::console_capture::supervisor_stderr` — shared
by `hvf.rs`, `hvf_restore.rs` and `libkrun.rs`, so they cannot drift again. It
falls back to `inherit` only when the state dir cannot be written, which is a
boot that is already failing and where losing the diagnostics would be worse
than a hang.

## A second defect this exposed

With the leak fixed, the SDK scenario failed differently: `machine run -d
--up-json` printed the human `started machine <name>` banner to **stdout** ahead
of the JSON envelope, so `json.loads(stdout)` died at line 1 column 1.
`machine start` already withheld the banner under `--json`; the `run` path
hardcoded `quiet: false`.

`machine_run_up_json_guards_stdout` existed the whole time and passed the whole
time — it pins `emits_machine_readable_stdout()`, which says stdout is
*reserved* and nothing about whether anything else writes to it. The
start-argument construction is now a named function (`start_args_for_run`) so
the wiring is pinned, not just the decision; a mapping that stopped consulting
`banner_suppressed` would otherwise leave a decision-only test green. Verified
red against a `quiet: false` mutation.

## A weak assertion of my own, fixed

`the guest printed` matched against the combined stdout+stderr. A short
expectation like `"2"` was matching digits in the `MVM_PHASE_TIMING` table, so
the `--cpus` scenario passed while the flag did nothing. It now reads the
guest's own stdout, and `the guest printed exactly` pins a whole line for
one-word answers like `nproc`.

## Three pre-existing defects found, filed, not hidden

All reproduce on a build predating this work. They were unreachable before
#2884, when no guest booted on this host at all.

- **#2887** — every guest-RPC `fs`/`proc` verb answers `Unexpected response to
  ... verb`, so the runtime SDK's live mode cannot complete a documented README
  example. `Sandbox.create` works; the verbs it then issues do not.
- **#2888** — `--cpus` is silently ignored; the guest always boots one vCPU and
  `/proc/cpuinfo` agrees with `nproc`.
- A signing trap, recorded in the plan and handled in the runner: `cargo build`
  re-links the per-VM supervisor whenever its dependency graph changes and does
  not re-sign it (`mvm-hvf-supervisor` has no `ensure_signed()`; the libkrun one
  does). macOS SIGKILLs the unentitled binary and the only symptom is
  `supervisor exited before writing its PID file (status: signal: 9)`, which
  reads exactly like a boot regression. `scripts/e2e-launch-modes.sh` now runs
  `mvmctl env sign` after every build.

Two scenarios stay `@wip`, re-attributed to #2887 and #2888 rather than deleted,
so the harness names them on every run.

## Result on macOS 26 / Apple Silicon / HVF

```
3 features
12 scenarios (12 passed)
71 steps (71 passed)
[bdd] 2 scenario(s) did NOT run:
[bdd]     2  @wip
[bdd] a green suite is not full coverage while any of these are nonzero.
```

Unit witnesses: `supervisor_stderr_creates_the_log_inside_the_state_dir`,
`supervisor_stderr_truncates_a_previous_boots_log`,
`supervisor_stderr_falls_back_when_the_state_dir_is_absent`,
`machine_run_up_json_withholds_the_started_banner`,
`machine_run_without_up_json_keeps_the_started_banner`.

`check-no-spec-refs-in-comments`, `check-plan-names`, `check-declared-backing`,
`check-honesty`, `check-no-overclaim` all clean.

### DIFF
--- a/crates/mvm-agentd/src/vsock/api.rs
+++ b/crates/mvm-agentd/src/vsock/api.rs
@@ -359,10 +359,11 @@ pub fn send_proc_request_on(stream: &mut UnixStream, req: GuestRequest) -> Resul
     let resp = call_unary(stream, &req)?;
     match resp {
         GuestResponse::ProcResult(r) => Ok(r),
-        GuestResponse::Error { message } => {
-            bail!("Guest proc-control transport error: {}", message)
-        }
-        _ => bail!("Unexpected response to proc-control verb"),
+        other => bail!(
+            "expected ProcResult for {}, got {:?}",
+            req.verb().name(),
+            other.variant()
+        ),
     }
 }
 
@@ -445,8 +446,11 @@ pub fn send_fs_request_on(stream: &mut UnixStream, req: GuestRequest) -> Result<
     let resp = call_unary(stream, &req)?;
     match resp {
         GuestResponse::FsResult(r) => Ok(r),
-        GuestResponse::Error { message } => bail!("Guest FS RPC transport error: {}", message),
-        _ => bail!("Unexpected response to FS RPC verb"),
+        other => bail!(
+            "expected FsResult for {}, got {:?}",
+            req.verb().name(),
+            other.variant()
+        ),
     }
 }
 
--- a/crates/mvm-backends/src/driver/hvf.rs
+++ b/crates/mvm-backends/src/driver/hvf.rs
@@ -341,7 +341,9 @@ fn boot_with_handoff(
     let mut child = bounded_supervisor_command(&supervisor, spec, &paths.state_dir)
         .stdin(Stdio::piped())
         .stdout(Stdio::null())
-        .stderr(Stdio::inherit())
+        .stderr(mvm_vmm::host::console_capture::supervisor_stderr(
+            &paths.state_dir,
+        ))
         .spawn()
         .map_err(|e| anyhow!("spawn {}: {e}", supervisor.display()))?;
     child
--- a/crates/mvm-backends/src/driver/hvf_restore.rs
+++ b/crates/mvm-backends/src/driver/hvf_restore.rs
@@ -251,7 +251,9 @@ pub fn restore_hvf_vm(req: &HvfRestoreRequest<'_>) -> Result<RestoredHvfVm> {
     let mut child = bounded_restore_command(&supervisor, req)
         .stdin(Stdio::piped())
         .stdout(Stdio::null())
-        .stderr(Stdio::inherit())
+        .stderr(mvm_vmm::host::console_capture::supervisor_stderr(
+            req.state_dir,
+        ))
         .spawn()
         .with_context(|| format!("spawning {}", supervisor.display()))?;
     child
--- a/crates/mvm-backends/src/driver/libkrun.rs
+++ b/crates/mvm-backends/src/driver/libkrun.rs
@@ -382,11 +382,7 @@ impl VmmDriver for LibkrunDriver {
         let stdout = mvm_vmm::host::console_capture::open_console_capture(&console_log)
             .map(Stdio::from)
             .unwrap_or_else(|_| Stdio::null());
-        let stderr = mvm_vmm::host::console_capture::open_console_capture(
-            &state_dir.join("supervisor.stderr.log"),
-        )
-        .map(Stdio::from)
-        .unwrap_or_else(|_| Stdio::inherit());
+        let stderr = mvm_vmm::host::console_capture::supervisor_stderr(&state_dir);
         let mut child = bounded_supervisor_command(&supervisor, spec, &state_dir)
             .stdin(Stdio::piped())
             .stdout(stdout)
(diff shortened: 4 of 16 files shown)


auser:
## Before merging: this is stacked on #2884 and must be rebased first

#2884 auto-merges with **squash**, which creates a new commit on `main` rather
than replaying this branch's base commits. So once it lands, this branch still
carries the pre-squash originals and merging it directly would re-apply work
already on `main`.

Resolution, once #2884 shows MERGED:

```sh
cd .worktrees/mvm-2885-persistent-hvf
git fetch origin
git rebase --onto origin/main 5f430d00911a54e912b27d38c2287e92e14bae54 fix/2885-persistent-machine-hvf
git push --force-with-lease
gh pr edit 2889 --base main
```

`5f430d00911a54e912b27d38c2287e92e14bae54` is the merge-base with the #2884 branch as of this comment — everything
up to it belongs to #2884, everything after is this PR's single commit.

Verify afterwards that the diff is only this PR's own change:

```sh
git log --oneline origin/main..HEAD     # expect exactly 1 commit
```

Do not delete `fix/egress-client-launch-gate` until this retarget is done —
deleting a squash-merged base out from under a stacked PR is what strands it.

PR #2911, opened 2026-08-26
test: stop two local-environment faults from faking test failures

auser:
Two faults that made `just test` report failures that were not in the code.
Both were invisible in CI, which is what let them persist.

## 1. 111 CLI tests "failing" under nextest

Every root-package CLI integration test failed with `CARGO_BIN_EXE_mvmctl is
unset` — **111 tests across 22 files** — while `cargo test --test <the same
test>` passed. The entire root CLI suite was gating nothing under `just test`.

**Cause.** The pinned nightly cargo emits test binaries under
`target/debug/build/<pkg>/<hash>/out/` rather than `target/debug/deps/`.
`assert_cmd::Command::cargo_bin` reads `CARGO_BIN_EXE_<name>` from the
environment, and otherwise falls back to popping one directory and expecting
`deps` — so under the new layout it looks for `mvmctl` inside `.../out/` and
finds nothing. `cargo test` survives because cargo exports the variable into
the child environment. **cargo-nextest 0.9.122 (2026-01) does not.**

Updating to 0.9.143 takes the suite from **111 failures to 0**.

Nothing in the tests is wrong, so nothing in the tests changes. What is added
is a floor: `nextest-min` in `Cargo.toml` beside the other tool pins, plus
`scripts/require-nextest.sh` wired into `test`, `test-ci`, `test-crate` and
`test-filter`.

**The guard is the point, not the version bump.** 111 red tests read as broken
code, not a stale binary — and CI never sees it, because the runner installs a
current nextest every time. So local goes red while CI stays green, which is
the worst way for this to present. `0.9.143` is a known-good floor, not a
bisected minimum.

Guard verified on all three paths: below-floor fails with an actionable
message, at-floor passes, above-floor passes.

## 2. A gate that walked into another checkout

`test_support_source_owners_match_the_targeted_ci_lane` failed with 34 paths
under `.claude/worktrees/<name>/` — a **gitignored worktree belonging to
another session**, nested inside the repo.

The 12 source-scanning gates walk the filesystem rather than `git ls-files`,
and `fs_walk`'s skip list covered only `target`, `.git` and `node_modules`, so
any nested checkout was scanned and its files reported as findings here.

`fs_walk::walk_files` now refuses to descend into a directory containing a
`.git` entry — a *file* for a linked worktree, a *directory* for a clone. The
root being scanned is never tested, so a normal run over the repo root is
unaffected.

Verified both ways: with the `.git` marker present the gate is green; with it
removed the same tree turns the gate red. So the skip is what does the work,
rather than the gate passing by accident.

## Gates

`cargo +nightly fmt --all --check`, stable `clippy --workspace --all-targets -D
warnings` (via the hook), full `cargo nextest run --workspace` — **12,548 tests,
0 failures** — and `xtask` `check-all` (61 gates) + `check-declared-backing` +
`check-nextest-groups`.

`check-fast-cargo` fails on this branch for a reason that predates it and is
unrelated: `bdd.yml` still pins `nightly-2026-08-08` while `rust-toolchain.toml`
is on `nightly-2026-08-25`. That is fixed by #2891, which is already in the
merge queue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -695,6 +695,19 @@ rust = "1.91.1"
 zig = "0.13.0"
 cargo-zigbuild = "0.23.0"
 
+# Developer tooling that is NOT vendored or pinned by rust-toolchain.toml, so a
+# contributor's machine can silently drift below what the workspace needs.
+#
+# nextest: the pinned nightly cargo emits test binaries under
+# `target/debug/build/<pkg>/<hash>/out/` rather than `target/debug/deps/`.
+# A nextest older than this does not export `CARGO_BIN_EXE_<name>` for that
+# layout, and `assert_cmd`'s fallback — pop one dir, expect `deps` — then looks
+# for `mvmctl` in the wrong place. Every one of the 111 root-package CLI
+# integration tests fails with "CARGO_BIN_EXE_mvmctl is unset", which reads as
+# broken code rather than a stale tool. `scripts/require-nextest.sh` turns that
+# into one actionable line. This is a known-good floor, not a bisected minimum.
+nextest-min = "0.9.143"
+
 # Per-arch musl target. build.rs picks the entry for CARGO_CFG_TARGET_ARCH
 # (the arch mvmctl is built for == the arch of the guest it boots, since the
 # local builder/Stage 0 VM is always same-arch as the host). Plan 164.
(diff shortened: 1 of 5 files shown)


auser:
reopening to fire CI (no check runs registered)

PR #2926, opened 2026-08-27
feat(hvf): give the guest the vCPUs it asked for, or say what it got

### LINKED ISSUE #2888: HVF has no SMP: --cpus>1 exits 0 but the guest gets one vCPU (FDT, PSCI CPU_ON, supervisor config all single-CPU)

### LINKED ISSUE #2924: Default 2 vCPUs makes every macOS launch fail since HVF began refusing multi-vCPU
## Impact

On macOS — where HVF is the default workload backend — **every launch that does
not pass `--cpus 1` now fails**, including the README quickstart:

```
$ mvmctl machine run --image alpine -- /bin/echo hi

Caused by:
    the hvf backend supports exactly 1 vCPU, but this launch requests 2
```

No `--cpus` is passed. The default is 2:

```
$ mvmctl machine run --image alpine --dry-run --json -- /bin/true
  "cpus": 2
```

## Cause

#2899 ("fix(hvf): refuse unsupported multi-vCPU launches", merged 2026-08-27
10:33) made the HVF driver refuse `spec.vcpus != 1`:

```rust
// crates/mvm-backends/src/driver/hvf.rs:132
if spec.vcpus != 1 {
```

The refusal is correct — running a guest that silently loses a vCPU is worse.
But the default vCPU count was not moved with it, so the default configuration
and the default backend are now mutually exclusive on the platform where that
backend is the default.

## Blast radius

Measured on a macOS/HVF host: **26 of 33 failing scenarios** in the
documented-surface e2e suite trace to this one refusal — every transient
`machine run`, the persistent lifecycle, `--env`, `--mount`, `--allow-host`,
exit-code propagation, the guest hostname, and the cold-launch dispatch window.
The Linux/Firecracker run is unaffected.

## Options

1. Default to 1 vCPU where the resolved backend is HVF.
2. Clamp to the backend's supported count with a warning rather than refusing.
3. Keep the refusal and change the global default to 1.

(1) preserves multi-vCPU on backends that support it and keeps the documented
commands working as written, which is what the docs promise.

## Detection

Found by the documented-surface e2e suite (#2910) on its first complete macOS
run. The hermetic lanes cannot see it: the command parses fine and only fails
when a guest is actually booted on that backend.

### LINKED ISSUE #2927: HVF vCPU ceiling of 4 is measured, not explained
`VmCapabilities::max_vcpus` for the HVF backend is `Some(4)`. That number is
measured, not derived, and the limit behind it is not understood.

## What was observed

On an Apple Silicon host (16 logical cores):

| `--cpus` | result |
|---|---|
| 1 | boots; `nproc` and `/proc/cpuinfo` both report 1 |
| 2 | boots; both report 2 |
| 4 | boots; both report 4 |
| 5, 6, 8 | guest agent never becomes reachable |

The failure is a timeout waiting for the agent, not a panic or a PSCI refusal.

## What has been ruled out

- **The address map.** The gap between `GICV3_REDIST_BASE` (`0x0808_0000`) and
  `SERIAL_MMIO_BASE` fits 124 redistributor frames at `GICV3_REDIST_STRIDE`, so
  the region is not being overrun at 5.
- **The hardware.** 16 logical cores available.
- **PSCI answering for absent CPUs.** `psci_cpu_on` returns
  `INVALID_PARAMETERS` only for targets no thread backs, and at `--cpus 5` all
  five have threads.

## Why it is recorded rather than derived

A principled-looking formula here would be a guess dressed as a limit. If the
real cause is a bug in the new per-vCPU threading — a race in bring-up, a lock
held across a vCPU exit, something in the shared device model — then deriving
the ceiling from the address map would hide that bug behind a number that looks
justified.

## Where to start

The guest agent not becoming reachable points at either the vsock device under
concurrent access from more vCPUs, or secondary bring-up wedging before
userspace. Capturing the guest console for a `--cpus 5` boot is the first step;
note that on HVF `console.log` flushes only at supervisor exit, so the VM has to
be stopped before the log is readable.

Follow-up to #2888 / #2926.

auser:
Closes #2888.
Closes #2924.
Closes #2927.

`--cpus` was accepted, recorded in the admitted plan, and then silently ignored:
the HVF VMM created exactly one vCPU whatever the caller asked for. A guest ran
on one CPU while its plan said four, with nothing to explain the difference.

## What changed

**The VMM creates the requested vCPUs.** The count reaches the process that makes
them (`HvfSupervisorConfig`), the device tree describes N `cpu@N` nodes with
`enable-method = "psci"`, the GIC redistributor region is sized to match, and
each secondary gets a thread. PSCI `CPU_ON` and `AFFINITY_INFO` are answered
against CPUs that actually exist.

**vCPU creation is ordered.** Hypervisor.framework assigns GIC redistributor
frames in `hv_vcpu_create` creation order. Allowing all creation threads to race
could swap two CPUs' frames, after which the guest could not match those CPUs to
their redistributors and never reached the agent. The frame identity check stays,
and creation now takes a turn so CPU N receives frame N.

**The ceiling comes from the host.** The apparent limit of four was the race
becoming more likely at larger counts, not a real platform ceiling. HVF now calls
`hv_vm_get_max_vcpu_count`; a host that cannot answer reports no backend ceiling
instead of using an invented constant. The portable request is still clamped
when a backend reports a real host limit, with a user-visible warning.

## Verification

- Live Apple Silicon guests at 1, 2, 4, 5, 6, 8, 16, and 32 vCPUs booted and
  reported their full count through both `nproc` and `/proc/cpuinfo`.
- The vCPU/GIC ordering tests preserve the invariant that PSCI never reports a
  CPU as available unless a backing thread exists.
- Workspace checks and the launch-mode regression suite cover the complete
  request-to-guest path. The separately tracked warm-dispatch budget gap is #2942.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -506,9 +506,10 @@ counts each machine's configured maximum rather than the balloon's
 current commitment. CPU is the partial half: a granted share wraps the VMM spawn in a
 systemd transient scope on Linux and the achieved tier is read back off
 `cpu.max`. On the in-house HVF VMM on macOS there is no host-level quota
-primitive, so the run loop enforces the share in-process using the vCPU
-thread's Mach CPU time; the achieved tier is read back from the scheduler's
-measured record and audited. libkrun has no in-process vCPU control, so a CPU
+primitive, so the run loop enforces the share in-process using the summed Mach
+CPU time of every vCPU thread — the sum, so the bound stays a bound on the
+machine rather than on one CPU of an SMP guest; the achieved tier is read back
+from the scheduler's measured record and audited. libkrun has no in-process vCPU control, so a CPU
 grant there stays `declared` and `--prod` refuses it. Wall clock is enforced
 by the per-VM supervisor on libkrun and HVF: the process that owns the guest
 for its whole life arms a timer from the admitted plan, and a workload that
--- a/Justfile
+++ b/Justfile
@@ -337,7 +337,8 @@ build-supervisors:
 # (mvm-build, mvm-core, ...) when you are about to boot a real VM; ordinary
 # check/test/clippy runs never need it. Release builds always rebuild.
 embed-refresh:
-    rm -rf target/*/build/mvm-cli-nested-target/host-vm-target
+    rm -rf target/*/build/mvm-cli/mvm-cli-nested-target/host-vm-target
+    rm -rf target/*/build/mvm-cli/mvm-cli-nested-target/aux-helper-target
 
 # Build the dm-verity-capable workload kernel into the local mvm cache.
 # Set MVM_KERNEL_SOURCE=download to use the hash-verified release artifact, or
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ mvmctl machine run --image python:3.12 \
 mvmctl machine run --image alpine -it -- /bin/sh
 
 # Give it resources; admit specific egress only (audited; TCP/22 always refused).
-# A request above the backend's vCPU ceiling is clamped to it, with a warning (HVF: 4).
+# A request above the backend's vCPU ceiling is clamped to it, with a warning.
 mvmctl machine run --image alpine --cpus 2 --memory 512M \
   --allow-host api.example.com:443 -- ./fetch
 
@@ -138,7 +138,7 @@ A persistent machine has a name and an on-disk spec: create once, start/stop/exe
 against it, reconfigure it, remove it when done.
 
 ```bash
-# A request above the backend's vCPU ceiling is clamped to it, with a warning (HVF: 4).
+# A request above the backend's vCPU ceiling is clamped to it, with a warning.
 mvmctl machine create web --image nginx --cpus 2 --memory 512M
 mvmctl machine start web
 mvmctl machine exec  web -- nginx -v
--- a/README.md
+++ b/README.md
@@ -109,8 +109,8 @@ mvmctl machine run --image python:3.12 \
 mvmctl machine run --image alpine -it -- /bin/sh
 
 # Give it resources; admit specific egress only (audited; TCP/22 always refused).
-# HVF currently supports exactly one guest vCPU; multi-vCPU backends may accept more.
-mvmctl machine run --image alpine --cpus 1 --memory 512M \
+# A request above the backend's vCPU ceiling is clamped to it, with a warning (HVF: 4).
+mvmctl machine run --image alpine --cpus 2 --memory 512M \
   --allow-host api.example.com:443 -- ./fetch
 
 # Build a Nix flake and run it transiently in one step
@@ -138,8 +138,8 @@ A persistent machine has a name and an on-disk spec: create once, start/stop/exe
 against it, reconfigure it, remove it when done.
 
 ```bash
-# HVF currently supports exactly one guest vCPU; multi-vCPU backends may accept more.
-mvmctl machine create web --image nginx --cpus 1 --memory 512M
+# A request above the backend's vCPU ceiling is clamped to it, with a warning (HVF: 4).
+mvmctl machine create web --image nginx --cpus 2 --memory 512M
 mvmctl machine start web
 mvmctl machine exec  web -- nginx -v
 mvmctl machine logs  web
--- a/crates/mvm-backends/src/driver/fc.rs
+++ b/crates/mvm-backends/src/driver/fc.rs
@@ -640,6 +640,11 @@ impl VmmDriver for FcDriver {
         // authenticated identity handshake still gates admission after
         // resume, so preloading changes placement of VMM work, not authority.
         VmCapabilities {
+            // None: this driver hands `vcpu_count` to Firecracker's
+            // machine-config API unmodified and imposes no ceiling of its own.
+            // `None` says exactly that — distinct from a backend that has a
+            // limit and has not declared one, which would read as `Some(large)`.
+            max_vcpus: None,
             pause_resume: true,
             snapshots: false,
             snapshot_capability: SnapshotCapability::Unsupported,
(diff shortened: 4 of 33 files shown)


auser:
Review note on the clamp and the signed plan — I think there's a gap.

`clamp_vcpus` runs in `resolve_launch` (`crates/mvm-cli/src/exec.rs:~1416`), but
the `ExecutionPlan` is admitted and signed **before** that, in the command
layer:

```
crates/mvm-cli/src/commands/vm/exec.rs
  786:  let admit_cpus = args.cpus;
  812:  let ctx = super::up::admit_plan_for_boot(... cpus: admit_cpus ...)
  1103: crate::exec::run_with_posture(req, audit.admit, &posture)   // -> resolve_launch -> clamp
```

So on `--cpus 8` against hvf's `Some(4)` the plan is signed for **8** and the
guest boots with **4**. That is the same shape as the bug this PR fixes — "a
guest ran on one CPU while its plan said four, with nothing to explain the
difference" — just moved from the VMM to the clamp. The console line explains it
to a human watching the terminal; the signed artifact still doesn't agree with
the machine.

`ReceiptInput.cpus` (~`exec.rs:1503`) takes `args.cpus` directly and has the
same issue.

Cheapest fix that keeps one source of truth: clamp once in the command layer
before admission, and let `resolve_launch` receive an already-clamped count —
`selected_backend` is in scope at line 780-781 (`admit_backend_kind` /
`admit_backend`), so `capabilities().max_vcpus` is reachable there. The
diagnostic line can stay where it is or move up with it.

Worth a test that a clamped launch admits a plan whose `cpus` equals what
booted; `cpu_on_never_reports_success_for_a_cpu_no_thread_backs` is the right
neighbour for it.

Context: I'd opened #2936 against the same user-visible bug before finding this
PR, and closed it — this PR is the better fix. The plan/receipt threading is the
one piece I'd carried that doesn't appear here.

auser:
Heads up — this went `DIRTY` because of #2928 (`refactor(backends): rename the
*_legacy driver modules to *_process`), which merged as `97a98f245c`. That's
mine, and the conflict is my doing.

It is a pure rename, so the resolution is mechanical. In `crates/mvm-backends/src/driver/`:

| was | now |
|---|---|
| `hvf_legacy.rs` | `hvf_process.rs` |
| `libkrun_legacy.rs` | `libkrun_process.rs` |
| `qemu_legacy.rs` | `qemu_process.rs` |

and the module paths with them (`driver::hvf_legacy::` → `driver::hvf_process::`,
`mod hvf_legacy;` → `mod hvf_process;`). No behaviour changed and no function
moved between files, so a rebase should be `git checkout --theirs` on the file
content plus a sed over the module path:

```sh
git rebase origin/main
# for each conflicted driver file:
git mv crates/mvm-backends/src/driver/hvf_legacy.rs \
       crates/mvm-backends/src/driver/hvf_process.rs   # if git didn't follow the rename
sed -i '' 's/hvf_legacy/hvf_process/g; s/libkrun_legacy/libkrun_process/g; s/qemu_legacy/qemu_process/g' \
  $(git grep -l '_legacy' -- crates/)
```

The only non-mechanical bit: I also rewrote those three module doc headers,
because each opened by describing a `VmBackend` impl that no longer exists in
the tree. If your branch edits those headers too, take yours for the content and
keep the `*_process` naming.

Happy to do the rebase for you if you'd rather not — say the word and I'll push
to `feat/2888-hvf-smp`. I didn't want to touch someone else's branch uninvited.

(Separately, the plan/receipt clamp note from my earlier comment still stands.)

auser:
Disregard the rebase offer — you'd already done it by the time I looked.

I had a rebase prepared and fully gated (`nextest --workspace` 12,262 pass,
test-support lane 3,860 pass, `clippy` clean, `xtask check-all` 61 gates,
`just check-gated`), but `--force-with-lease` correctly refused: the branch had
moved from `5463fd473c` to `be3f5530b7` while those were running, and pushing
would have dropped `refactor(hvf): make vCPU creation order structural rather
than remembered`. I threw mine away rather than reconcile them.

The PR now reads `MERGEABLE`, no failing checks, so the conflict my #2928 caused
is gone. Sorry for the churn.

The plan/receipt clamp note from my earlier comment is the one thing still worth
a look before this merges.

PR #2936, opened 2026-08-27
fix(cli): take the default vCPU count from the selected backend

auser:
`mvmctl machine run --image alpine -- sh -c 'echo hi'` fails on macOS today:

```
Caused by: the hvf backend supports exactly 1 vCPU, but this launch requests 2
```

Nobody asked for 2. `--cpus` carried `default_value = "2"`, and #2899 had just turned the HVF vCPU mismatch from a silent ignore into a hard refusal — correct on its own, but it made a fixed default fatal on the macOS default backend. The primary documented command fails on a bare invocation; `--cpus 1` works.

Reproduces on plain `main` at `efa77021c5`.

## The fix

`BackendKind::default_vcpus()` — an exhaustive match, so a new backend is a compile error here rather than a launch that refuses in the driver. HVF answers 1 because that is all it runs (`fdt.rs` emits a single `cpu@0`, PSCI implements no `CPU_ON`); everything else answers 2.

`--cpus` becomes `Option<u32>`: `None` = "unspecified, ask the backend", `Some(n)` passes through untouched. **`--cpus 2` on HVF still refuses loudly** — #2899's behaviour is preserved exactly, and only the meaning of *unspecified* changed.

Resolution happens where the backend is already known:

- `resolve_launch`, for the boot itself
- the signed `ExecutionPlan` (`admit_cpus`) via the already-selected `admit_backend_kind`, so the plan records what will actually boot
- the signed run receipt (`ReceiptInput`) and the preflight summaries, through a shared `effective_cpus` helper — so the number a dry run prints is the number a real run boots

## Verified live, not just in CI

Bare invocation prints its output; `--cpus 2` on HVF still fails with the #2899 message. Both run on a real macOS/HVF host.

## Gates

`fmt --all`, `clippy --workspace --all-targets` (zero warnings), `nextest --workspace` against an empty `MVM_HOME` (12,240 pass), the test-support lane (3,839 pass), `xtask check-all` (61 gates), `just check-gated`.

The test-support lane was the only one that caught two `LaunchShape` fixtures in `pool.rs` — that file's tests compile only under `--features test-support`.

## Follow-up, not in this PR

Making `--cpus 4` *mean* something on HVF is #2888: `HvfSupervisorConfig` has no `vcpus` field, `fdt.rs` emits one hardcoded `cpu@0`, and PSCI implements no `CPU_ON`. That needs all three plus per-vCPU threads and GIC sizing — a feature, not a fix.

### DIFF
--- a/crates/mvm-cli/src/commands/machine/mod.rs
+++ b/crates/mvm-cli/src/commands/machine/mod.rs
@@ -635,7 +635,7 @@ fn machine_run_spec(
         peer: Vec::new(),
         ai: None,
         ports: args.port.clone(),
-        cpus: args.run.cpus,
+        cpus: crate::exec::effective_cpus(args.run.cpus, args.run.hypervisor.as_deref()),
         memory: args.run.memory.clone(),
         mem_initial: None,
         profile,
--- a/crates/mvm-cli/src/commands/machine/runtime.rs
+++ b/crates/mvm-cli/src/commands/machine/runtime.rs
@@ -329,7 +329,7 @@ fn run_entrypoint_action(args: MachineRunArgs, resolved_flake_slot: Option<Strin
         source,
         stdin,
         timeout: args.run.timeout.unwrap_or(30),
-        cpus: args.run.cpus,
+        cpus: crate::exec::effective_cpus(args.run.cpus, args.run.hypervisor.as_deref()),
         memory_mib,
         from_workload_ir: args.from_workload_ir.clone(),
         agent_verb_override: args.run.agent_verb.clone(),
--- a/crates/mvm-cli/src/commands/machine/tests.rs
+++ b/crates/mvm-cli/src/commands/machine/tests.rs
@@ -482,7 +482,9 @@ fn transient_run_without_argv_is_rejected_at_dispatch() {
 #[test]
 fn run_defaults_match_the_lower_level_runner() {
     let args = parse_run(&["run", "--image", "alpine", "--", "true"]).expect("parse");
-    assert_eq!(args.run.cpus, 2);
+    // Unspecified: the selected backend's `default_vcpus` decides at launch,
+    // so parsing must not bake a count in here.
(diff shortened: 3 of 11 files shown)


auser:
Superseded by #2926, which closes #2924 (the same bug) properly.

I diagnosed this against `efa77021c5` and did not find #2926 before opening — my
fault, and the reason the two overlap.

#2926 is the better fix, and not just because it landed first:

- **It makes HVF actually run the vCPUs.** Mine defaulted HVF to 1 because that
  was all it could do; #2926 wires per-vCPU threads, PSCI `CPU_ON` /
  `AFFINITY_INFO`, the multi-CPU FDT and GIC sizing, so the ceiling is 4, the
  default of 2 boots, and #2924 disappears without a special default at all.
- **It clamps rather than refuses**, with a rationale mine did not have:
  `--cpus` is a portable flag meeting a per-host limit, so refusing makes the
  same script pass on Linux and fail on macOS for something the caller cannot
  act on.
- **It puts the ceiling in the right place.** `VmCapabilities::max_vcpus:
  Option<u32>` is where a per-backend capability belongs. I considered exactly
  that field and backed off because it meant touching 31 exhaustive
  constructors, and settled for `BackendKind::default_vcpus()` instead — which
  would now be a second, conflicting source of truth for the same fact. The
  churn was worth paying; #2926 paid it.

Landing mine on top would default HVF to one vCPU on a backend that can run
four, so it goes.

One thing worth carrying over rather than losing: the vCPU count also reaches
the **signed `ExecutionPlan`** and the **signed run receipt**. If #2926 clamps 8
to 4 at the launch seam, the plan and receipt should record 4 — what booted —
not 8. Otherwise a signed artifact describes a machine that never existed. Worth
a check on `admit_cpus` in `commands/vm/exec.rs` and `ReceiptInput` before
merge; if #2926 already handles it, ignore me.

PR #2989, opened 2026-08-28
test(docs): gate a release on every documented example running

auser:
Follow-on to #2984. That PR fixed one broken README example; this one makes it
impossible to cut a release while any of them is broken.

## Three gaps, each of which looked like coverage

**No live lane blocked a release.** `release.yml` needed `[bdd, build,
initramfs-image]`. `bdd` is the hermetic lane — it drives `--help`, refusal
paths, and cross-reads of in-tree docs, and boots no guest — so it cannot see a
documented command that parses and then fails at runtime, which is the entire
failure mode of #2984.

The lanes that *do* boot guests on both backends already existed and already
passed: `e2e-docs-linux` (Firecracker/KVM) and `e2e-docs-macos` (HVF). They live
in `ci-full.yml`, which runs at 04:51 nightly and gates nothing. The
launch-modes lane ran in no workflow at all.

Both moved into `.github/workflows/e2e-docs.yml` as a reusable `workflow_call`.
Extended CI calls it nightly; `release.yml` calls it too and the release job now
waits on it. Extracted rather than duplicated — two callers needing the same
lane had been getting different ones. Both backends, because #2984 reproduced on
HVF, the macOS default, and a Linux-only gate would have shipped it.

**Coverage was per command path, not per example.** `tiers.toml` tiers each
*command path*. `machine run` was `live`, so one live `machine run --image
alpine -- true` discharged the obligation for all twelve `machine run` variants
the README prints, including `-it`. The tier was accurate; the coverage was not.

`features/suites/s8_readme_contract/readme_examples.toml` is per invocation. All
38 distinct README `mvmctl` commands carry exactly one of `witness` (a `@live`
scenario), `hermetic_witness` + `reason` (really executed, but there is no guest
to boot — `doctor`, `template list`), or `exempt`.

The witness is checked, not trusted: same verb **and** at least the same flags,
with short and long spellings reconciled against the real clap arguments so
`--out` and `-o` compare equal. Values may differ — a scenario booting `--image
alpine` stands in for a README line booting `--image python:3.12` — since
pinning values would make the suite a transcription of the README rather than a
test of it.

**A witness could prove the opposite of its claim.** Matching on shape alone,
`-it` matched *"machine run refuses an interactive PTY without a terminal"* and
`bootstrap` matched *"--help lists the documented top-level verbs"*. Both name
the right verb with the right flags; neither runs the command, and the first
asserts it does not work. A `witness` must therefore be `@live`.

## What it found

Nine documented shapes had no live witness. New scenarios:

- **`--allow-host` with `--mount`** — the README's headline "install a dependency
  at boot" example. The allow-host scenario had no mount and the mount scenario
  had no egress; the mount is set up by the same launch that installs the egress
  policy, so covering them separately covered neither.
- **`--cpus`, `--memory` and `--allow-host`** on one launch.
- **`-vvv`** on a live launch — verbosity changes what the launch path logs and
  nothing had ever booted a guest with it set.
- **`machine reconfigure`** — documented in the same README block as the verbs
  around it, and the only one nothing ran.
- **`machine run --entrypoint --flake`** — the last step of "from dev loop to
  attested image". `build compile` and `machine build` were covered; booting the
  result was not.

Exemptions each name what does cover them: `bootstrap` (every live lane runs it
as setup, so a break fails the suite at the door), both `kernel build` arms, the
three `deps` commands, `run --peer`. One records a mechanism limit rather than a
cost: `generate template` *is* executed, by a scenario driving it through a
bespoke step rather than a quoted command line, so the structural check cannot
confirm the shape and is not asked to pretend.

## The gate's first run found a second broken README example

`mvmctl run --mode live --profile dev ./script.py` is in the README and does not
work. `--profile dev` clears the grant gate — the ProdSafe refusal does not fire
— and the script then dies on `mvmctl machine proc start` exit 1 against the
guest it just booted. That is the residual named in the *title* of #2887,
"fs/proc verb refusals were reported as protocol mismatches; **dev-mode launches
still cannot grant DevOnly verbs**", closed 2026-08-27 with the reporting half
fixed and this half not.

Nothing had run it: the existing scenario drives the same fixture *without*
`--profile dev` and asserts it is refused, which is the opposite claim.

It is tagged `@wip`, not `@live`, so a pre-existing defect does not red the
release lane on day one — and because it is not `@live` it cannot be a witness,
so the manifest records the line as **unproven rather than covered**. Retag it
and delete the exemption when the verb works.

## Red-first verification

Both failure modes were confirmed to fail before they pass:

- a new README example with no entry → *"these README examples have no entry in
  … so nothing proves they work"*
- the `-it` example pointed at the plain transient-run scenario → *"witness …
  does not exercise README example … naming the same verb is what let a broken
  `-it` ship while `machine run` looked covered"*

The second is #2984 expressed as a test failure.

## Results

- `just bdd` — 238/239 (1 env-skipped), including the new gate
- `just e2e-launch` — **22/22 scenarios, 127/127 steps, exit 0** on macOS/HVF
- `cargo fmt --all --check`, `clippy -p mvm-conformance --all-targets -D
  warnings`, `actionlint`, `xtask check-all` (61 gates) — clean
- All three workflows parse; neither e2e job carries a skip condition;
  `release.needs` is `[bdd, e2e-docs, build, initramfs-image]`

## Scope

README only, per the agreed scope. The same machinery extends to
`public/src/content/docs` — the extractor already walks it for the tier gate —
but that is 115 tiered paths and many more exemptions to adjudicate, best done
once this mechanism has proven itself.

A tag now waits on both live lanes. That is the slowest thing in `release.yml`,
deliberately: a tag that cannot wait for it is a tag published without evidence
that its own README works.


auser:
Came at this from the performance side and ended up designing the same gate before finding this PR — this is the better implementation and I've dropped mine. Binding each command to a named `witness` scenario is stronger than the text-matching I had, and the `exempt` entries carrying a reason someone can disagree with is the right call.

One gap I think is worth closing before this blocks releases, because it's the difference between a gate and a green light.

**A witness that didn't run is indistinguishable from one that passed.**

`readme_examples.toml` proves a witness *exists*. Nothing proves it *ran*. `should_run` filters scenarios through `scenario_gate_for_ci`, `record_gate` tallies whatever it declined, and `report_skips` prints:

```
[bdd] {} scenario(s) did NOT run:
[bdd] a green suite is not full coverage while any of these are nonzero.
```

…and then `main` exits on `execution_has_failed()` alone. A skip is not a failure. So a `witness` scenario that gets gated out — `NeedsFirecracker` on the macOS lane, `NeedsMemorySnapshot` where the backend can't snapshot, `NeedsNode`, `NeedsWorkloadKernel` — leaves its README example unexecuted, and the release lane goes green anyway. `scripts/e2e-documented-surface.sh:452` says the same thing to a human and also returns 0.

This is the exact shape CLAUDE.md already calls out for `check-claim-catalog`: "the gate proves a named witness *exists*, never that anything calls the code it tests." Same failure mode, one layer up.

It also interacts with the two-backend design, which I think is right and worth protecting. Some witnesses legitimately can't run on both — a dm-verity tamper scenario needs block+ext4, HVF-only snapshot scenarios need `MVM_BDD_SNAPSHOT`. Today that asymmetry is invisible: the Linux lane silently skips what macOS covers and vice versa, and nothing checks the union is complete.

Two things would close it:

1. `record_gate` records `(scenario_name, gate)` rather than the gate alone, and under an env flag set only in the release lane, a skip whose scenario is a declared `witness` fails the run unless an entry says which platform does cover it. `scenario_should_run` is already unit-testable without a cucumber runner, so the decision can be tested directly.
2. A third job `needs:` both lanes, reading their skip reports and failing when any witness scenario was skipped on *every* platform. That's what makes "covered by the other backend" true rather than asserted.

Happy to send this as a follow-up PR against your branch if you'd rather keep this one focused — the wiring and the registry are the hard parts and they're already done here.

Separately, `--profile dev` being exempted as `DOES NOT WORK TODAY` with the residual from #2887 named in full is exactly the right way to handle it. That entry is more useful than a passing test would have been.

auser:
Pushed the two gaps flagged on the original PR description.

## A skipped scenario no longer passes as a green lane

The suite printed a tally and a line saying a green suite is not full coverage
while it is nonzero. That is advice, and advice is not read at the moment it
matters: a runner that quietly loses a capability produces a green run that
proved less than the one before it, and nothing says so.

`MVM_BDD_STRICT_SKIPS` makes the suite exit nonzero on any skip the lane did not
declare, naming the reason and both ways to resolve it. Per-lane allow-lists,
spelled with stable `ScenarioGate` names:

| lane | tolerates |
|---|---|
| launch | `pending`, `needs-perf-budget-host` — verified against a real run as the only two it skips |
| documented surface | those plus `needs-memory-snapshot` (Firecracker genuinely reports `unsupported`; the macOS job sets `MVM_BDD_SNAPSHOT`), `needs-bundle-fixture`, `needs-tls-tunnel-client` |

Deliberately absent from both: `needs-workload-kernel`, `needs-guest-bin-dir`,
`needs-firecracker`, `needs-live-opt-in`, `needs-sdk-sidecar`, `needs-node`. If
one of those fires, the lane did not boot what it claims to boot, and that has to
be a failure rather than a footnote.

Unset, the variable changes nothing — a laptop run still gets the tally, not a
failure. Verified across all three modes: strict with the reason disallowed exits
1, strict with it allowed exits 0, default exits 0.

**Known:** the Linux documented-surface lane has not run under this policy — it
cannot run on this host. If its allow-list is short an entry, its first run fails
loudly naming exactly which one. That is the intended way to find out, and the
nightly Extended CI run surfaces it before any tag.

## The website is ratcheted, not adjudicated

461 distinct commands across 86 files. Hand-writing 400-odd exemptions would
manufacture the appearance of review rather than perform it — an exemption is a
claim someone has to be able to disagree with, and nobody can disagree with four
hundred of them written in an afternoon.

So `s29_doc_examples/docs_coverage.toml` is **computed, not authored**, using the
same rule as the README gate. Three properties:

- a covered command may not become uncovered
- a newly documented command must be classified before it merges
- one that becomes covered must be moved out of the debt list, so the number
  keeps meaning something

**Baseline: 170 covered, 267 uncovered.** That is the honest state of website
example coverage today — now visible and monotonic rather than unknown.
Regenerate with `MVM_UPDATE_DOCS_COVERAGE=1`.

Both arms verified red-first: a covered command losing its scenario fails with
*"were exercised by a scenario and no longer are"*; a newly documented one fails
with *"newly documented and are not in the coverage ledger"*.

This is weaker than the README's gate on purpose. The README says every example
is proven or someone said why not. The website says coverage cannot silently
decay and new documentation cannot skip the decision — which makes the remaining
267 ordinary incremental work against a number that does not drift.

## Results

- `just bdd` — 240 scenarios, 239 passed, 1 env-skipped
- `just e2e-launch` under the new strict policy — **22/22, exit 0**
- `cargo fmt --all --check`, `clippy -p mvm-conformance --all-targets -D warnings`, `xtask check-all` (61 gates) — clean

auser:
## Fixed: the README example now runs

`mvmctl run --mode live --profile dev ./script.py` works. It was recorded as
unproven; it is now a real witness.

### #2887's premise was stale

That issue was closed having escalated a policy question — fs/proc are DevOnly,
the grant comes from `prod_safe_verb_names()`, the override rejects anything
non-ProdSafe, so "a `machine run` launch can never carry a grant that authorizes
these verbs", and changing that was a claim-4 call needing an owner.

Against a live guest today, it already does:

```
$ mvmctl machine run -d --name devprof --image alpine --profile dev --ttl 300s
$ mvmctl machine fs read devprof /etc/hostname
localhost
$ mvmctl machine proc start devprof -- /bin/uname -s
ptok-9dd27e1b4d20e8fe16d2f620f9faa527
```

The grant half was fixed by something that landed since, and nothing updated the
issue or the `@wip` tag citing it. **No claim-4 decision was needed.**

### The actual defect

```
$ mvmctl machine proc start devprof -- uname -s
Error: Guest proc error (InvalidArgv): argv[0] "uname" must be an absolute path
```

Three parts disagreed:

- **README** documents bare names: `commands.start(["python", "/app/main.py"])`, `exec("uname", "-sr")`
- **SDK** forwards argv verbatim (`shell += ["--", *argv]`) — and cannot resolve host-side, it doesn't know the guest's filesystem
- **Guest** refused any non-absolute `argv[0]`, with no rationale comment; it arrived in a bulk crate-consolidation commit

Meanwhile `exec` — the other half of that same documented pair — *did* accept a
bare name, resolving through the image's `PATH`
(`ad_hoc_exec_resolves_a_bare_command_from_the_image_path`). Two sibling verbs,
two answers to the same input.

### The fix

`resolve_argv0` reuses the search `exec` already uses:

- absolute path → unchanged
- **bare name** → looked up in the image's declared `PATH`, FHS order as fallback
- **relative path** (`./run`) → still refused: it resolves against a working directory the same request may be setting

The property the old rule protected is intact — what reaches `execve` is still an
absolute path chosen before the fork. The request's own `PATH` is deliberately
not consulted (the caller supplies that env; the image is not the caller), and a
source-text test asserts the resolver reads no process environment.

`build_command_rejects_relative_argv0` asserted a bare `echo` was refused — but a
bare name is not a relative path, and refusing it *was* the bug. Split so the
refusal it was named for is preserved and the thing it actually tested is gone.

`files.write` is restored to the fixture, as the issue asked once fixed.

### The strict-skip gate paid for itself

Its first real run caught that the `@dir_share` `--mount` scenario — **the witness
for both README `--mount` examples** — was skipped for want of
`MVM_BDD_DIR_SHARE=1`, on a host that serves virtio-fs. Two documented examples
had a witness that never executed. Both lanes now opt in, and tolerate the skip
only where the backend genuinely cannot (Firecracker has no virtio-fs device).

It also surfaced that my three new egress scenarios gated on a single ICMP reply;
one went red on a dropped packet while the same command passed immediately after.
They now assert on `ping -c 3`'s exit code — enough to show egress, not enough to
fail a release on one lost packet.

### Results

- `just e2e-launch` — **23/23 scenarios, 130/130 steps, exit 0** (was 22 executed; the `--mount` witness now runs)
- `just bdd` — 240 scenarios, 239 passed, 1 env-skipped
- `cargo nextest run --workspace` — 12692/12692
- fmt, clippy, `xtask check-all` (62 gates) — clean

PR #3032, opened 2026-08-31
test(e2e): rebuild the per-VM supervisors before the launch gate boots

auser:
## The gap

`scripts/e2e-launch-modes.sh` builds `mvmctl` and `xtask`, then signs and boots.
It never relinks `mvm-hvf-supervisor`, and `cargo build --bin mvmctl` does not
either — so the supervisor sits at whatever revision last built it while the
host side keeps gaining config fields.

The supervisor parses the config `mvmctl` pipes to its stdin with
`deny_unknown_fields`. One field newer than the binary and it exits 1 before a
guest exists:

```
Error: parse HvfSupervisorConfig JSON from stdin

Caused by:
    unknown field `vcpus`, expected one of `kernel`, `cmdline`, `memory_mib`, …
```

which the launch path reported only as `hvf supervisor exited before writing its
PID file`, pointing at a guest console that is empty precisely because no guest
ever ran.

`scripts/e2e-documented-surface.sh` already rebuilds the helpers for exactly this
reason — its comment records the 33 scenarios that failed that way, none of them
naming the stale binary. The launch gate never got the same step and is just as
exposed: it failed its warm launch against a supervisor several days stale, with
nothing to go on but a PID file that never appeared.

## The change

`just build-supervisors` after the `mvmctl` / `xtask` builds. cargo makes it a
fingerprint check when the binaries are already current, so the cost of doing it
unconditionally is that check — and presence is not freshness, which is why this
is not a test that the files exist.

Two placement details, both load-bearing:

- **Before `mvmctl env sign`, not after.** Relinking drops the
  Hypervisor.framework entitlement, and macOS kills an unentitled binary with
  SIGKILL. That surfaces through the *same* sentence as the stale-binary case,
  so the signing comment now says which status means which: `signal: 9` is
  unentitled, `exit status: 1` is stale.
- **Not `just embed-refresh`.** That clears the nested musl cross-compile of the
  *embedded* host-vm binaries — a different set — and this gate boots no builder
  VM that would rebuild them.

The mechanism itself is documented once, in the sibling script; this comment
points there rather than keeping a second copy to drift.

## Verification

`bash -n` clean. `shellcheck` reports only a pre-existing SC2155 on line 44,
untouched here.

A full `just e2e-launch` against this change, in a cold worktree, is running as
this PR opens; I will post the result. Worth stating plainly: **CI cannot check
this file.** The launch gate needs a host that boots a guest and no hosted macOS
runner does, so a local run is the only verification this change will ever get.

The prior run of this same gate — same machine, same suite, with the supervisor
rebuilt by hand instead of by this step — passed end to end, including
`transient_and_persistent_lifecycle_over_hvf`. This change automates the step
that made that run possible.

## Related

Complementary to #3027, not redundant with it. That one makes the failure name
its own cause wherever `mvmctl` runs, including outside this gate. This one stops
the gate from meeting it at all. Either alone leaves a real hole.

### DIFF
--- a/scripts/e2e-launch-modes.sh
+++ b/scripts/e2e-launch-modes.sh
@@ -154,6 +154,10 @@ echo "==> building mvmctl + host helpers"
 # cargo makes this a no-op when they are already current, so always doing it
 # costs a fingerprint check. The signing step below must follow it: a re-linked
 # supervisor loses its Hypervisor.framework entitlement.
+#
+# Not `just embed-refresh`: that clears the nested musl cross-compile of the
+# *embedded* host-vm binaries, a different set, and this gate boots no builder
+# VM that would rebuild them.
 echo "==> building the per-VM host helpers"
 just build-supervisors
 


auser:
Verification result, as promised in the description.

`just e2e-launch` against this change, on macOS 26 / Apple Silicon:

```
==> building the per-VM host helpers
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
==> signing VMM binaries
  ✓ …/target/debug/mvmctl: signed
  ✓ …/target/debug/mvm-libkrun-supervisor: signed
  ✓ …/target/debug/mvm-hvf-supervisor: signed
…
[Summary]
23 scenarios (23 passed)
130 steps (130 passed)
```

Exit 0, including `transient_and_persistent_lifecycle_over_hvf`.

Two things that run demonstrates beyond "it is green":

- **The step costs 0.4s when the binaries are current** — the two `Finished` lines
  above are the fingerprint checks the comment claims, not a rebuild. Doing it
  unconditionally is cheap.
- **Signing still covers the supervisor afterwards.** The ordering is the point:
  had the rebuild landed after `env sign`, the relink would have stripped the
  Hypervisor.framework entitlement and every HVF boot in the suite would have
  died with `signal: 9`.

An earlier attempt at this same run failed, and it is worth recording why it was
not this change: it died ~30 minutes in, in the warm step, on
`mvmctl kernel build --which workload` — the Stage 0 `nix build` exceeded the
builder VM's 1800-second wall-clock cap and was killed. `~/.mvm/cache` held no
workload kernel, so the run had to build one from scratch while the host sat at
load average 11-14. The step added here had already completed successfully
28 minutes before that failure. Nothing this change touches can reach a nix
build running inside a guest.

That the kernel disappeared from a warm `~/.mvm` between two runs of this gate is
its own question, unrelated to this PR and not investigated here.

auser:
Closing: superseded by #3025, which landed the same step in `main` as
`ad45d00f3d` while this was open.

That commit already carries everything here — `just build-supervisors` after the
`mvmctl` / `xtask` builds, placed before `mvmctl env sign` so the relink cannot
strip the Hypervisor.framework entitlement, the explicit "not `just
embed-refresh`" rationale, and the `signal: 9` (unentitled) vs `exit status: 1`
(stale) distinction in the signing comment. It also does more than this PR:
pre-builds `mvm-client`'s test targets so the library-seam phase stops compiling
mid-run, adds the README's `--image python:3.12` scenarios, and corrects
`MIN_SCENARIOS` to 24.

Nothing here is worth salvaging on top of it, so this branch goes away rather
than being rebased into a duplicate.

The verification in the comment above stands as corroboration, with one honest
caveat: that run was against this branch, based on `d590651f` — the equivalent
change, not the merged commit. It showed the step costing 0.4s as a fingerprint
check when the binaries are current, signing still covering the supervisor
afterwards, and 23/23 scenarios and 130/130 steps passing on macOS 26 / Apple
Silicon.

The one observation from it that outlives this PR and belongs to neither: a
`just e2e-launch` failed earlier when `~/.mvm/cache` turned out to hold no
workload kernel, so the warm step had to build one and the Stage 0 `nix build`
blew the builder VM's 1800-second cap under load. Why a warm cache lost its
kernel between two runs of this gate is unexplained and worth someone's
attention separately.

PR #3061, opened 2026-09-01
feat(builder): move the persistent HVF builder onto the disk transport

auser:
Finishes Stage C of `specs/plans/2026-08-31-remove-virtio-fs.md` for the persistent HVF builder. Stacked on #3056 (the gate + the guest half); rebases onto main once that lands.

## What changes

The persistent HVF builder exchanges jobs and artifacts over two raw block devices instead of four virtio-fs shares. `persistent_builder_spec` declares no shares, so `check-no-virtio-fs` **drops `builder_runner/spec.rs` from its pinned table entirely** — the signal the plan names for Stage C working. The gate now ratchets 22 sites across 10 files, none of them a builder spec.

The two builder cmdlines collapse into one: a persistent builder now boots the same contract as a one-shot, `vda`–`vdd` in the same order, so a second copy of the same string is gone rather than maintained.

Per-dispatch lifetime lives in a new `persistent_builder_transport` module. Both dispatch clients use it — `PersistentBuilderVm::run_build` (what `mvmctl build` routes through) and the `persistent-builder submit` verb.

## Three things the plan did not anticipate

**Readiness could not become connect-polling.** The backend binds the host UDS during VM setup, *before* the guest boots, so `connect` succeeds against a guest whose vsock driver is not up. The bridge only closes that connection when the guest kernel answers `OP_RST`, so early in boot a probe neither connects-and-fails nor gets refused — it hangs open and reads as ready, which is the same silent hang the move was meant to remove. It is a round trip instead: `submit_workload_status` for the nil UUID. Any well-formed reply proves the loop is accepting and framing. Existing request, no side effect, no protocol change.

**A fifth coordinated change, and the one that would have shipped broken.** The guest tars `/out` — and only `/out` — onto the output disk, but both host stagers rendered a `cmd.sh` writing to `/job/<job_id>/out`, which under the disk transport is the guest's own input stage, never read back. The four planned changes alone would have produced a build that reported success and emitted nothing.

**The install arm has the same shape and no host-side fix.** It hardcodes `/job/<job_id>/out`. Rather than lose a claim-11 sealed volume's SBOM and CVE sidecars silently, an install dispatch on a disk-backed session is refused with a message naming the missing half.

## Live validation — macOS 26.5.2, Apple Silicon

Proven end to end against a rebuilt builder image: the input disk attaches as `vdc` and the output as `vdd`; the guest binds `/job`, `/work` and `/mvm-bins` off the input disk (Nix evaluated a flake at `path:/work`, only possible if the workspace crossed on that disk); the new readiness path returns with no `dispatch.ready` file anywhere the host can see; and **two dispatches into one live session** each rewrote the input disk, re-staged `/job/<job_id>` off it, and streamed stderr back.

**Not proven: a completed build**, and therefore `read_dispatch_artifacts` against a real output disk (unit-tested only). The blocker is not the transport — the session's `mvm-network-endpoint` dies with the command that spawned it, so the guest loses egress mid-build. The one-shot builder never hits this because its endpoint's lifetime *is* the VM's. Tracked in the plan; it is claim-10 lifecycle design and wants its own change.

## Pre-existing gaps this uncovered

The persistent HVF builder had never booted. Reaching Stage C at all required fixing three things unrelated to virtio-fs, each by calling the helper the one-shot builder already calls: no runtime overlay was resolved, no FlowMux identity drive was minted, and no network endpoint was spawned. A fourth was this branch's own: the input pack must use `stage_filtered_work_input`, or 45 GB of `target/` lands on the input disk and fills the store image mid-extraction.

`trusted_build_egress` gains one allow-listed caller — the same trusted builder tier as the one-shot runner, differing only in lifetime. No workload microVM boots from this policy.

## Verification

`just check-gated`, `cargo nextest run --workspace` (12854 passed), `cargo test --workspace --doc`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `cargo run -p xtask -- check-all` (63/63) all clean.

### DIFF
--- a/crates/mvm-backends/src/driver/fc.rs
+++ b/crates/mvm-backends/src/driver/fc.rs
@@ -1300,6 +1300,8 @@ mod tests {
                 log_path: "/tmp/console.log".into(),
             },
             trusted_builder: false,
+            // Not a persistent builder: nothing here outlives its launcher.
+            builder_egress_endpoint: None,
             plan_binding: None,
         }
     }
--- a/crates/mvm-backends/src/driver/hvf.rs
+++ b/crates/mvm-backends/src/driver/hvf.rs
@@ -238,6 +238,9 @@ fn relay_supervisor_config_with_handoff(
             .collect(),
         vsock: true,
         trusted_builder_egress: spec.trusted_builder,
+        // Only a VM that outlives its launcher sets this; everyone else spawns
+        // their own endpoint and stays alive to parent it.
+        builder_egress_endpoint: spec.builder_egress_endpoint.clone(),
         console_log: paths.console_log.clone(),
         pid_file: paths.pid_file.clone(),
         workload_exit: paths.workload_exit.clone(),
@@ -1081,6 +1084,7 @@ mod tests {
 
     fn spec_with(kernel: KernelImage, vsock: Vec<VsockPort>, blocks: Vec<BlockDev>) -> VmmSpec {
         VmmSpec {
+            builder_egress_endpoint: None,
             name: "w".into(),
             kernel,
             initramfs: Some("/img/initrd.cpio".into()),
--- a/crates/mvm-backends/src/driver/hvf_restore.rs
+++ b/crates/mvm-backends/src/driver/hvf_restore.rs
@@ -153,6 +153,11 @@ pub fn hvf_child_restore_config(
     Ok(HvfSupervisorConfig {
         // Same tier as the parent it forked from.
         trusted_builder_egress: parent.trusted_builder_egress,
+        // Never inherited. The request names the *parent's* VM, state dir and
+        // identity drive; a child adopting it would spawn a second endpoint
+        // over the parent's socket and overwrite the identity drive of a VM
+        // that is still running off it. A restored child is not a builder.
+        builder_egress_endpoint: None,
         kernel: parent.kernel.clone(),
         cmdline: parent.cmdline.clone(),
         memory_mib: parent.memory_mib,
@@ -324,6 +329,7 @@ mod tests {
     fn parent_config(state: &Path, disks: Vec<HvfDisk>) -> HvfSupervisorConfig {
         HvfSupervisorConfig {
             trusted_builder_egress: false,
+            builder_egress_endpoint: None,
             kernel: state.join("Image"),
             cmdline: Some("console=ttyAMA0 root=/dev/vda ro".into()),
             memory_mib: 512,
@@ -434,6 +440,34 @@ mod tests {
         // The half this function exists for is still done: a restored child is
         // itself checkpointable only if its own launch config is on disk.
         assert!(dir.join("supervisor.json").is_file());
+    }
+
+    #[test]
+    fn a_restored_child_never_inherits_the_parents_egress_endpoint_request() {
+        // The request names the *parent's* vm, state dir, socket and identity
+        // drive. A child that inherited it would spawn a second endpoint over
+        // the running parent's socket and rewrite the identity drive that
+        // parent's guest is still authenticating against — so this is the same
+        // family as the relay drop the module docs describe, not an
+        // optimisation.
+        let tmp = tempfile::tempdir().unwrap();
+        let dir = tmp.path();
+        let anchors = materialized(dir);
+        let mut parent = parent_config(dir, vec![]);
+        parent.builder_egress_endpoint =
+            Some(mvm_vmm::host::hvf_supervisor::BuilderEgressEndpoint {
+                vm_name: "the-parent".into(),
+                state_dir: dir.join("parent-state"),
+                socket: dir.join("parent.sock"),
+                identity_drive: dir.join("parent-identity.ext4"),
+            });
+
+        let cfg = hvf_child_restore_config(&parent, &anchors, &request("child", dir)).unwrap();
+
+        assert!(
+            cfg.builder_egress_endpoint.is_none(),
+            "a restored child must not adopt its parent's endpoint request"
+        );
     }
 
     #[test]
--- a/crates/mvm-backends/src/driver/libkrun.rs
+++ b/crates/mvm-backends/src/driver/libkrun.rs
@@ -666,6 +666,8 @@ mod tests {
                 log_path: "/tmp/console.log".into(),
             },
             trusted_builder: false,
+            // Not a persistent builder: nothing here outlives its launcher.
+            builder_egress_endpoint: None,
             plan_binding: None,
         }
     }
--- a/crates/mvm-backends/src/driver/qemu.rs
+++ b/crates/mvm-backends/src/driver/qemu.rs
@@ -682,6 +682,8 @@ mod tests {
                 log_path: "/state/w/console.log".into(),
             },
             trusted_builder: false,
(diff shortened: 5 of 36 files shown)


auser:
Rebased onto main now that #3056 has merged (`--onto` against the old base, so only this branch's three commits replay — verified by content, `git diff origin/main...HEAD` shows no #3056 files).

Two things changed in the rebase, both because main moved underneath:

**1. The repack now uses `repack_input_disk_in_place`.** Main gained that helper while this was in flight, and it is strictly safer than what this branch had. This branch used `pack_input_disk` with the current file length as a floor, which truncates before rewriting, grows the file if a job exceeds it, and never flushes. The growth case is the dangerous one: the guest's `DiskImage` captured the device length once at open and zero-fills reads past it, so a grown archive is read *short* — and `tar` reports success on a short archive, so the build runs whatever the truncation left with no error anywhere. The in-place form refuses instead, and syncs so the guest's pread cannot race the host page cache. The shrink-floor test became a capacity-unchanged test, plus a new one asserting an over-capacity dispatch is refused.

**2. Independent agreement on the readiness finding.** Main's plan revision reached the same conclusion this branch did from the other direction — that a bare connect-poll reports ready for a guest that never started — and cited `hvf.rs`'s existing agent probe saying so in as many words. It offered two fixes: a `Ping`/`Pong` wire pair (a second guest change and image rebuild), or deleting the wait and letting the first dispatch prove readiness (moves the failure out of `start()`). What shipped here is a third that costs neither: a round trip on the existing `submit_workload_status` request. The plan's conflicted section is resolved keeping both accounts.

Also merged rather than replaced in the plan: main's QEMU-builder findings. Worth flagging that they correct a claim in this PR's description — **`qemu_builder.rs` still serves shares, and QEMU is the Linux auto-detect default builder**, so "no builder constructs a share" is true only of the `VmmSpec` path, not of every builder. `check-no-virtio-fs` does not count `qemu_builder.rs` because virtiofsd is spawned rather than attached through the spec.

Re-verified on the rebased tree: `just check-gated`, `cargo nextest run --workspace` (12868 passed), doctests, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `cargo run -p xtask -- check-all` (63/63).

The merge blocker is unchanged: the session's network endpoint dies with the command that spawns it, so no build completes through the persistent HVF builder yet.

auser:
**Converted to draft deliberately — please do not re-queue.** This PR was added to the merge queue; it is not ready, and a completed live build through the persistent HVF builder still does not exist. Draft is the only lever that reliably keeps it out (`--disable-auto` reports "already queued", and a queued branch cannot be pushed to, which is also blocking a commit I have locally).

## Resolved since the last update: the endpoint lifetime

The diagnosis I posted earlier was wrong in a way worth recording. The endpoint was **not** leaking and was **not** missing a detach — `spawn_network_endpoint` already calls `setsid()`. It self-reaps: `mvm_hostd::parent_death::exit_when_orphaned`, armed in the endpoint's own `main`, watches its parent via kqueue and exits with **status 0** the instant it is reparented. A clean exit is why its log simply stopped after a successful FlowMux handshake with nothing logged.

That guard is deliberate — the endpoint holds a workload's secrets in the clear and must not be left serving as an orphan. A persistent session is just the first case where the VM outlives the command that started it, so parent-death stopped being a good proxy for VM-death.

Fixed by keeping the guard and giving it the right parent: the **supervisor**, whose death is the VM's. No key material crosses the new wire — the endpoint's config carries the host signer *by value* and `supervisor.json` is persisted beside the VM, so the supervisor mints the FlowMux identity itself from the same `~/.mvm/keys/host-signer.ed25519` the config already names by path, and writes the identity drive before it opens disks (it is one of them). Policy and the empty secret set are applied at that call site rather than read off the wire, so nothing crossing it can widen what the endpoint serves; a restored child never inherits the request.

**Verified live** on macOS 26.5.2 / Apple Silicon: the endpoint stays up with the supervisor as its parent (`ps` confirms the ppid), serves the guest's egress through a whole `nix build` (303 log lines, streams opening throughout), and self-reaps when the supervisor exits.

## The remaining blocker, newly identified

`CONNECTION_IDLE_TIMEOUT = 60s` in `crates/mvm-vmm/src/vmm/vsock_transport.rs`. The HVF host-dial bridge's `evict_idle_at` closes any connection idle for 60 seconds, so a `nix build` that evaluates quietly for over a minute loses its host-side dispatch connection: the client reports `dispatch ended without Result frame after 0 stderr chunks` while the build carries on normally inside the guest. Confirmed twice, including with a connection the guest had not yet read — so it is the idle policy, not anything job-specific.

A one-shot builder never hits it: it holds no dispatch connection across a build, it powers off and the host reads the output disk. Only a persistent session keeps a control channel open across a long quiet operation.

Not a one-line raise — the timeout sits next to `MAX_CONNECTIONS` as a shared host resource boundary and also governs workload console streams. Options are a per-service idle policy (the builder dispatch port is long-lived and legitimately quiet) or a keepalive on the dispatch channel. Tracked in the plan.

auser:
**Done — taken out of draft. The live build completes.**

Two `nix build` dispatches into one persistent HVF session on macOS 26.5.2 / Apple Silicon, both `exit_code: 0`:

- first: **465s** cold, streaming build output back live
- second: **797ms** off the warm store — which is the entire point of a persistent session

Each read `vmlinux` (8,636,424 bytes), `rootfs.ext4` (11,609,088 bytes) and the verity sidecars off the output disk into its own artifact dir. That closes the last unproven leg: `read_dispatch_artifacts` against a real output disk, from a real build.

## The last blocker took two fixes, not one

`CONNECTION_IDLE_TIMEOUT` is applied in **two independent layers**, and I only found the second by watching a live build fail after fixing the first:

1. `host_dial_bridge::evict_idle_at` — closes the host socket.
2. `vsock_transport::evict_idle_connections_at` — sends the guest an `OP_RST`.

Exempting the port in the bridge alone still let the transport tear the stream down. The signature was distinctive: the guest logged `write StderrChunk failed` repeatedly and then `flake command exited code=0 elapsed_ms=464997` — the build ran to completion and wrote a full artifact set while the host had already given up on it.

Both layers now take the same per-port exemption: builder control ports exempt, console data ports still evictable. What makes that safe is that idle was never the mechanism that reclaims a dead peer — `drain_host`'s EOF arm is, and it still runs for every connection. Idle only ever described a stream that is alive and quiet, which for a request/response control channel is indistinguishable from working normally. A test pins that a peer hangup still reclaims an exempt port, since that is the property the exemption rests on.

## Gates

`cargo nextest run --workspace` (12,873 passed), `cargo test --workspace --doc`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `cargo run -p xtask -- check-all` (63/63) — all clean.

## Scope note for review

Stage C's own change is the disk transport. Most of this PR is pre-existing breakage found underneath it: the persistent HVF builder had **never booted**. It was missing a runtime overlay, a FlowMux identity drive, and a network endpoint; its endpoint self-reaped when orphaned; and its dispatch channel was severed by idle eviction in two places. Each was only visible after fixing the one before it. Those fixes are separate commits if you want to read them independently.

PR #3064, opened 2026-09-01
fix(vmm): declare the vCPU ceiling each VMM boots, not the wire type's

### LINKED ISSUE #3051: machine run --cpus above the Firecracker ceiling is not clamped; the launch fails
The README documents clamping:

> A request above the backend is vCPU ceiling is clamped to it, with a warning.

That holds on HVF. On Firecracker the value is passed straight through and the launch dies:

```
Caused by:
    0: Firecracker API PUT /machine-config
    1: PUT /machine-config failed: HTTP 400
       {"fault_message":"An error occurred when deserializing the json body of a
        request: invalid value: integer `9999`, expected u8 ..."}
```

`vcpu_count` is a `u8` in the Firecracker API, so anything above 255 is rejected at deserialization before any ceiling logic runs.

## Evidence

`features/suites/s31_launch_e2e/cli_launch_modes.feature:231` — *"a vCPU request beyond the backend ceiling is clamped and reported"*, running `machine run --image alpine --cpus 9999`.

- **HVF**: passes — the request is clamped and the guest boots.
- **Firecracker**: fails — HTTP 400, exit 1.

Backend-divergent behaviour on a documented flag, where the documentation describes the HVF behaviour without qualification.

## Worth deciding

Clamping presumably belongs above the backend, where the ceiling is known, rather than in each driver. Whatever the placement, the `u8` boundary means Firecracker cannot even receive an out-of-range request to clamp it — so the clamp has to happen before the API call.

### LINKED ISSUE #3065: libkrun declares the u8 wire ceiling as its vCPU max; 128 vCPUs aborts the supervisor
The libkrun driver declares `max_vcpus: Some(u32::from(u8::MAX))`
(`crates/mvm-backends/src/driver/libkrun.rs:281`) — the wire ceiling of a `u8`,
not a count libkrun will boot. So the clamp above the backend reduces an
oversized `--cpus` to 255 and the launch then dies inside the supervisor.

This is the same defect as #3051, on the other backend. #3051 fixed Firecracker
by declaring the ceiling the VMM actually boots on; libkrun is untouched because
its real ceiling is not yet known and its own API does not report one.

## Evidence

x86_64 Linux/KVM, libkrun 1.18.0 / libkrunfw 5.4.1, 8-core host, current `main`:

```
$ mvmctl machine run --hypervisor libkrun --image alpine --cpus N -- sh -c 'nproc'
N=2     -> 2        exit 0
N=8     -> 8        exit 0
N=9     -> 9        exit 0
N=16    -> 16       exit 0
N=32    -> 32       exit 0
N=64    -> 64       exit 0
N=128   -> Error: libkrun supervisor exited before binding vsock socket
                  (status: signal: 6 (SIGABRT) (core dumped)); console.log empty
N=255   -> Error: libkrun supervisor exited before binding vsock socket
                  (status: exit status: 1); console.log empty
N=9999  -> [mvm] libkrun supports at most 255 vCPU(s); 9999 requested,
                  booting with 255
           Error: ... exit status: 1
```

So a `--cpus` request the CLI reports as granted ("booting with 255") is one the
backend cannot boot, and the user is told the clamp succeeded before the launch
fails with an unrelated-looking message about a vsock socket.

## Why the obvious fix does not work

Firecracker's fix was to declare the VMM's real limit. HVF asks the host
(`hv_vm_get_max_vcpu_count`). libkrun exposes what looks like the same thing —
`krun_get_max_vcpus()`, already bound in
`crates/deps/libkrun-sys/src/libkrun_bindings.rs:319` and called from nowhere —
but on this host it answers **4096**, while 128 aborts. Whatever that number is,
it is not the bound a clamp can use.

## Worth deciding

- Where the real bound comes from: a probed constant like Firecracker's, the
  host CPU count, or a fail-closed refusal above some threshold. Bisecting
  64→128 on one host does not establish a portable ceiling, and a number
  measured on one machine and hardcoded is how HVF ended up with `Some(4)`
  recording a bug rather than a limit (#2927).
- Whether the SIGABRT at 128 is libkrun's own limit, a guest-memory layout
  consequence, or KVM's `KVM_CAP_MAX_VCPUS` on this host. The console log is
  empty and the supervisor's abort message was not captured, so this has not
  been diagnosed.
- libkrun is an explicit opt-in on Linux (`--hypervisor libkrun`; auto-detect
  picks Firecracker) and the default on macOS 13–25, where none of the above has
  been measured.

auser:
Closes #3051. Closes #3065.

`mvmctl machine run --image alpine --cpus 9999` booted on HVF and failed on
Firecracker and libkrun, which made a documented, portable flag
backend-divergent — the README says a request above the backend's vCPU ceiling
is clamped to it with a warning, and that described one backend of three. The
contract is that an over-large request is *granted the maximum*: never refused,
never fatal.

## The clamp was not missing

It lives above the backends and asks each one for its ceiling —
`exec::resolve_launch` for transient runs (the function `mvmctl run`, `mvmctl
machine run` and the warm-pool resolver all reach) and
`mvm_client::clamp_vcpus_for_backend` for the persistent OCI path. Both read
`capabilities().max_vcpus`, so neither had to change here. On current `main`,
live on x86_64 Linux/KVM, it fires on both failing backends exactly as it does
on HVF:

```
[mvm] firecracker supports at most 255 vCPU(s); 9999 requested, booting with 255
  0: Firecracker API PUT /machine-config
  1: PUT /machine-config failed: HTTP 400 {"fault_message":"Machine config
     error: The number of vCPUs must be greater than 0, less than 32 ..."}

[mvm] libkrun supports at most 255 vCPU(s); 9999 requested, booting with 255
  libkrun supervisor exited before binding vsock socket ...
  (status: signal: 6 (SIGABRT) (core dumped))
```

What was wrong is the number each clamped *to*. Both drivers declared
`max_vcpus: Some(u8::MAX)`, reasoning from the wire format — Firecracker's
`vcpu_count` and libkrun's `krun_set_vm_config` both take the count as a byte,
so 255 is the largest value that survives the call. True of both, and the limit
of neither. Each clamp faithfully produced a count its VMM would not boot, and
the launch died one step further along than it used to. The #3051 report quotes
the older, pre-clamp Firecracker message (`invalid value: integer 9999, expected
u8`); today's failure is the second one. Same defect.

HVF never had this problem because its ceiling is asked of the host
(`hv_vm_get_max_vcpu_count`) rather than derived from a type.

## The change

Each driver declares the ceiling it will actually boot on, and holds the value
it hands the VMM to that same constant — so the declared ceiling and the value
on the wire cannot drift apart again, which is the shape of this bug. The
reporting clamps stay exactly where they were, above the backends, where the
request is still the user's and there is somebody to warn; because they read the
declared ceiling rather than carrying their own, correcting the declaration
fixed both without touching either. The driver-side bounds are the floor under
callers that never passed through them, and match what the qemu driver already
did for its own.

One test had to change: `mvm-client`'s
`firecracker_vcpus_are_clamped_to_a_count_the_vmm_boots` asserted the clamp
landed on `u8::MAX`, pinning the defect rather than the contract. It and its new
libkrun sibling now name the counts the VMMs boot.

**Firecracker: 32.** Probed against the API rather than read out of
Firecracker's source: `vcpu_count: 32` answers 204, `64` and `255` answer 400,
`9999` fails to deserialize (v1.14.1).

**libkrun: 64, measured rather than asked**, because nothing will say.
`krun_set_vm_config` accepts every count from 1 to 255 and returns 0; the ones it
cannot honour abort inside `krun_start_enter`, which is why the failure surfaced
as a supervisor that died before binding its vsock socket, naming neither vCPUs
nor the count. `krun_get_max_vcpus()` answers **4096** — KVM's own
`KVM_CAP_MAX_VCPUS` forwarded verbatim, which the host confirms — while libkrun
aborts at 65 regardless. The cliff sits exactly at 64/65 and does not move when
the guest is given eight times the memory, so it is a constant and not a
resource limit. A number measured on one host and hardcoded is how HVF once
carried `Some(4)` recording a bug rather than a limit (#2927); that memory
control is the part that makes this one safe to write down.

## Witness

`features/suites/s31_launch_e2e/cli_launch_modes.feature:231` — *"a vCPU request
beyond the backend ceiling is clamped and reported"* — now passes on Firecracker
and libkrun as it does on HVF. Live on x86_64 Linux/KVM, with this branch built:

```
$ mvmctl machine run --image alpine --cpus 9999 -- sh -c 'echo clamped-and-booted'
[mvm] firecracker supports at most 32 vCPU(s); 9999 requested, booting with 32
clamped-and-booted                     # exit 0

$ ... --cpus 9999 -- sh -c 'echo $(nproc)/$(grep -c ^processor /proc/cpuinfo)'
32/32                                  # a real machine shape, not just a message
$ ... --cpus 32   -- sh -c 'nproc'  ->  32     # at the ceiling: unclamped, unwarned
$ ... --cpus 2 --memory 512M -- ...  ->  2/2   # ordinary requests unchanged

$ ... --hypervisor libkrun --cpus 9999 -- sh -c 'nproc'
[mvm] libkrun supports at most 64 vCPU(s); 9999 requested, booting with 64
64
$ ... --hypervisor libkrun --cpus 65 -- sh -c 'nproc'
[mvm] libkrun supports at most 64 vCPU(s); 65 requested, booting with 64
64
$ ... --hypervisor libkrun --cpus 64 -- sh -c 'nproc'  ->  64
$ ... --hypervisor libkrun --cpus 2  -- sh -c 'nproc'  ->  2
```

Unit witnesses, one pair per driver, in `crates/mvm-backends/src/driver/`:
`an_oversized_vcpu_request_is_held_to_the_ceiling_the_driver_declares` reads the
ceiling off `capabilities()` and asserts the value handed to the VMM lands on it,
so an edit cannot move one without the other;
`a_vcpu_request_within_the_ceiling_reaches_{the_api,libkrun}_unchanged` asserts
the bound bounds rather than rewrites.

## The fourth backend, checked

qemu was verified rather than assumed: `--hypervisor qemu --cpus 9999` boots and
exits 0 (the guest reports 64). It declares no ceiling, so the reporting clamp
stays silent and the driver's own `clamp(1, u8::MAX)` carries it — the user gets
a working machine with no warning about the count they did not get. That is the
cosmetic half of this bug on a dev/test-only backend that carries no workload
and is never auto-selected, and it is not a failure, so it is left alone rather
than fixed against a ceiling nothing has measured.

## Gates

`cargo fmt --all -- --check`, `cargo nextest run --workspace`,
`cargo test --workspace --doc`,
`cargo clippy --workspace --all-targets -- -D warnings`, `just check-gated`.


auser:
Sibling defect found while verifying this one, filed separately as #3065: libkrun declares the same u8 wire ceiling, boots 64 vCPUs and aborts at 128, and its own `krun_get_max_vcpus()` reports 4096 — so the fix used here does not transfer without its own investigation.

PR #3073, opened 2026-09-01
fix(build): stop the bootstrap helper rebuilding the payload it is running

### LINKED ISSUE #3067: Builder-VM bootstrap helper is built without embed-host-bins, so it cannot bootstrap
On a cold builder-image cache, `mvmctl` spawns a second `mvmctl` — the builder-VM bootstrap helper — to refresh `~/.mvm/cache/builder-vm/<arch>/`. `builder_vm_bootstrap_helper_build_command` (`crates/mvm-build/src/libkrun_builder.rs:2593`) builds it with:

```rust
.args(["build", "-q", "--bin", "mvmctl"]);
```

No `--features embed-host-bins`. The helper therefore refuses at `host_binaries::extract`, which is the one thing it was spawned to do:

```
Error: building the source-checkout builder VM image via root-dir Stage 0
Caused by: extract embedded host-vm binaries: this mvmctl was built without the
           embedded Linux host binaries, so it cannot bootstrap a builder VM.
           Rebuild with `just embed`
Error: builder VM
Caused by: extracting artifacts from builder sandbox: builder VM bootstrap helper
           .../target/mvm-builder-vm-bootstrap/debug/mvmctl exited with 1
```

Reproduced on the Linux/KVM box with a scratch `MVM_HOME` and a source checkout:

```sh
cd examples/sleeper
MVM_HOME=/tmp/scratch mvmctl machine build --builder qemu
```

It only bites on a cold cache — a contributor with a populated per-arch cache dir never reaches the helper, which is presumably why it has survived.

There is a test pinning that exact argument list (`libkrun_builder.rs:6328`), so the gap is asserted rather than accidental; whoever fixes this should work out whether the featureless build was deliberate before changing it.

**Workaround**: build a helper with the feature and pin it via the supported env lever, which skips the rebuild entirely:

```sh
CARGO_TARGET_DIR=target/mvm-builder-vm-bootstrap cargo build --features embed-host-bins --bin mvmctl
export MVM_BUILDER_VM_BOOTSTRAP_BIN=$PWD/target/mvm-builder-vm-bootstrap/debug/mvmctl
```

Found while validating #3066; unrelated to that change.

auser:
Closes #3067.

On a cold builder-image cache a source-checkout `mvmctl` spawns a second `mvmctl` — the builder-VM bootstrap helper — to refresh `~/.mvm/cache/builder-vm/<arch>/`. It built that helper without `--features embed-host-bins`, so the helper refused at `host_binaries::extract` doing the one thing it was spawned for.

#3066 hit this and worked around it with `MVM_BUILDER_VM_BOOTSTRAP_BIN`; #3062 then added the feature to the build command in passing. That closes the refusal and opens a different hole — the helper build now silently requires the pinned zig + musl Rust toolchain, and spends minutes reaching a build-script panic for anyone who has never run `just toolchain-embed`. This finishes the job from the other end.

## The helper is only ever needed for its payload

Every reason to spawn one reduces to: get hold of an `mvmctl` carrying the embedded Linux host binaries. So the first question is whether the running binary already carries them — and for anyone who followed `just embed`, it does. Compiling a second `mvmctl` to obtain a payload already loaded in this process is the bug; fixing the build command does not address it.

`mvm-build` cannot read the embed table (it sits below the crate that owns it), so `mvmctl` declares the answer once at startup alongside the existing `register_*` seams, and `resolve_builder_vm_bootstrap_bin` returns `current_exe()` on the strength of it. The ladder is now: explicit `MVM_BUILDER_VM_BOOTSTRAP_BIN`, the running binary, a fresh cached helper, then a build. The declaration defaults to `false`, so a test binary or library embedder keeps the path it had.

An embedded `mvmctl` now bootstraps with no cargo invocation at all — `mvmctl bootstrap` short-circuits the re-exec and runs in-process; cold-cache auto-bootstrap spawns itself.

## The refusal moved to before the compile

`crates/mvm-cli/build.rs`'s toolchain resolution moved down to `mvm_build::embed_toolchain`, where the bootstrap can reach it, and grew `try_` variants that report instead of panicking. The preflight runs the same two resolutions the build script does — pinned zig, pinned Rust with the musl target — before spawning cargo.

The build script still `#[path]`-includes the file and still panics; it has nowhere to fall back to. It is deliberately *not* preflighted inside `build.rs`, because the embed arm restores from the content store before compiling anything and a cache hit needs no toolchain — an unconditional check there would break the restore path CI and releases run on.

## Every refusal names every way out

`MVM_BUILDER_VM_BOOTSTRAP_BIN` is the supported escape hatch and no error mentioned it. Both helper-acquisition failures now end with the same tail — run `just embed` so this binary carries the payload itself, or point the env var at one that does — and the toolchain refusal adds `just toolchain-embed` and quotes the underlying reason. The zig message previously said `pip install ziglang==<pin>`, which installs zig but not the musl Rust targets.

## One fork bomb closed on the way past

`auto_bootstrap_builder_vm_image` had no guard against being reached from inside a bootstrap; making the running binary the helper puts that door in plain sight. Spawned bootstraps now carry `MVM_BUILDER_VM_BOOTSTRAP_ACTIVE`, and a process holding it reports the cold cache instead of forking another child.

## Second commit is a pure move

`libkrun_builder.rs` was at 1491 of the 1500 production lines `check-file-size` allows, so this did not fit beside it — and the helper is not libkrun's anyway, since `ensure_builder_vm_image` reaches it on every backend. It now lives in `mvm_build::builder_vm_bootstrap`.

`bootstrap_helper_build_command_uses_isolated_target_dir` still pins the dedicated `target/mvm-builder-vm-bootstrap` dir, which is load-bearing: a plain `cargo build` into the main target dir swaps `mvmctl`'s embedded payload out.

## Validation

`just check-gated`, `cargo nextest run --workspace` (12893 passed), `cargo test --workspace --doc`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo run -p xtask -- check-all` (63 gates).

Live on the Linux/KVM box: cold-cache `mvmctl machine build --builder qemu` from `examples/sleeper` with a scratch `MVM_HOME` and `MVM_BUILDER_VM_BOOTSTRAP_BIN` unset — the env var is the workaround, so setting it would hide the thing under test. The helper build never runs (`target/mvm-builder-vm-bootstrap` is never created) and the "cannot bootstrap a builder VM" refusal never appears; the run proceeds straight into the guest kernel build. Full-run result added below when it lands.


auser:
## Live validation complete

Two cold-cache runs on the Linux/KVM box, both against the head of this branch, both `{"exit_code":0}` in ~27 minutes:

```
mvmctl machine build --builder qemu     # from examples/sleeper
MVM_HOME=<fresh scratch dir>            # cold builder-image cache
MVM_BUILDER_VM_BOOTSTRAP_BIN            # unset — it is the workaround
```

The host's shared `~/.mvm/cache/builder-vm/x86_64/` carries no image, so the opportunistic seed declines and auto-bootstrap is genuinely reached rather than short-circuited.

Each run built the builder image from the checkout via Stage 0 — `.mvm-provenance.json` records `source_kind: source_checkout_stage0` — then built the sleeper workload and registered it as a template revision with `rootfs.ext4` + `vmlinux`.

`target/mvm-builder-vm-bootstrap` was never created, and the "cannot bootstrap a builder VM" refusal never appeared: the running binary was the helper.

The one `cargo` compile visible in those logs is `mvm-network-endpoint`, a separate pre-existing source-checkout auto-build, not the bootstrap helper.

PR #3078, opened 2026-09-01
Remove committed agent memory

auser:
## Summary

- ignore the root `.agent-memory/` directory
- remove the tracked agent notes, documentation, and Just recipes
- remove the `check-agent-notes` xtask gate and its command registrations
- remove the dedicated rollout plan and delivery record

## Verification

- `cargo clippy --workspace -- -D warnings`
- `cargo clippy -p xtask --all-targets -- -D warnings`
- `cargo test -p xtask` (683 passed)
- `git diff --check`

A full workspace test run was attempted, but an existing test cleanup collision removed Cargo’s active worktree-local target directory. No test assertions failed before that infrastructure error; PR CI will run the complete matrix.


auser:
Closing so the merge resolves deliberately rather than by queue order. **Reversible — the branch `chore/remove-agent-memory` is untouched; reopen if this is the call you want.**

This PR and #3081 were both sitting in the merge queue, and they are mutually exclusive: this one deletes `.agent-memory/` and the `check-agent-notes` gate, #3081 adds a note to that directory.

### They do not cleanly exclude each other, which is the problem

Merging both produces exactly **one** conflict — a modify/delete on `specs/plans/2026-09-01-committed-agent-notes-and-asserted-absence.md`. Everything else merges silently. Resolving that one conflict the obvious way yields this tree:

| | |
|---|---|
| `.agent-memory/notes/witness-reachability-gate-measured-and-refuted.md` | still tracked |
| `/.agent-memory/` in `.gitignore` | line 55 |
| `xtask/src/check_agent_notes.rs` | deleted |

A tracked note orphaned inside a gitignored directory, with no README beside it and no gate validating it. CI would not flag that — it is a stray markdown file. Whichever of the two lands second gets ejected and re-run, and the surviving orphan is not something either PR intended.

### Why this one and not #3081

#3081 is green and already queued, and the note it adds is a real finding — the measurement showing a witness-reachability gate finds nothing across all 84 `fn:` witnesses, so nobody re-runs that afternoon.

The staged version of this same change was cleared from a worktree earlier today under the same reasoning, so this is consistent with that call rather than a new one.

### If the removal is what you want

Reopen and rebase onto `main` after #3081 lands, so the deletion covers all six notes rather than five. And consider whether `/.agent-memory/` belongs in `.gitignore` at all: it makes the directory unwritable rather than merely empty, which forecloses the decision instead of reversing it.

PR #3082, opened 2026-09-02
fix(egress): bound simultaneous raw TCP tunnels per workload

auser:

| | |
|---|---|
| Files this branch changes | 3 |
| Of those, still differing from `main` | **2** |
| Behind `main` by | 430 commits |
| Last commit | 2026-08-14 |

### Commits

- \`8934fac99c\` fix(egress): bound simultaneous raw TCP tunnels per workload

### Read CI here with care

This branch is **430 commits behind `main`** and has not been rebased or re-run since 2026-08-14.

### DIFF
--- a/crates/mvm-hostd/src/supervisor/dns_handler.rs
+++ b/crates/mvm-hostd/src/supervisor/dns_handler.rs
@@ -413,13 +413,19 @@ mod tests {
 
     #[tokio::test]
     async fn rate_guard_denies_when_bucket_is_empty() {
-        let guard = EgressRateGuard::new(0, 1);
+        let guard = EgressRateGuard::builder()
+            .requests_per_sec(0)
+            .max_inflight(1)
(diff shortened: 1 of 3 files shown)


auser:
Closing: the change has nowhere left to apply.

Both files this branch modifies existed at its merge base and are **deleted on `main`**:
`crates/mvm-hostd/src/supervisor/raw_egress.rs` and `crates/mvm-hostd/src/supervisor/dns_handler.rs`.
The commit bounds simultaneous raw TCP tunnels inside `raw_egress.rs`, so there is nothing
to bound and nothing to rebase onto. Its third file already matches `main`.

The branch `fix/2519-raw-tcp-tunnel-bound` stays on the remote — it was a detached HEAD before
this pass and the commit is worth keeping addressable — but the work needs re-deriving against the
current egress code, not reviving.

PR #3083, opened 2026-09-02
fix(lifecycle): dispatch pause and resume through VM owner

auser:

| | |
|---|---|
| Files this branch changes | 5 |
| Of those, still differing from `main` | **4** |
| Behind `main` by | 107 commits |
| Last commit | 2026-08-27 |

### Commits

- \`8137a7ac4f\` fix(lifecycle): dispatch pause and resume through VM owner

### Read CI here with care

This branch is **107 commits behind `main`** and has not been rebased or re-run since 2026-08-27.

### DIFF
--- a/crates/mvm-client/src/local.rs
+++ b/crates/mvm-client/src/local.rs
@@ -104,24 +104,36 @@ impl LocalBackend {
         infos.into_iter().map(|i| to_state(i, None)).collect()
     }
 
-    /// Whether this client drives the hermetic in-memory mock backend. The mock
-    /// has no live Firecracker socket and no guest agent, so the snapshot path
-    /// swaps in canned bytes and skips the guest-facing signals.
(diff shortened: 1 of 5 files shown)


auser:
Closing: already landed, under a different commit subject.

Rebasing this branch onto `main` drops the commit entirely — it applies to nothing. Checking
line by line, **68 of 68 substantive lines** this commit adds to `crates/mvm-client/src/local.rs`
and `crates/mvm-core/src/client/dto.rs` are present on `main` today, with zero absent.
`lifecycle_backend_for` and the marker-based owner resolution are both there.

Worth noting for the other PRs in this batch: matching commit *subjects* against `main`'s history
said this branch was 0-of-1 landed, which was wrong. A squash merge renames the commit, so the
subject test gives false negatives. The rebase-and-measure test above is the reliable one.

PR #3085, opened 2026-09-02
fix(sdk): slim the host-services cdylib closure

auser:

| | |
|---|---|
| Files this branch changes | 10 |
| Of those, still differing from `main` | **10** |
| Behind `main` by | 70 commits |
| Last commit | 2026-08-28 |

### Commits

- \`6962128506\` fix(sdk): slim the host-services cdylib closure

### Read CI here with care

This branch is **70 commits behind `main`** and has not been rebased or re-run since 2026-08-28.

### DIFF
--- a/crates/mvm-cli/Cargo.toml
+++ b/crates/mvm-cli/Cargo.toml
@@ -29,7 +29,7 @@ mvm-client.workspace = true
 mvm-mcp.workspace = true
 # `mvmctl compile` reads the Workload IR (mvm_contract::ir, re-exported from
 # mvm_contract::ir) and routes it through the compile pipeline.
-mvm-sdk.workspace = true
+mvm-sdk = { workspace = true, features = ["remote-deploy"] }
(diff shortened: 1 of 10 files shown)


auser:
Closing: superseded by #3096, which lands the part of this branch that was still outstanding.

Half of this branch already landed — `main` has `deploy-remote` as an off-by-default feature
carrying the only `mvm-http` edge, and `cargo tree -p mvm-sdk -e no-dev,no-build` is clean today.

What had not landed was the **gate** pinning that. #3096 extracts
`xtask/src/check_sdk_cdylib_deps.rs` verbatim onto current `main` and registers it in
`check-all`, verified to fail on `default = ["deploy-remote"]` rather than merely pass.

Extracted rather than rebased because this branch's `Cargo.toml` changes now conflict with the
shipped feature layout — it uses `remote-deploy`, main uses `deploy-remote` — and a rebase would
have discarded them anyway. The gate itself needed no changes.

PR #3086, opened 2026-09-02
fix(xtask): make the kernel-reads gate see hand-rolled cache paths

auser:

| | |
|---|---|
| Files this branch changes | 3 |
| Of those, still differing from `main` | **1** |
| Behind `main` by | 22 commits |
| Last commit | 2026-08-31 |

### Commits

- \`374c188ece\` fix(xtask): make the kernel-reads gate see hand-rolled cache paths

### Read CI here with care

This branch is **22 commits behind `main`** and has not been rebased or re-run since 2026-08-31.

### DIFF
--- a/specs/REFACTOR-STATUS.md
+++ b/specs/REFACTOR-STATUS.md
@@ -216,6 +216,25 @@ Last updated: 2026-09-01
 
 ## Completed
 
+- [x] **Kernel cache moved out of the Stage 0 blast radius — PR #3038.**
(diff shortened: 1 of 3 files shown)


auser:
Closing: already landed.

The code change matches `main` exactly — the kernel-reads gate fix shipped. The only remaining
difference is a `specs/REFACTOR-STATUS.md` entry for PRs #3038 and #3028, and `main` already
records both. Nothing here is new.

PR #3094, opened 2026-09-02
docs(plan): declare parse-tier backing (+1 more)

auser:

| | |
|---|---|
| Commits ahead of `main` | 2 |
| Files this branch changes | 4 |
| Of those, still differing from `main` | **3** |
| Behind `main` by | 50 commits |
| Last commit | 2026-08-30 |
| Prior PR | #3010 — **already merged**; what is here is what the branch accumulated afterwards |

### Commits

- \`bf19d06a73\` docs(plan): declare parse-tier backing
- \`3ab039f240\` test(bdd): ratchet the parse tier so it cannot grow unnoticed

### Read CI here with care

This branch is **50 commits behind `main`** and has not been rebased or re-run since 2026-08-30.

### DIFF
--- a/crates/mvm-conformance/src/lib.rs
+++ b/crates/mvm-conformance/src/lib.rs
@@ -427,6 +427,90 @@ mod live_home {
     }
 }
 
+/// The tier manifest, read only for the fields the ratchet needs.
+#[derive(serde::Deserialize)]
+struct TierManifest {
+    command: Vec<TierEntry>,
+}
+
+#[derive(serde::Deserialize)]
+struct TierEntry {
+    tier: String,
+}
+
+/// How many documented command paths sit at the weakest tier.
+///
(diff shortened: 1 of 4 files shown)


auser:
Closing: already landed, and this version would regress it.

`main` has the ratchet — `PARSE_TIER_PIN` at `crates/mvm-conformance/tests/steps/doc_examples.rs:1429`
— and the plan doc at `specs/plans/2026-08-30-doc-example-parse-tier-promotion.md` shipped with it.
Of the 60 code lines this branch adds, 59 are already on `main`.

The one line that is not is `const PARSE_TIER_PIN: usize = 65;`. `main` pins **62**. Landing this
would raise the pin by three and loosen a ratchet whose whole purpose is that the parse tier cannot
grow unnoticed — the opposite of what the branch set out to do. Its snapshot of the count is simply
older than main's.

PR #3100, opened 2026-09-02
chore(vmm): delete the host-side virtiofsd daemon

auser:
Closes the Stage C item the whole plan opened on.

## What goes

`crates/mvm-vmm/src/host/virtiofsd.rs` — 382 lines spawning and supervising a `virtiofsd` process per share — and its only consumer, the QEMU **workload** driver's share arm.

**This is the line that started the plan.** The original finding was `virtiofsd.rs:253`:

```rust
.args(["--sandbox", "none"])
```

virtiofsd's own namespace/seccomp confinement, disabled, with no comment, no ADR, and no mention in the commit that introduced it. That flag no longer exists anywhere in the tree, because the process it configured no longer exists.

## Why it is safe now

The deletion was gated on *both* builders reaching the disk transport. Both have: the QEMU builder (#3066), then the persistent HVF builder (#3061). That left the QEMU workload driver as virtiofsd's last consumer.

Unreachability was **proven before deleting, not assumed**: every `VmmSpec.shares` assignment in the tree — `spec_map`, both builder specs, all five drivers, the mock, every test fixture — is empty. No construction anywhere produces a share, so `if !spec.shares.is_empty()` could never run.

Also removed: the `memory-backend-memfd` object, the `-numa` node, the `vhost-user-fs-pci` loop, `qemu_virtiofs_socket_path`, the `VirtiofsdGuard` field and its teardown, `mvm-vmm`'s `which` dependency (there solely to locate the binary), and `mvm-build`'s re-export.

`mem_arg` went with the memfd backend it sized. Worth stating plainly since it looks alarming: `-m` is pushed independently and a test pins it, so QEMU still gets its memory — the orphaned local was the only casualty.

## The test was replaced, not dropped

`argv_maps_virtio_fs_shares` asserted the mapping that no longer exists. Deleting it would have left nothing pinning the absence, so it became `argv_emits_no_virtio_fs_arguments_even_for_a_spec_that_carries_shares`: it builds a spec **with** a share and asserts the argv still carries no virtiofs wiring. The arm is pinned as *deleted* rather than merely unreached, so re-introducing the mapping fails the suite instead of silently restoring a guest-driven FUSE server on the host.

## The gate

`check-no-virtio-fs` drops from **46 sites across 14 files to 41 across 13**, with the `virtiofsd.rs` row deleted outright.

It has **not** reached FFI-only rows, which the plan names as the end state. Still pinned: the libkrun builder's Stage 0 `RootDir` path (it boots a virtio-fs root and predates the disk transport), the libkrun C FFI, and the now-dead HVF device model — `HvfVirtioFsShare`, the `hvf.rs` mapper, `virtio.rs`'s VirtioFs MMIO device, `kernel_boot.rs`'s attach. Those map and attach nothing today and are the natural next deletion.

## Verification

`RUSTFLAGS="-D warnings" just check-gated` (CI's flags, not the recipe default — an unused local is a warning locally and an error there, which bit the previous PR), `cargo nextest run --workspace` (12,949 passed), `cargo run -p xtask -- check-all` (66/66), `cargo fmt --all -- --check`.

No live run: this deletes an unreachable path, and the absence is pinned by a test and the ratchet rather than by a boot.


auser:
Closing as **superseded by #3072**, which landed the same work and took a better approach.

Both PRs deleted `crates/mvm-vmm/src/host/virtiofsd.rs` and the QEMU workload share arm — two sessions working the same Stage C item in parallel. #3072 got there first.

The approaches differed, and #3072's is the one worth keeping. I made QEMU silently ignore a spec carrying shares, with a regression test asserting no virtiofs argv is emitted. **#3072 makes `boot` refuse such a spec outright**, mirroring Firecracker's `boot_rejects_virtio_fs_shares`. Refusing beats ignoring: a caller that somehow constructs a share now gets an error naming the problem instead of a VM that quietly booted without it.

I checked for anything salvageable before closing rather than assuming duplication:

- `virtiofsd.rs` — gone on main ✓
- QEMU share arm — gone on main, with a stronger refusal ✓
- `which` dependency in `mvm-vmm` — already dropped on main ✓
- My regression test — superseded by the refusal test, which is the better pin

Nothing unique is lost. The two PRs stacked behind this one (#3106, #3109) remain necessary — main still carries `HvfVirtioFsShare` and the 1,108-line `virtio_fs.rs` FuseServer — and are being rebased onto `main` directly.

PR #3106, opened 2026-09-02
chore(hvf): delete the now-dead virtio-fs share wiring

auser:
Stacked on #3100 (retarget to `main` once that lands).

## What goes

The HVF backend can no longer express a share at all: `HvfVirtioFsShare` and its config field, the `hvf.rs` mapper, the supervisor's pass-through, the restore path's copy, and `kernel_boot.rs`'s FDT nodes and device construction. The VirtioFs device is now unreachable **by construction** rather than by policy.

## The MMIO hole stays, deliberately

`RNG_MMIO_BASE` and `RNG_IRQ` are computed *from* `FS_MMIO_BASE` and `MAX_VIRTIOFS_SHARES`. Collapsing the freed range would silently move the entropy device to a different address and SPI — changing the device tree a guest boots against, and the layout a saved snapshot was captured under. That is a boot and restore-compatibility break that would pass every test and fail live.

So the constants stay, documented as a reserved hole, and `RNG_MMIO_BASE` now derives from it:

```rust
const RNG_MMIO_BASE: u64 = FS_MMIO_BASE + (MAX_VIRTIOFS_SHARES as u64 + 1) * MMIO_STRIDE;
```

Same address as before (`base + 16*stride`), with the dependency stated once instead of restated in two places. Reclaiming the range is a deliberate, separately-validated change.

## What this deliberately does *not* do

I started on the `virtio.rs` VirtioFs device itself and pulled back once the size became clear. Behind it sit its test suite, the **1,108-line `virtio_fs.rs` FuseServer**, and that module's **fuzz target** — whose crate carries its own lockfile that a dep change has to refresh.

Deleting a guest-driven parser's fuzzing harness deserves its own reviewable change rather than riding along in a wiring removal. The plan now carries it as the next item. I checked: the FUSE parser is cited as a witness in neither ADR-001 nor `model/claims.toml`, so that change will not touch the claims ledger.

## Also closed: two stale plan entries

- The `virtiofsd.rs` deletion was listed **twice**; both boxes were open. The second is now ticked with a note.
- The **Stopgap** section — restore `--sandbox namespace` as a hedge while the removal ran — is moot and marked so. There is no daemon left to pass a sandbox argument to, which is the outcome it was hedging against not happening.

## The gate

`check-no-virtio-fs` drops three rows: **41 sites across 13 files → 38 across 10**.

## Verification

`RUSTFLAGS="-D warnings" just check-gated` (CI's flags — it caught an orphaned constant this change created), `cargo nextest run --workspace` (12,949 passed, including the 69 HVF boot/layout tests), `cargo run -p xtask -- check-all` (66/66), `cargo fmt --all -- --check`.

No live run: this deletes wiring that nothing could reach, and the MMIO addresses are unchanged by construction rather than by measurement.


auser:
Rebased onto `main` and retargeted (was stacked on #3100, now closed as superseded by #3072).

**#3072 landed the same work as #3100 while this stack was open** — two sessions on the same Stage C item. It also took the better approach: QEMU now *refuses* a spec carrying shares, mirroring Firecracker's `boot_rejects_virtio_fs_shares`, rather than ignoring one as mine did. This PR is unaffected by that overlap — main still carries `HvfVirtioFsShare` and the HVF wiring this deletes.

Conflict resolution, both in the plan doc:

- The HVF wiring checkbox — took this branch's version, which ticks it and splits the device deletion into its own item (#3109).
- The Stopgap section — took **main's** version. Both said the same thing, but main's additionally names `specs/plans/2026-08-31-virtiofsd-sandbox-parity.md` as a superseded standalone plan, which I did not know about.

I also corrected the plan's status line, which claimed pre-#3072 counts, and made the end state explicit: **FFI-only rows, not zero.** 16 of the remaining sites declare libkrun's C API — it exports `krun_add_virtiofs` whether or not we call it — and three more are the `VirtioFsShare` type plus the QEMU and Firecracker tests that assert a share is *refused*.

Re-verified on the rebase: `check-no-virtio-fs` clean at 38 sites across 10 files, `cargo nextest run --workspace` 12,979 passed, `check-all` 67/67, `RUSTFLAGS="-D warnings" just check-gated` clean.

auser:
**Release coordination — this is wanted in 0.18.0, please land it.**

I'm driving the 0.18 release ([#3117](https://github.com/tinylabscom/mvm/pull/3117), plan at `specs/plans/2026-09-02-cut-the-0-18-release.md`). Ari asked specifically for virtio-fs to be gone in this release, so this PR and #3109 are both in scope rather than deferred.

**Status:** this PR is `CLEAN` and fully green — nothing is blocking it that I can see. It is the third bullet of Stage C's open boxes in `specs/plans/2026-08-31-remove-virtio-fs.md` ("Now-dead HVF share plumbing").

**Two things I verified so you don't have to re-derive them:**

1. **No claim witness dies.** The four share-related witnesses — `dir_share_two_part_defaults_ro`, `relay_config_maps_dir_shares_with_dax_and_read_only`, `enforce_admitted_shares_refuses_unadmitted_or_mismatched` (MVM-SEC-01) and `virtiofs_mount_flags_keep_workspace_read_only` (MVM-SEC-02) — live in `mvm-cli`, `mvm-hostd` and `mvm-build`, outside what this stack deletes. `check-conformance` stays green.

2. **The docs are handled.** Six stale virtio-fs claims across README.md, `boot-flow.md`, `nix-for-mvm.mdx`, `nix-flakes.md` and `sandbox-parity-status.md` are fixed on #3117 — they were already wrong before this stack, so they are not your problem, but they would have become conspicuously wrong after it.

**One sequencing request.** The release gate now requires a recorded macOS documented-surface run whose digest covers `crates/`, so anything touching `crates/` has to land *before* the final evidence run. Please land this (and #3109 behind it) rather than after — otherwise the two-hour suite gets repeated. Ping me here when it's in the queue and I'll sequence around it.

PR #3109, opened 2026-09-02
chore(vmm): delete the virtio-fs device and its FUSE server

auser:
Item 4 of 4. Stacked on #3106 (retarget as the stack lands).

## The argument, in its strongest form

The fuzz target I'm deleting states its own premise at the top of the file:

> The `VirtioFs` device hands guest-supplied bytes straight to `FuseServer::dispatch`. The guest is untrusted, so a byte pattern that panics the parser is a guest-triggerable **host** DoS.

That premise became false when the last wiring went in #3106 — no device hands it guest bytes any more. So the parser is **deleted rather than fuzzed**: the DoS surface is removed, not defended. That is the plan's whole thesis ("a block device is a byte array with no protocol for a guest to attack") reaching its end state for this device.

## What goes

- The `VirtioFs` MMIO device and its `RunDevice` impl
- The 1,108-line `virtio_fs.rs` FuseServer
- Its test suite, including the differential reference-walk equivalence tests
- The shared-memory-region MMIO registers — a DAX-window concern, virtio-fs only
- Sixteen FUSE/FS constants
- The `fuse_dispatch` fuzz target, its `security.yml` step, and the fuzz crate's **lockfile entry** (refreshed, not left stale — a detached fuzz lockfile breaks the merge queue)

## Claims ledger: untouched

I checked before deleting a security harness. The FUSE parser is cited as a witness in neither `specs/adrs/001-microvm-security-posture.md` nor `model/claims.toml`, so no claim loses a witness here.

## One gate needed a deliberate adjustment

`the_fuzz_lane_fits_inside_its_own_timeout` asserts a floor on the resolved target count. Removing a target took it 15 → 14 and tripped the assertion.

The floor's purpose is catching a **parser that fails to resolve targets** and therefore reports a trivially-fitting budget — not mandating a particular number of them. A legitimately deleted target should lower it. I updated the floor and said so in the comment, which had separately drifted to claiming "seventeen".

## The gate

`check-no-virtio-fs`: **38 sites across 10 files → 37 across 9.**

## Verification

`RUSTFLAGS="-D warnings" just check-gated`, `cargo nextest run --workspace` (12,920 passed), `cargo run -p xtask -- check-all` (66/66), `cargo fmt --all -- --check`, plus `actionlint` via the pre-commit hook for the workflow edit.


auser:
Rebased onto the rebased #3106, which now targets `main` directly (#3100 closed as superseded by #3072). Replayed cleanly with no conflicts.

Re-verified: `check-no-virtio-fs` clean at **37 sites across 9 files**, `cargo nextest run --workspace` 12,950 passed, `check-all` 67/67, `RUSTFLAGS="-D warnings" just check-gated` clean.

Merge order: #3106 first, then this retargets to `main`.

auser:
The `Test eBPF telemetry load/attach` failure here is a flaky test, not a regression from this PR.

Re-running the identical job on the identical commit came back green with nothing else changed (`gh run rerun 33653066182 --job 100324800583`).

`supervisor::audit_mirror::tests::mirror_event_carries_non_sensitive_envelope_fields` captured 0 events instead of 1. `mvm-hostd` has 121 `tracing::subscriber::with_default` call sites, and while the `Dispatch` is thread-local, the max-level filter `tracing::info!` consults first is a process-global atomic that is restored on scope exit. This lane runs `cargo test -p mvm-hostd --lib` — threads in one process — so one test's scope exit can drop another's event. `cargo nextest`, the named gate in CLAUDE.md, gives each test its own process and cannot reproduce it.

Written up in `.agent-memory/notes/mvm-hostd-lib-tests-race-on-the-global-tracing-filter.md` (landing on #3117) so the next person doesn't bisect their PR against it. Nothing to change here — this PR is otherwise green.

auser:
**Release coordination — this is wanted in 0.18.0.**

Following up on my earlier comment about the flake: I'm driving the 0.18 release ([#3117](https://github.com/tinylabscom/mvm/pull/3117)), and Ari asked specifically for virtio-fs to be gone in this release. So this is in scope, not deferred.

**Status:** the only red check was `Test eBPF telemetry load/attach`, which I re-ran on the identical commit and it came back green — a pre-existing parallel-test race, written up in `.agent-memory/notes/mvm-hostd-lib-tests-race-on-the-global-tracing-filter.md` (landing on #3117). Nothing to fix here.

**Stacked on #3106**, which is `CLEAN` and green — retarget once that lands.

**Verified independently, so you don't have to re-derive it:** no claim witness dies. `dir_share_two_part_defaults_ro`, `relay_config_maps_dir_shares_with_dax_and_read_only`, `enforce_admitted_shares_refuses_unadmitted_or_mismatched` (MVM-SEC-01) and `virtiofs_mount_flags_keep_workspace_read_only` (MVM-SEC-02) all live outside what this deletes. Your PR body's claim that the FUSE parser is cited nowhere is correct and I confirmed the broader set too.

**Also confirmed your central premise.** Every `VirtioFsShare { … }` construction in the tree is inside `mod tests`, and every production `shares:` initializer is `Vec::new()`/`vec![]` — so `kernel_boot.rs` builds its device list from an always-empty vector and the FuseServer is unreachable on every path. Deleting rather than fuzzing is the right call.

**Sequencing:** the release gate now requires a recorded macOS documented-surface run whose digest covers `crates/`, so please land this before the final evidence run rather than after — otherwise the two-hour suite gets repeated. Ping me when it's queued.

**What's left after this stack**, for whoever picks up Stage C (from `specs/plans/2026-08-31-remove-virtio-fs.md`): the libkrun persistent builder still serves `work`/`mvm-bins`/`job`/`out` as shares, and the guest's install arm writes to `/job/<job_id>/out` which the disk transport doesn't collect — that one needs a guest change and an image rebuild. Both are the trusted builder tier, not the untrusted workload path, so the threat model is materially weaker there.

PR #3129, opened 2026-09-02
feat(contract): make SDK sidecar delivery explicit

### LINKED ISSUE #3068: The SDK sidecar is a convenience, not a requirement: a guest can speak the broker protocol directly
`libmvm_host_services.so` is a native ELF shared object mvm mounts at `/mvm/sdk` and the workload `dlopen`s. Because it is native, it has to match the guest's libc — and the guest's libc is whatever OCI image the user named. That single fact is the origin of a large amount of machinery:

- two nix derivations (`sdk-cdylib-glibc` / `sdk-cdylib-musl`) and two sidecar images
- a libc-keyed cache (`sdk-sidecar/<version>/<arch>/<libc>/`)
- per-image variant selection (#3044, and #3060 for arbitrary `--image`)
- an ELF `DT_NEEDED` reader so a mislabelled artifact is caught (#3060)
- two published release assets per architecture, each signed
- a fail-closed admission gate that refuses a mismatch

## What the cdylib actually does

From its own module doc:

> The contract is deliberately tiny and language-agnostic: JSON in, JSON out, one paired free function… The veneer holds no key and is not a security boundary… Every gate — the `ExecutionPlan.services` binding, the category-forcing and size/rate caps on `host.audit.v1`, the correlation-id assignment — lives host-side. This layer only translates between the C ABI and the typed Rust transport.

Concretely: marshal JSON → `connect_host_vsock(BROKER_PORT)` → write a 4-byte big-endian length-prefixed frame → read the reply → hand back JSON.

## A guest can do that itself

Verified live on Apple Silicon, in a booted `python:3.12-alpine` guest, with stock CPython and no native code:

```
KV-OK-NO-SIDECAR     exit 0
```

The fixture is ~60 lines of stdlib Python: `socket.AF_VSOCK` to `(HOST_CID=2, BROKER_PORT=5300)`, the same framing, unwrapping `{"kind":"ok","payload":…}`. It asserts the same three things the real `kv_roundtrip.py` does — absent before write, `replaced: false`, value round-trips. Script at the bottom.

To be precise about what that run does and does not show: `host.kv.v1` is in `SDK_HOST_SERVICES`, so the sidecar *was* still attached — it simply went unused. The claim is that the cdylib is not needed to reach the service, not that admission currently tolerates its absence. It does not: `sdk_sidecar_required_for` keys on the service being SDK-served, and both the resolver and `enforce_sdk_sidecar_attachment` refuse without an attachment.

**This is not a new capability and adds no attack surface.** `broker_client`'s own doc states the client is advisory and that "a compromised workload can bypass it and write raw frames to the port," because every gate is host-side. The spike is that documented property exercised deliberately. Traffic goes over virtio-vsock, which is the mandated chokepoint — there is no NIC, no IP, no TAP/TUN involved. The host still refused an unbound service with `not_bound`.

## Proposal

Make `sdk_sidecar_required_for` conditional on the SDK actually needing native delivery, rather than on the service being SDK-served. An SDK that speaks the protocol natively binds the service and attaches nothing; the libc machinery stays correct and stops being on the critical path for those workloads.

## What this does not solve

- **It is not uniform across languages.** Python has `AF_VSOCK` in the stdlib (Linux); Go has it via `golang.org/x/sys`. Node has no stdlib AF_VSOCK, so the TypeScript SDK would still need the sidecar or a helper process. This retires the sidecar for some SDKs, not all.
- **Framing and the error vocabulary get reimplemented per language.** That is exactly what the cdylib centralizes — one implementation, so a new SDK is a shim. It was ~60 lines in Python, but it is N copies to keep in step with the wire types.

So this is a trade, not a free win. The reason to consider it is the recurring cost: the sidecar's *delivery* has consumed a long run of issues (#2924/#2926, #2949/#2950, #2941/#2954, #2951/#2952, and #2969 with five separate blockers before it reached the code it was about), all for a component that enforces nothing and holds no secret.

## The spike

```python
import json
import socket
import struct
import sys

HOST_CID = 2        # mvm_agentd::vsock::connection::HOST_CID
BROKER_PORT = 5300  # mvm_agentd::vsock::BROKER_PORT
SERVICE = "host.kv.v1"


def call(verb: str, payload: dict, timeout_secs: int = 30) -> dict:
    """One ServiceCall out, one ServiceResponse back, on its own connection.

    One connection per call is the protocol, not a simplification: the host
    reassigns correlation_id at ingress, so replies cannot be matched to
    requests on a shared connection.
    """
    if not hasattr(socket, "AF_VSOCK"):
        raise RuntimeError("this Python was built without AF_VSOCK support")
    s = socket.socket(socket.AF_VSOCK, socket.SOCK_STREAM)
    s.settimeout(timeout_secs)
    try:
        s.connect((HOST_CID, BROKER_PORT))
        frame = json.dumps({
            "service": SERVICE,
            "verb": verb,
            "correlation_id": "spike-0",  # the host rewrites this at ingress
            "payload": payload,
        }).encode()
        s.sendall(struct.pack(">I", len(frame)) + frame)
        (length,) = struct.unpack(">I", recv_exact(s, 4))
        response = json.loads(recv_exact(s, length))
    finally:
        s.close()

    if response.get("kind") != "ok":
        raise RuntimeError(
            f"{verb} refused: code={response.get('code')} "
            f"message={response.get('message')}"
        )
    return response["payload"]


def recv_exact(s: socket.socket, n: int) -> bytes:
    buf = b""
    while len(buf) < n:
        chunk = s.recv(n - len(buf))
        if not chunk:
            raise RuntimeError(f"broker closed after {len(buf)} of {n} bytes")
        buf += chunk
    return buf


def main() -> int:
    key, value = "vsock-spike-witness", [1, 2, 3]

    first = call("get", {"key": key})
    assert first.get("value") is None, f"expected absent, got {first}"

    put = call("put", {"key": key, "value": value})
    assert put["replaced"] is False, f"first write must not replace: {put}"

    got = call("get", {"key": key})
    assert got.get("value") == value, f"round trip lost the value: {got}"

    print("KV-OK-NO-SIDECAR")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

Run with:

```
mvmctl run --image python:3.12-alpine --host-service host.kv.v1 \
  --mount <dir>:/work/spike:ro --timeout 300 \
  -- python /work/spike/kv_roundtrip_vsock.py
```

auser:
## Summary

- add a signed `sdk_uses_sidecar` execution-plan field
- let direct broker-protocol workloads bind SDK host services without a native sidecar
- preserve the legacy compatibility default (`true`) when older serialized plans omit the field
- fail closed when a direct-protocol plan carries a stray sidecar
- keep the existing macOS documented-surface release gate intact

## Validation

- `cargo test --workspace --quiet`
- `cargo clippy --workspace -- -D warnings`
- `cargo check --workspace`
- targeted contract, admission, and workflow regression tests

Fixes #3068.


auser:
**The `.github/` half of this repeats a change that was closed twice today — please drop it and keep the rest.**

This removes `macos_blocks_on_unusable_host`, the `e2e-docs-macos-host-check` job, and the macOS lane from `e2e-docs.yml`, plus the caller arguments in `ci-full.yml` and `release.yml`. That is the same workflow change as [#3122](https://github.com/tinylabscom/mvm/pull/3122) and [#3123](https://github.com/tinylabscom/mvm/pull/3123), both closed earlier today with the reasoning below. Their heads are preserved as `backup/pr-3122` and `backup/pr-3123` on origin if you want the diff.

**Why it should not land:** it buys a green release by deleting the requirement rather than satisfying it. After it, a tag cuts with no macOS evidence of any kind and nothing says so — the same silent-gate shape that let `machine run -it` ship broken on every OCI image.

**The approach Ari picked instead is in [#3117](https://github.com/tinylabscom/mvm/pull/3117):** the host check stays, and when no runner can boot a guest the release falls back to a *committed evidence record* that `xtask check-release-evidence` verifies against a content digest of the tree being tagged. It retires itself the moment a self-hosted Apple Silicon runner appears (#3011) — the job's condition is `supported != 'true'`, so pointing `runs-on` at real hardware is the whole change.

#3117 also pins this with tests, so this PR will go red against main once #3117 lands rather than landing quietly:
- `releases_still_block_on_a_macos_host_that_cannot_boot_a_guest`
- `an_unusable_macos_host_falls_back_to_a_checked_evidence_record`

**The `sdk_uses_sidecar` half is wanted** — it is byte-identical to the code in #3118/#3122/#3123/#3124 (verified: same MD5 of each PR's crate diff against its own merge base). Dropping the three workflow files from this branch should leave it landable on its own.

Worth knowing that #3117 is close: it is merged up to main as of `6837a9424f` and only waiting on CI plus a fix for #3130.

PR #3145, opened 2026-09-02
feat(evidence): boot beacon, deployments inventory, verity-sealed provenance mark + in-toto sidecar

auser:
Phase 1 slice of `specs/plans/2026-09-02-supply-chain-evidence-carryover.md` — closes the supply-chain evidence gaps against the competitive baseline in that plan: WS1.1 boot beacon, WS1.4 deployments inventory, WS2.1 verity-sealed provenance mark, WS3.1 in-toto/DSSE provenance sidecar.

## `host.beacon.v1` — boot liveness on the chain

The guest agent reports its own boot once over the existing vsock broker; the host handler stamps the supervisor-authoritative identity (`workload_id` / `tenant_id` / `session_id` / `correlation_id`) from `ServiceCallCtx` and appends a chain-signed `lifecycle.beacon_reported` entry through the audit-signer.

- **Host-derived identity by construction** — the guest payload cannot express identity fields (`deny_unknown_fields`), so unlike a `workload_audit` entry the binding is not guest-asserted; a compromised guest can lie about its version string but cannot forge which workload the entry proves alive.
- **Default-on** wherever an audit signer exists (not gated on `services_bindings`, since the emitter is the platform agent, not workload code); plan-policy suppression seam (`SubprocessConfig` flag) is the designed follow-up, deferred to the environment-model work.
- **Fail-open**: guest-side bounded retries (3 × 2 s) cover host-broker boot races; every failure mode logs and continues — a missing beacon never delays a workload boot.
- **Bounded**: 1 token/s per-workload rate limit caps chain bloat under a hostile guest; 4 KiB payload cap.

Complements `plan.launched`: that entry proves the supervisor started a backend; this one proves the admitted workload's agent actually came alive inside it — the raw signal behind "is this deployment actually running?".

## `mvmctl deployments ls [--workload] [--json]`

The first read path for the local-first deploy store (`<mvm_home>/deployments/<ir-hash>/deploy.json`). Unreadable records (missing, malformed, future schema) surface as named skips instead of failing the listing or silently disappearing. `--env` filtering waits for the environment model in WS1.3 — records today pin only the kernel digest. Row shape is designed to match the future control-plane response so studio renders one type.

## Verity-sealed provenance mark (WS2.1)

Every **sealed** OCI rematerialize now writes a provenance mark into the rootfs at `/mvm/provenance.json` + detached Ed25519 signature **before** the dm-verity hash is computed — so a tampered mark breaks the verified boot chain (kernel-enforced), not just a signature check. The mark is canonical JSON (JCS) covering input reference, resolved digest, builder identity, timestamps, and tool version; the SBOM reference is reserved as `None` until SBOM emission lands (WS6). Signed with the host signer (`~/.mvm/keys/host-signer.ed25519`, the same key as `artifact pack`). Gated on the existing `sealed` request flag — sealed images always carry the mark; a sealed request with no evidence logs a warning instead of failing the seal.

## in-toto / SLSA v1 provenance sidecar (WS3.1)

The seal also emits an in-toto Statement (SLSA v1.0 provenance predicate) wrapped in a DSSE envelope (PAE signing, builder Ed25519 key) as `rootfs.intoto.json` beside the sealed image. The statement subject binds the sha256 of the sealed rootfs's exact bytes, so the sidecar is a standards-shaped interchange view of the same facts as the in-artifact mark. Keyless Sigstore and audit-chain digest recording are deferred (WS4 covers publication).

## Validation

- `cargo test --workspace` green
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `just check-gated` (Linux cross-compile of all targets + BDD-feature conformance check) clean
- `xtask check-all` clean (plan names, spec-ref lint, sprint append, CLI-help-vs-docs, honesty, …)
- 32 new tests: WS1.1 payload roundtrips + spoof refusal, client envelope/socket-pair, handler stamping/rate-limit/signer-error mapping, retry semantics; WS1.4 store enumeration/skips, CLI end-to-end with hermetic `MVM_HOME`; WS2.1 mark roundtrip/tamper-rejection/JCS determinism, seal-gate behavior (sealed+evidence → verifiable mark, sealed+none → warn, unsealed → no mark), CLI evidence threading (prod vs dev rematerialize); WS3.1 PAE encoding, statement shape, DSSE wrong-key rejection, sidecar read/verify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2983,6 +2983,7 @@ dependencies = [
  "nix 0.29.0",
  "rand 0.10.2",
  "serde",
+ "serde_jcs",
  "serde_json",
  "sha2 0.11.0",
  "tar",
--- a/crates/mvm-agentd/fuzz/Cargo.lock
+++ b/crates/mvm-agentd/fuzz/Cargo.lock
@@ -1072,6 +1072,7 @@ name = "mvm-build"
 version = "0.18.0"
 dependencies = [
  "anyhow",
+ "base64",
  "chrono",
  "ed25519-dalek",
  "flate2",
@@ -1087,6 +1088,7 @@ dependencies = [
  "mvm-vmm",
  "nix",
  "serde",
+ "serde_jcs",
  "serde_json",
  "sha2",
  "tar",
--- a/crates/mvm-agentd/src/bin/mvm-guest-agent.rs
+++ b/crates/mvm-agentd/src/bin/mvm-guest-agent.rs
@@ -727,6 +727,19 @@ fn main() {
     let guest_signing_key = Arc::new(SigningKey::from_bytes(&guest_seed));
     boot_state.mark_vsock_bound();
 
+    // Boot beacon: report agent liveness to the host once, on a
+    // background thread with bounded retries. Fail-open by contract —
+    // a missing host broker (dev images without an audit signer, a
+    // boot-time race the retries don't cover) must never delay or
+    // fail the workload's boot.
+    let beacon_boot_unix_ms = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| u64::try_from(d.as_millis()).unwrap_or(u64::MAX))
+        .unwrap_or(0);
+    std::thread::spawn(move || {
+        mvm_agentd::host_beacon::report_boot_beacon(beacon_boot_unix_ms);
+    });
+
     // On legacy per-rootfs boots the agent is not PID 1 and the workload
     // root is already in place, so validate the entrypoint and start the warm
     // pool now. On the universal initramfs path validation is deferred until
--- a/crates/mvm-agentd/src/host_beacon.rs
+++ b/crates/mvm-agentd/src/host_beacon.rs
@@ -1,0 +1,325 @@
+//! In-guest `host.beacon.v1` typed method — the boot beacon.
+//!
+//! The platform's own guest agent calls `report` exactly once at boot,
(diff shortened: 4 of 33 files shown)


auser:
**The `Invariant` failure is the detached fuzz lockfiles, not your code.**

Failing step: `cargo-fuzz crates still compile`.

This PR changes `crates/mvm-build/Cargo.toml` and `Cargo.lock`. The fuzz crates are **outside the workspace** and carry their own lockfiles, so a workspace dependency change leaves them stale and they stop resolving. There are six:

```
crates/mvm-contract/fuzz/Cargo.lock
crates/mvm-agentd/fuzz/Cargo.lock
crates/mvm-http/fuzz/Cargo.lock
crates/mvm-fs/fuzz-ext4/Cargo.lock
crates/mvm-fs/fuzz-oci/Cargo.lock
crates/mvm-runtime/fuzz-backend/Cargo.lock
```

Refresh whichever ones resolve the crates you touched (`cargo update -w` inside each affected fuzz directory) and commit the lockfiles.

Worth knowing about this one specifically: **`check-all` cannot see it.** The fuzz crates are not workspace members, so a full local `just ci` passes while this lane is red — it is one of the few failures you cannot reproduce with the normal gate. It also fails the *whole merge-queue batch* rather than just this PR, so it is worth clearing before enqueueing.

**Release context, no pressure implied:** we are sequencing the 0.18.0 cut, and the plan is now to reach zero open PRs, then run a release dry-run, then the documented-surface evidence run, then tag. So this one is on the path rather than blocked behind it — happy to help if the lockfile refresh turns up anything unexpected.

PR #3151, opened 2026-09-03
feat(volume): snapshot a --host directory into an image instead of refusing it

auser:
feat(volume): snapshot a --host directory into an image instead of refusing it

--host registered a live host-directory share, which no workload backend can
serve. The two launch paths then disagreed about it: machine start refused at
boot ("a live host-directory share can't be expressed", measured on both
Firecracker and HVF) while machine run booted without the mount and only
warned. The flag was documented, with two examples and a rationale, and had
never worked on any path.

It now snapshots the directory into an ext4 image at registration and
registers that - the same treatment --mount HOST:/GUEST already gives a
transient run, using the same pure-Rust writer. So a persistent machine gets
the host-directory story it did not have, rather than a better error message
for a feature that cannot work.

Three small pieces make it fit rather than bolt on. resolve_mount_entry
honours the kind the registration recorded instead of hardcoding Directory for
every ad-hoc entry - the hardcode was accurate when a directory share was the
only ad-hoc shape and coerced a perfectly good image back into a share once it
was not. Entries written before kind existed still decode as Directory through
its serde default, so they resolve exactly as before.
prepare_ad_hoc_image_attachment is additive rather than a change to
prepare_ad_hoc_host_attachment, whose contract other consumers of mvm-client
depend on. And require_host_encryption is exposed so the source directory
still satisfies the at-rest policy it did when attached directly - moving
materialization to the caller must not drop that check silently.

The docs claimed the directory was "the source of truth", a live semantic that
died with virtio-fs. They now say snapshot-at-registration and that
re-registering refreshes it. That correction was missing from the previous
attempt at this: machine volume mount is tier = "parse" in tiers.toml, so the
doc gate checks the command parses and cannot see that its behaviour changed.

Verified live on macOS 26: registration writes
volumes/host-snapshots/<vm>/<volume>.ext4, machine start boots where it
previously refused, and the guest reads the seeded marker back. BDD 247
passed, workspace 12,987 passed, 68 gates, check-gated and clippy clean.

### DIFF
--- a/crates/mvm-cli/src/commands/vm/volume.rs
+++ b/crates/mvm-cli/src/commands/vm/volume.rs
@@ -368,6 +368,21 @@ fn host_snapshot_image(vm_name: &str, volume_name: &str) -> std::path::PathBuf {
         .join(format!("{volume_name}.ext4"))
 }
 
+fn ensure_private_snapshot_parent(output: &Path) -> Result<()> {
+    let parent = output
+        .parent()
+        .with_context(|| format!("snapshot path {} has no parent", output.display()))?;
+    std::fs::create_dir_all(parent)
+        .with_context(|| format!("creating snapshot dir {}", parent.display()))?;
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt as _;
+        std::fs::set_permissions(parent, std::fs::Permissions::from_mode(0o700))
+            .with_context(|| format!("restricting snapshot dir {}", parent.display()))?;
+    }
+    Ok(())
+}
+
 /// Materialize `host_dir` into an ext4 image for registration.
 ///
 /// The same pure-Rust writer `--mount` uses — no `mkfs`, no subprocess, and it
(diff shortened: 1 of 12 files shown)


auser:
**The BDD failure is a host-dependent assertion, not a logic bug — it will pass on your machine and fail on every Linux runner.**

`features/suites/s26_volumes/volume_lifecycle.feature:137`:

```
Error: LocalBackend volume mounts require the mounted host directory to live on
encrypted backing storage. /tmp/.tmpFi…
  left: Some(1)
 right: Some(0)
```

The scenario mounts a `tempfile` directory under `/tmp` and asserts exit 0. On macOS with FileVault the whole volume is encrypted, so the probe admits it; on a GitHub Linux runner `/tmp` is plain `tmpfs`/ext4 and the probe correctly refuses. Same shape as the wall-clock refusal I gated earlier today: an assertion that is true on one tier and false on another, with nothing expressing which tier it needs.

`Test Linux and conformance` fails for the same reason; `Test` is the aggregate reporting both.

**There is machinery for exactly this.** `crates/mvm-conformance/src/lib.rs` carries operator-declared capability tags — `@dir_share`, `@snapshot`, and as of today `@unenforceable_wall_clock` — each gating on a `RuntimeCaps` field set from an env var rather than probed, so the harness never re-derives host or backend selection. Adding an `encrypted_backing` capability alongside them, declared by the lanes whose host has it, would let this scenario skip cleanly on Linux instead of failing.

The alternative — pointing the fixture at a directory that is encrypted on every runner — is probably not available, which is likely why it landed on `/tmp`.

**One thing worth deciding explicitly while you are here:** the PR title changed from *"refuse a host-directory registration when it is registered"* to *"snapshot a --host directory into an image instead of refusing it"*. That second form overlaps the `--mount` materialization path directly — `crates/mvm-cli/src/exec/mounts.rs`, which now streams the tree into an ext4 image rather than buffering it (landed in #3117's branch, currently in the merge queue). If both paths are materializing directories, they should share `materialize_mount_image` rather than growing a second implementation; the CLAUDE.md "reuse first" rule calls duplicated logic this repo's most common bug source, and this is the shape it takes.

**Release context:** we are driving to a 0.18.0 tag once open material PRs reach zero, so this one is on the path. Happy to help wire the capability tag if that is useful — I added the last one and the pattern is fresh.

PR #3156, opened 2026-09-03

auser:
This PR updates the content address fixtures in the replay vector tests to account for the sdk_uses_sidecar field addition in main.

The sdk_uses_sidecar field was added in #3129 with a different serde attribute
(#[serde(default = "default_sdk_uses_sidecar")] with a const fn). This PR
changes it to #[serde(default)] to use Rust's default value.

This PR also includes the e2e-docs.yml workflow fixes that ensure the macOS e2e
tests use the correct runner with HVF support via a host check job and
conditional execution.

Generated with Qwen Code


auser:
Closing after a commit-by-commit and final-tree audit against current main. The content-address fixture update is already present from #3129, and this branch has no remaining fixture delta. The macOS host probe plus the extended-CI false and release true caller policies are present on main. The AGENTS main-checkout invariant landed through #3160. Commit 9f8bbe11 is intentionally not preserved: bare #[serde(default)] makes an absent bool false and breaks signed-plan compatibility; #3158 landed as 3ec161139f with regression tests preserving true and a dormant-control guard. The merge commit adds no unique work. SECURITY.md is identical between this branch and main at Git blob 004ad809a740303e8e007d2b6849e1961026efe2. The remaining branch result consists of stale formatting/comment removal and the unsafe default change, so there is no valid unpreserved work to merge.

auser:
Commit accounting after retirement:

- a8c444e32b removed release-workflow behavior and did not preserve the documented true default; that workflow surface is now correctly covered by merged PR #3117.
- e1e1baadd2 restored the documentation E2E workflow that the prior commit had removed; the net release-gate behavior is covered by merged PR #3117.
- 9f8bbe11c8 set sdk_uses_sidecar to false, which contradicts the documented contract. Merged PR #3158 implements the required true default and pins it with the contract test.

The intended release-gate and sidecar-contract work is implemented in #3117 and #3158, with the incorrect default excluded.

PR #3161, opened 2026-09-03
fix(contract): pin the sdk_uses_sidecar default, and declare it dormant

auser:
Fix CI failures for sdk_uses_sidecar by pinning the default and adding tests.

Generated with Qwen Code

### DIFF
--- a/crates/mvm-contract/src/plan/execution_plan.rs
+++ b/crates/mvm-contract/src/plan/execution_plan.rs
@@ -432,6 +432,54 @@ mod tests {
 
     use super::*;
 
+    /// A plan written before `sdk_uses_sidecar` existed must deserialize as
+    /// *requiring* the sidecar, not forbidding one.
+    ///
+    /// The field is a two-sided assertion: admission reads `true` as "the
+    /// sidecar must be attached" and `false` as "it must not be"
+    /// (`enforce_sdk_sidecar_attachment`). So the default is not a tie-break
+    /// between equivalent options — one of them refuses to admit every plan
+    /// that predates the field, and it is the one a bare `#[serde(default)]`
+    /// would pick, because `bool::default()` is `false`.
+    ///
+    /// Four PRs proposed exactly that substitution. This test is what makes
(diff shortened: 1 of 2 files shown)


auser:
Closing as a duplicate of **[#3158](https://github.com/tinylabscom/mvm/pull/3158)**, which is the same work opened at 18:20 and already has CI running. Head preserved as `backup/pr-3161`; reopens with one click.

The two diffs are identical apart from one sentence in `xtask/dormant-controls.toml`:

```
#3158  Wiring it requires a launch-time way to declare that a workload speaks
       the broker protocol directly. This entry fails the moment that path lands…

#3161  Wiring it — a launch-time way to declare a workload speaks the broker
       protocol directly — is issue #3068's work. This entry fails the moment…
```

#3158 has the newer wording. The `issue #3068` reference was edited out deliberately — this project keeps process citations out of source, and I had put one in the same file earlier today and had it corrected for me. So this branch looks like a copy taken *before* that fix.

No work lost either way: the test, the assertion message and the dormancy declaration are byte-identical between them.

Worth flagging, since it is the second time today: measuring a branch with `git diff origin/main` conflates "what I changed" with "what main changed under me". `git diff $(git merge-base origin/main HEAD) HEAD` is the honest form. If this branch was created by copying a worktree rather than branching from it, that is also worth knowing — it is how a pre-correction state gets resurrected.

auser:
Superseded by #3158, which merged as 3ec161139f after both pull-request CI and merge-group CI passed. The same sidecar-default regression tests and dormant-control declaration are now on main; #3158 also removes the process-artifact reference that caused this PRs invariant failure. The remote branch remains available, but there is no unique valid change left to merge.

auser:
Commit accounting after retirement:

- e0efb1bb9d contains the same sidecar-default correction as the replacement, but its source comment carried a prohibited process-artifact reference and failed the repository invariant gate.
- Merged PR #3158 contains the equivalent implementation and regression test with the comment corrected to state the invariant directly.

The requested contract change is fully landed in #3158.

PR #3164, opened 2026-09-03
Refresh host directory snapshots at machine start

auser:
## Summary

- refresh registered `--host` directory snapshots immediately before machine launch when their emitted filesystem content changes
- use one verified content-addressed ext4 image cache for persistent `--host` volumes and cacheable transient `--mount` inputs, without a source-size ceiling
- refuse launch when a registered source directory is missing instead of silently serving the last snapshot
- attach immutable cache images directly for read-only mounts and create private reflink/copies for writable registrations
- populate the existing `mount_fingerprint`, `mount_cache_lookup`, and materialization timing spans

## Change detection and cache safety

The source fingerprint covers the filesystem semantics written into the ext4 image: relative paths, file bytes, permission modes, symlink targets, and guest-visible xattrs. It deliberately excludes mtimes, uid, and gid because the materializer normalizes those values.

Files are streamed in bounded blocks at every source size. Each file digest is verified between the source walk and image emission, so same-size mutation is rejected. Cache entries are atomically published, private, read-only, and verified against a strict manifest plus the image digest before use. Writable and encrypted destinations are materialized into private copies, with destination encryption verified separately.

## Verification

- full workspace unit/integration suite green serially; the isolated `mvm-build` doctest also passed after one workspace rustdoc dependency-resolution transient
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `just check-gated`
- all 68 repository policy gates (the SDK cdylib gate passed standalone after overriding the host sccache wrapper denied by the sandbox)
- `just bdd` — 250 scenarios: 249 passed, 1 capability skip; 956 passing steps
- live Firecracker validation: changed source bytes while restoring the original mtime, stopped and restarted the machine, and observed `dir-volume-refreshed` inside the guest


auser:
Merge-queue live BDD exposed a deterministic fixture precondition: the hosted runner places its isolated MVM_HOME on an unencrypted /tmp filesystem, so the new cache correctly failed closed before boot. Commit 52f29022f2 makes the live README scenario explicitly declare encrypted backing via the same hermetic findmnt/lsblk probes already used by the host-directory volume scenarios; production probing is unchanged. Local verification on the updated head: cargo fmt --check; BDD 251 scenarios (250 passed, 1 capability skip), 959 passing steps; BDD-target Clippy with -D warnings. Auto-merge is re-enabled for a fresh exact-head run.

Issue #3155, opened 2026-09-03
Decide whether sdk_uses_sidecar should default to true or false


auser:
Four PRs have now tried to change how `ExecutionPlan::sdk_uses_sidecar` is defaulted, and none has landed. Filing the underlying question so it gets decided on its merits rather than re-litigated inside a fixtures PR.

## Where it stands on main

```rust
/// `true` (default) preserves existing behavior: workloads bound to SDK
/// services receive the glibc sidecar. Set to `false` when the workload
/// can speak the broker protocol natively (e.g., Python with AF_VSOCK,
/// Go with golang.org/x/sys/unix).
///
/// Always serialized so a plan that doesn't set it still states the
/// default in the signed bytes.
#[serde(default = "default_sdk_uses_sidecar")]
pub sdk_uses_sidecar: bool,

const fn default_sdk_uses_sidecar() -> bool { true }
```

Landed via #3129. Main is green.

## The proposed change, and why it was held

[#3152](https://github.com/tinylabscom/mvm/pull/3152) proposed replacing that with a bare `#[serde(default)]`. On a `bool` that resolves to `Default::default()` — **`false`**, not `true`.

That is not a cosmetic default. `enforce_sdk_sidecar_attachment` (`crates/mvm-hostd/src/plan_admission.rs:1407`) reads `false` as a prohibition:

```rust
if !plan.sdk_uses_sidecar {
    // ...
    anyhow::bail!(
        "refusing to attach '{}' at {SDK_SIDECAR_GUEST_PATH}: the signed ExecutionPlan \
         sets sdk_uses_sidecar=false, so this workload must not carry the SDK sidecar", ...
    );
}
```

So any signed plan written before the field existed would go from *attach the sidecar* to *refuse to attach it*, and every workload binding an SDK host service would stop being admitted.

## The question

1. **Should the default stay `true`?** It preserves behaviour for plans that predate the field, which is what the current doc comment commits to.
2. **Or should the sidecar become opt-in (`false`)?** Defensible — it is the smaller default surface, and #3068 argues the sidecar is a convenience rather than a requirement. But it is a behaviour change to a signed, security-bearing type, and it needs a migration story for plans already in the wild.

If (2), the change should say so in its own PR with that reasoning, and should probably come with a note in the security posture ADR about what an unset field now means.

## A loose end worth checking

[...]

PR #3165, opened 2026-09-03
fix(contract): pin the sdk_uses_sidecar default, and declare it dormant

auser:
Fix CI failures for sdk_uses_sidecar by pinning the default and adding tests. This applies the fix from PR #3155 to fix the wasm32-wasip1 CI failures.

Generated with Qwen Code

### DIFF
--- a/crates/mvm-contract/src/plan/execution_plan.rs
+++ b/crates/mvm-contract/src/plan/execution_plan.rs
@@ -432,6 +432,54 @@ mod tests {
 
     use super::*;
 
+    /// A plan written before `sdk_uses_sidecar` existed must deserialize as
+    /// *requiring* the sidecar, not forbidding one.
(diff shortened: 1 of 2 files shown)


auser:
Commit accounting before retirement:

- 56be93ac2458c5ec00404c057f465ff contains the same two regression tests and dormant-control declaration as merged PR #3158.
- Its production-source test comment refers to prior PR activity, which violates the repository rule against process-artifact references in code comments. PR #3158 landed the equivalent behavior with that comment rewritten as a direct invariant.

The sidecar default, both regression tests, and dormant-control documentation are already on main through #3158; this branch adds no unlanded behavior.

auser:
Retiring this duplicate after verifying that its complete functional change landed in corrected form through #3158.

PR #3169, opened 2026-09-03
feat(release): attest build provenance for the release tarballs

auser:
Implements WS-B of `specs/plans/2026-09-03-release-artifact-authenticity-and-provenance.md`.

> **Stacked on #3162.** Base is `feat/claim-release-artifact-authenticity`, because this extends claim 20's witnesses and that claim doesn't exist on `main` yet. Retarget to `main` once #3162 merges.

## What changed

`publish-npm.yml` already publishes with `--provenance`, so the npm package carried build provenance while the binaries most people install did not. The release job now attests the tarballs with `actions/attest-build-provenance@v4`, placed after signing and before `gh release create`, with `attestations: write` added to the workflow permissions.

Subject set is **the tarballs only**, matching the directly-signed set. Kernels, root filesystems and metadata carry no per-file signature and are anchored by digest inside the signed checksum manifests — attesting them individually here would assert a per-file guarantee the pipeline does not make.

## Stated honestly, in four places

This is **SLSA Build L2**, and the ADR, `CLAUDE.md`, `model/claims.toml` and the workflow comment all say so. The attestation is produced by the same workflow that produces the artifact, so a compromised release job forges both. What provenance adds over the cosign signature is the commit and build inputs in a machine-readable predicate — it is **not** evidence the build was uncorrupted. `ci:reproducibility` under claim 7 remains the control for that, because an independent rebuild landing on the same bytes needs no trust in this builder at all.

ADR-001 and `CLAUDE.md` previously stated outright that no provenance was emitted. This makes that false, so both were rewritten rather than left to rot.

## The witness is the CI anchor, not the test

The structural test lives in root-level `tests/release_assets.rs`, which `resolve_fn_needles` still cannot cite — the gate defect WS-A documented. Rather than widen the gate, the attesting step carries a parenthesised token in its `name:` (`release-provenance`), which `ci_anchors` resolves. So claim 20 gains `ci:release-provenance`, the test stays an uncited regression guard, and the WS-A scoping decision is untouched.

## The test was falsified before landing

`the_release_attests_build_provenance_for_the_signed_tarballs` pins the subject set to the signed tarballs, asserts the permission, and asserts ordering against `gh release create`. Both assertions were verified to actually bite:

- mutating `subject-path` to `artifacts/*.tgz` → red, `provenance must cover the same tarballs the signing loop signs directly`
- deleting `attestations: write` → red, `actions/attest-build-provenance records nothing without it`

## Verification

- `cargo run -p xtask -- check-all` — 68/68 clean
- `just bdd` — 248 scenarios, 247 passed, 1 skipped, 0 failed
- `cargo nextest run -p mvm-conformance` — 97/97
- `cargo nextest run --test release_assets` — 31/31
- `actionlint` via the pre-commit hook

## What this cannot verify

The attestation step itself only runs on a real `v*` tag. Nothing here proves it works end-to-end — the structural test proves it is present, correctly scoped and correctly ordered, and `actionlint` proves the YAML is valid. First real exercise is the next release; if `attest-build-provenance` misbehaves it surfaces there, after signing and before publish, so it fails the release rather than shipping a bad one.


auser:
Parent PR #3162 is merged as 6718abc647a47bbb90677aaa3171d29be49a800e. This branch has been rebased onto that exact main commit and the PR base is now main; the focused release-assets suite passes 31/31 on the rebased tree.

PR #3170, opened 2026-09-04
fix: execute every README example, and fix the three defects that found

auser:
Closes the gap between "the README documents this" and "something runs it," and fixes the three defects that turned up while doing so.

## The README register

Exemptions went **8 → 1**, and 34 of 35 examples are now executed.

| example | before | now |
| --- | --- | --- |
| `bootstrap` | exempt | `@live` scenario |
| `build kernel build --which workload --source download` | exempt | `@live` scenario |
| `generate template python ./my-python-app` | exempt | hermetic witness, run verbatim |
| `run --peer db.mvm.peer:5432=127.0.0.1:34567` | exempt | `@live` scenario, real guest dial |

Two of those exemptions asserted coverage that did not exist. `--source compile`
claimed the "builder-VM image lane in Extended CI" — that lane builds a
different derivation, and both kernel lanes run `scripts/build-kernel-artifacts.sh`,
which calls `nix build` directly and never goes through `mvmctl`. `deps install`
claimed the app-deps-audit lane, which hand-seals its fixture and never runs the
command. The one remaining exemption now says plainly that nothing executes it.

`generate template` needed a harness rather than a scenario: its argument is a
relative path, so from the workspace root it scaffolds a project into the tree.
The new scratch-directory steps run a documented command verbatim inside a
throwaway cwd, so the structural check can read it and the repository stays
clean.

The `deps` examples are removed from the README rather than witnessed —
`deps install` does not work on either macOS builder — and tracked in
`specs/plans/2026-09-03-restore-the-sealed-dependency-flow-to-the-readme.md`
with both measured failure modes.

## `--peer` did not work as documented

    mvmctl run --peer db.mvm.peer:5432=127.0.0.1:34567 -- ./my-service

`ConnectionRefusedError` inside the guest. A workload has no NIC, so it dials
anything through the guest agent's SOCKS5 listener, and two gates decided
whether that listener exists by asking `NetworkPolicy::allows_egress()` — which
reports the host allow-list only. A peer-only policy is deny-all for ordinary
egress, so neither the endpoint, the identity drive, nor `mvm.vsock_egress=1`
happened, and the dial died before the gate that was meant to decide the peer
name ever saw it.

Adding `--allow-host` makes it pass, which is why nothing caught it: the
hermetic peer scenarios stop at the CLI or call `EgressGate` directly, and the
one place a guest could have dialed a peer also admitted a host. Fixed with
`NetworkPolicy::admits_outbound`, used at both gates. Verified live on HVF: the
documented line now reaches a host service at the peer's target address and
reads back its greeting.

## Builder backends declare what they can do

`BuilderVm::run_stage0` had a default impl returning "implemented for the
libkrun backend only" — false, since qemu had grown one. A second shared
sentence in the hvf builder said the install pipeline "isn't wired for any
backend yet" while libkrun and qemu both served it. Both rotted for the same
reason: a default is inherited by saying nothing, so nothing forces the text to
stay true or a new backend's author to decide.

`run_stage0` and a new `capabilities` are now required. Seven impls had to
answer — one of which the compiler found and a grep for `impl BuilderVm for` did
not, because it is spelled `impl crate::builder_vm::BuilderVm for`. `mvmctl
doctor` renders the matrix, reading a declared table rather than constructing
each backend: construction does real work, so asking through it reported hvf as
*unavailable* on a host where hvf was the working default.

The auto-fallback now also fires when a backend *declines* an operation, not
only when a VM fails to boot. An explicit `--builder` still cannot be
overridden.

`deps install` was the one caller of the infallible builder resolver, so a
missing embed payload surfaced as a panic and "CLI thread exited unexpectedly"
instead of the rebuild instruction the code had already written.

## One mode of a flag no longer witnesses another

The witness matcher compared flag *names* and discarded values, so
`--source download` and `--source compile` were indistinguishable — as were the
security profiles. `run --profile dev` was standing in for `--profile
permissive` **and** `--profile restrictive`.

Discarding values is right for most flags and deliberate: `--image alpine`
standing in for `--image python:3.12` is the same request shape. It is wrong
when the value selects what the command does. Clap already draws that line — an
argument with possible values is an enum — so values are pinned for those and
free for the rest.

Three documented commands lose coverage as a result, and should. `covered` goes
175 → 173: the ratchet's forbidden direction, moved deliberately, because the
number was overstated rather than the coverage lost.

## Suite cost

The two help-width scenarios spawned a fresh `mvmctl` per command path — about
350 spawns of a 137 MB debug binary, measured at 0.120s each. `help_text_for` is
the binary's own help arm with the process removed: same clap tree, same argv,
same `DisplayHelp` error, same output constraint. Because clap does the argv
handling, `--help`, `-h` and `help <path>` still dispatch as they really do, so
no path and no entry-point form is dropped. Three real invocations are kept as
probes, so a binary that stopped reaching that arm could not pass.

This is ~42s on an idle machine. It is **not** a fix for the suite appearing to
hang — that was contention between concurrent BDD runs in separate worktrees,
each shelling out to `cargo run` mid-scenario and serialising on the
machine-global cargo lock.

## Verification

`just bdd` green: 248 scenarios, 247 passed, 1 skipped, 0 failed. Workspace
suites green for every touched crate. The live documented-surface lane is the
remaining gate and runs against this tree.


auser:
The aarch64 TCG failures were a stable runner-lifetime defect, not a guest assertion: three independent jobs were terminated by the runner service at 27–28 minutes despite timeout-minutes: 300; source compilation consumed ~13.5m and bootstrap ~7m, leaving only ~5m for the real workload. Commit 26b41747e4 splits exact-binary preparation from live execution. The preparation job uploads the source-channel subject and release-only bootstrap helper; the dependent smoke restores those immutable binaries on a fresh arm64 runner and performs no rebuild. Structural regression tests, Extended-CI contract tests (28 total), check-workflow-paths, actionlint, bash syntax, formatting, and zero-warning Clippy pass locally. Fresh exact-head Extended CI: https://github.com/tinylabscom/mvm/actions/runs/33832468701.

Issue #3039, opened 2026-09-01
Warm-standby claim fails at the post-restore identity handshake, then silently cold-boots


auser:
The warm path is reached and then fails one step in. `#3005` fixed eligibility — the claim is genuinely attempted now — but the forked child never answers the post-restore handshake, the standby dies, and the run cold-boots while reporting success.

## Reproduction

macOS / HVF, artifact-warm home:

```
$ mvmctl pool warm 1 --image alpine
$ mvmctl pool status
Supervisor standby pool: 1 idle, 0 claimed, 0 parked, 0 dead
  standby-a8b3b097f721f88a · idle · pid 82608 · 2 vcpu / 512 MiB

$ MVM_RESIDENCY=warm mvmctl machine run --image alpine --timeout 120 -- /bin/echo claim-probe
WARN standby claim failed standby=standby-a8b3b097f721f88a
  error=claim standby: forked child vm-4a5b7ba2 never answered the post-restore
        identity handshake: signaling post-restore to vm-4a5b7ba2
claim-probe

$ mvmctl pool status
Supervisor standby pool: 0 idle, 0 claimed, 0 parked, 1 dead
```

## Call chain

- `WorkloadRunner::take_fresh_child_identity` — `crates/mvm-runtime/src/workload_runner/runner.rs:777`
- → `driver.deliver_child_identity(child_vm_name, token, grant_envelope)`
- → `signal_post_restore` — `crates/mvm-vmm/src/post_restore.rs:167`
- the wrapped error text `signaling post-restore to <vm>` comes from line 186, i.e. `signal.post_restore(vm_name)` itself failed, **after** `wait_for_primed_polling` reported the agent reachable.

## The parent is not the problem

`wait_for_guest_agent` (`crates/mvm-backends/src/driver/hvf.rs:770`) probes the authenticated RPC stream and the caller at :534 kills the standby if it never answers, so a captured parent had a serving agent. `pool status` confirms it: the standby was `idle`, not `dead`, before the claim. The failure is on the **child** side after fork/restore.

Its doc comment states the invariant this depends on: *"a parent captured before the agent listener is serving would make every later post-restore identity handshake impossible."* That invariant appears to be held.

## The part that matters more than the bug

[...]

PR #3184, opened 2026-09-06
fix(e2e): the three real failures the Linux lane was hiding

auser:
Refs #3007. Follows #3178, which is what let this lane finish and produce a
verdict at all.

Three scenario failures, three distinct root causes, all verified on real KVM
hardware (Hetzner x86_64, `/dev/kvm`) rather than only in unit tests. None could
have been seen elsewhere: the macOS/HVF documented-surface job is **skipped** on
hosted runners, so Linux/Firecracker is the only lane in the repository that
boots a guest.

## 1. Every OCI image lost its own environment

`pull_core.rs`'s fresh-pull path built its `MaterializeCall` with
`entrypoint: None` — forty lines after computing `config_path`, and never using
it. The sibling cache-hit path had always passed it.

That blob is the image's declaration of `Env`, `WorkingDir` and
`Entrypoint`/`Cmd`. Dropping it means `/etc/mvm/image-runtime.json` is never
written, so the guest falls back to `workload_env::DEFAULT_PATH`.

Measured on hardware, `machine run --image rust -- /bin/bash -c 'command -v cargo; echo PATH=$PATH'`:

```
before  PATH=/run/mvm/bin:/usr/local/sbin:/usr/local/bin:...     (cargo: nothing)
after   PATH=/run/mvm/bin:/usr/local/cargo/bin:/usr/local/sbin:...
        /usr/local/cargo/bin/cargo
```

and `/etc/mvm/` in the materialized rootfs goes from `name variant` to
`entrypoint image-runtime.json name variant`.

The scenario that caught this was written for a *different* bug (#3104 — a
mounted PTY run routed through `/bin/sh -lc`). That fix is intact; it was
failing underneath for this reason, which is exactly why the mount and the
terminal looked implicated and were not. A plain-path scenario now states the
defect as it actually is, so the two can never again be confused.

## 2. A machine with no volumes was refused on an unencrypted disk

`refresh_registered_host_snapshots_*` constructed `MountImageCache` on entry,
and constructing it verifies encrypted backing storage:

```
mount image cache is not on encrypted backing storage
/root/.mvm/cache/mount-images is backed by /dev/md2, which does NOT
appear to sit on a dm-crypt/LUKS mapping
```

That check is about where a *mounted host directory's* bytes come to rest.
Charging it to every launch refused `machine run -d` on any host without
dm-crypt, with no registered volumes and no `--mount` at all, for a cache it
would never write to. **This is a production defect, not just a lane failure.**

The cache now opens only once an attachment carrying a host snapshot is found.
`require_host_encryption` still guards each snapshot's source separately, so
enforcement is unchanged where it applies.

It survived because `MountImageCache::new` is `#[cfg(test)]`-stubbed to skip
the very verification that fails — eager and lazy construction were
indistinguishable under test. The constructor is now injected at that seam and
the test asserts *whether* it runs. Verified the test bites: reverting to eager
construction fails it by name.

## 3. The SDK fixture pinned an architecture

`sandbox_script.py` pinned `alpine@sha256:e7a1a92a…`, which the registry reports
as a single-platform manifest:

```
mediaType: application/vnd.oci.image.manifest.v1+json
config.architecture: arm64 | os: linux | variant: v8
```

So on x86_64 the guest gets aarch64 binaries and `uname` fails with
`Exec format error (os error 8)`. It passes on Apple Silicon and cannot pass on
the only lane that boots a guest.

Repinned to alpine:3.22's **index** digest
(`application/vnd.oci.image.index.v1+json`, amd64 + arm64 + 5 more). Still
immutable and content-addressed, so claim 14's plan-mode refusal of mutable
references is unaffected — verified plan mode still reports `ADMITTED`.

A digest pin looks architecture-neutral and is not. Every other `@sha256:` image
pin under `features/`, `fixtures/` and `scripts/` was audited; this was the only
real one.

## 4. What hid all of it

`SandboxLiveError` captured `mvmctl`'s stderr and rendered only its summary
line, so the CI log said `failed with exit code 1` and nothing else — for a
class whose entire diagnosis is in that stderr, against a docstring promising it
carries "exactly which verb refused and why". Fixed in both SDKs. The first run
with it produced the encrypted-storage refusal immediately, which is how cause 2
was found.

## Validation

- **On real KVM hardware:** `--image rust` PATH before/after; the new scenario's
  exact command; the SDK live script end-to-end (`exit=0`); plan mode still
  `ADMITTED`; rootfs contents inspected with `debugfs`.
- `cargo nextest run --workspace` — **13132 passed**, 0 failed, 22 skipped.
- `just bdd` — 966 steps, 965 passed, 1 skipped, 0 failed.
- Python SDK — 231 passed. TypeScript SDK — 153 passed.
- `cargo clippy --workspace --all-targets -- -D warnings` exit 0; `cargo fmt
  --all --check`; `just check-gated`.
- `check-sprint-append`, `check-plan-names`, `check-declared-backing`,
  `check-conformance`.

## Scope

`pool warm` (#3039) is the third lane failure and stays separately tracked; it
is now gated into the "did NOT run" tally rather than failing.

The wiring in fix 1 is witnessed by the live scenario, not by a hermetic test —
the pull path does network I/O before it reaches the materializer, so there is
no seam that proves the config arrives without booting. Stating that plainly
rather than adding a test that would not have caught it.

#3007 should stay open until an Extended CI run reports the lane clean.

### DIFF
--- a/crates/mvm-cli/src/commands/image/pull_core.rs
+++ b/crates/mvm-cli/src/commands/image/pull_core.rs
@@ -365,12 +365,19 @@ fn pull_image_ref(
     let rootfs_path = format!("rootfs/{manifest_hex}-{runtime_tag}/rootfs.ext4");
     let rootfs_abs = cache_root.join(&rootfs_path);
     super::cache::write_deferred_nodes(cache_root, &manifest.digest, &deferred_nodes)?;
+    // The config blob written above is the image's own declaration of `Env`,
+    // `WorkingDir` and `Entrypoint`/`Cmd`. Materializing without it discards
+    // all three: the guest then falls back to `workload_env::DEFAULT_PATH`, so
+    // an image's own tools are off `PATH` even though the binaries are in the
+    // rootfs. The cache-hit path beside this one has always passed it; this
+    // one computed `config_path` forty lines earlier and dropped it.
+    let entrypoint = oci_entrypoint_from_cache_path(cache_root, config_path.as_deref())?;
     inject_runtime_and_materialize(super::materialize::MaterializeCall {
         cache_root,
         unpacked_root: &unpacked_root,
         rootfs_abs: &rootfs_abs,
         image_label: &image_ref.canonical(),
-        entrypoint: None,
+        entrypoint: entrypoint.as_ref(),
         sealed: false,
         deferred_nodes,
         evidence: None,
--- a/crates/mvm-cli/src/commands/vm/volume.rs
+++ b/crates/mvm-cli/src/commands/vm/volume.rs
@@ -531,8 +531,46 @@ fn refresh_registered_host_snapshots_with_service(
     volume_service: &LocalVolumeService,
     vm_name: &str,
 ) -> Result<()> {
-    let cache = crate::mount_cache::MountImageCache::new()?;
-    for attachment in volume_service.list_attachments(vm_name)? {
+    refresh_registered_host_snapshots_using(
+        volume_service,
+        vm_name,
+        crate::mount_cache::MountImageCache::new,
+    )
+}
+
+/// The cache constructor is a parameter so a test can observe *whether* it
+/// runs.
+///
+/// It has to be: `MountImageCache::new` is `#[cfg(test)]`-stubbed to skip the
+/// encrypted-backing verification, so under test the eager construction this
+/// function used to do was indistinguishable from the lazy one. The check that
+/// failed in production did not exist in the build that tested it, which is
+/// how a launch-blocking refusal reached a live lane unnoticed.
+fn refresh_registered_host_snapshots_using(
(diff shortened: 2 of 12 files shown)


auser:
Filed the volume.rs half separately as #3186 — `machine run -d` refused on any host without dm-crypt is a broken user-facing verb, not an e2e lane failure, and it should be visible as such rather than buried in this PR. Raised by another session reviewing this change; agreed. The fix stays here.

Also dispatched Extended CI on this branch: [34003859189](https://github.com/tinylabscom/mvm/actions/runs/34003859189). The Linux/Firecracker documented-surface lane is the only one that boots a guest, so it is the only place these two fixes can be witnessed — worth having that verdict against the branch rather than discovering it after merge.

PR #3185, opened 2026-09-06
fix(ci): write an entrypoint marker the guest agent can accept

auser:
## The hosted no-KVM witness refused its own fixture

`Hosted no-KVM bundle smoke (QEMU TCG)` has never passed. It was added in its
current form on 2026-09-04 (#3170) and every run since ends the same way:

```
mvm-guest-agent: entrypoint validation failed at boot:
  canonicalize /bin/busybox sh -c 'exit 7': No such file or directory (os error 2)
[mvm] invoke: entrypoint invalid: ...
::error::expected guest exit code 7 from installed bundle, got 1
```

The product was right and the fixture was wrong. `/etc/mvm/entrypoint` is a
marker, not a command line. `EntrypointPolicy::production()` resolves its
contents as an absolute path to a wrapper under `/usr/lib/mvm/wrappers/`, and
the only other shape the agent accepts is `sealed_script_marker()`, which takes
the marker file *itself* as the program when it begins with `#!` and carries
mode 0755 root:root on the same filesystem as `/etc/mvm`. A command line
satisfies neither, so the agent refused the entrypoint at boot exactly as it is
supposed to.

There were **two** defects. `debugfs write` creates its file 0644, so even a
correct wrapper path would have been refused on mode.

## Verified against the shipped bytes

The `boot-image/v0.1.3` release rootfs was downloaded and read with `debugfs`.
Its marker is a `#!/bin/sh` script at mode 0755, uid 0, gid 0. The fixture now
writes that same shape rather than inventing a third one.

The rewritten step was then extracted from the YAML and executed verbatim
against those release bytes on x86_64 Linux:

- the marker lands at mode 0755, uid 0, gid 0;
- `stat -c %d` reports the same device for `/etc/mvm` and the marker;
- `veritysetup format` still yields a 64-character roothash over the edited image;
- `chroot` into the mounted rootfs, running `/etc/mvm/entrypoint` in the image's
  own userspace, **exits 7**.

That is every condition `sealed_script_marker()` checks, plus the exit code the
lane asserts. The boot itself is answered by dispatching the lane.

## Two guards now sit earlier than the guest

The lane is nightly, so it cost a full cycle per attempt to learn anything, and
the only thing checking the marker was a guest three jobs downstream.

- The fixture step asserts its own output — mode, ownership, shebang — and fails
  at assembly time with the offending `stat` in the message.
- `no_kvm_pipeline_prepares_a_bounded_bundle_then_boots_with_tcg` pins the
  shebang line and the `sif … mode 0100755` call, so a revert to a command line
  fails on **every pull request**. Confirmed red against the old fixture before
  being kept.

## Scope

`entrypointArgv` stays `["/bin/busybox", "sh", "-c", "exit 7"]` — still true of
what runs, since the marker `exec`s exactly that, and inert for this lane, which
passes no `--stdin`.

The other two Extended CI failures in #3007 are separately owned and land on
their own.

Gates: `cargo fmt --all --check`, `cargo clippy --workspace -- -D warnings`,
`cargo nextest run --workspace` (13132 passed), `xtask check-all` (67 gates
clean), `actionlint`.

Refs #3007.


auser:
## Live verdict: the lane is green for the first time

Extended CI dispatched on this branch —
[run 34000666379](https://github.com/tinylabscom/mvm/actions/runs/34000666379).
The whole no-KVM chain passed:

```
Prepare hosted no-KVM smoke binaries:               success
Bootstrap hosted no-KVM launch artifacts:           success
Prepare bounded signed bundle for hosted no-KVM boot: success
Hosted no-KVM bundle smoke (QEMU TCG):              success
```

Two step outcomes carry the result:

- **`Assemble and export the bounded exit-code fixture`** — green, so the three
  new in-step assertions (mode 0755, root:root, `#!/bin/sh` … `exit 7`) all held
  against the real release rootfs on the runner, not just against the copy I
  tested by hand.
- **`Install and boot the sealed bundle under QEMU TCG`** — green. That step's
  only success path is `actual -eq 7`; anything else prints
  `::error::expected guest exit code 7 …` and exits 1. So the guest booted under
  TCG, the agent accepted the marker, and the entrypoint returned 7.

`Dump diagnostics on failure` is `skipped`, which is the corroborating negative.

Raw log text is not retrievable yet — GitHub withholds it until the whole run
finishes, and the Linux/Firecracker documented-surface job is still going. The
step semantics above are what the verdict rests on; I'll attach the log line if
anything looks off once it lands.

PR checks: all pass, none pending.

auser:
## Log confirms the exit code literally

The run has finished, so the logs are retrievable. Closing the caveat from the
comment above — the verdict no longer rests on step semantics:

```
Trusted key_id 650317d24a94cd00c7c8df5a8eeca9ec
installed bundle content address: 4be2a3b011dd2499984386daf74e0225ada67dfae9efe00a2e683405fe745857
ok: installed x86_64 bundle exited 7 under QEMU TCG
```

And the corroborating negative: `entrypoint validation failed` appears **zero**
times in the job log, against every prior run of this lane where it was the
first thing the guest printed.

Nothing else in the run regressed. The one failing job,
`Documented surface e2e (Linux, Firecracker)`, fails on the two scenarios
already owned in #3184 — `an interactive mounted run preserves the image
environment` and `a runtime-SDK script reaches DevOnly surfaces under an
explicit dev profile` — which is expected, since this branch changes only
`ci-full.yml`, an xtask assertion and a delivery note. It is main's result, not
this PR's.

PR #3187, opened 2026-09-06
fix(image,volume): carry the image environment on a cold pull, and build the mount cache on first need

auser:
Fixes both Linux/Firecracker failures from the first Extended CI run that was
allowed to finish. Refs #3007.

Each passes on macOS/HVF, which is why neither surfaced until #3178 stopped the
nightly lanes cancelling themselves. That run reported `312 scenarios (309
passed, 1 skipped, 2 failed)` — the Linux verdict that had not existed before.

## The image's environment was discarded on a cold pull

`pull_core.rs` passed `entrypoint: None` into materialization, throwing away the
image's own runtime config: argv, working directory, and the `Env` that carries
its `PATH`. A guest booted from a freshly pulled image then cannot resolve a
binary the image puts on its own PATH — `command -v cargo` finds nothing in
`rust`, which is what `an interactive mounted run preserves the image
environment` asserts.

The already-cached arm a few lines above resolves it correctly via
`oci_entrypoint_from_cache_path`; only the cold-pull arm did not. That is why
reproducing it needs an empty cache, and why a warm developer machine never sees
it.

Worth noting against #3104, which fixed the same *symptom* from the other end by
dropping `req.dir_shares.is_empty()` from `direct_pty_inline_argv`. That fix was
correct and this one is independent: the routing was repaired while the
environment was still being discarded upstream, so the scenario stayed red for a
second reason.

## Every launch paid for a capability only some launches need

`refresh_registered_host_snapshots_with_service` built `MountImageCache` before
the loop that uses it. `MountImageCache::new` can fail on a host without the
encryption support it wants, so constructing it up front made that failure the
answer for *every* launch — including the common one with no registered host
snapshot to refresh, where the cache is never touched.

Built on first need instead. A machine that registers no host directory has
nothing to do there and should not be refused for lacking a capability that work
would have needed.

**This one is not limited to the e2e lane.** It refused `machine run -d` on any
host without that support. Nothing on macOS exercises the path, so no local run
would have caught it.

## Verification

`cargo nextest run -p mvm-cli`: 2100 passed, 3 skipped. Both changes compile
clean under `-D warnings`.

The scenarios themselves are Linux-only and cannot be exercised on this host —
the Extended CI lane is the confirming signal, and it can now finish.

### DIFF
--- a/crates/mvm-cli/src/commands/image/pull_core.rs
+++ b/crates/mvm-cli/src/commands/image/pull_core.rs
@@ -365,12 +365,21 @@ fn pull_image_ref(
     let rootfs_path = format!("rootfs/{manifest_hex}-{runtime_tag}/rootfs.ext4");
     let rootfs_abs = cache_root.join(&rootfs_path);
     super::cache::write_deferred_nodes(cache_root, &manifest.digest, &deferred_nodes)?;
+    // The image's own runtime config — argv, working dir, and the `Env` that
+    // carries its `PATH`. Passing `None` here discarded all three, so a guest
+    // booted from a freshly pulled image could not resolve a binary the image
+    // puts on its own PATH: `command -v cargo` finds nothing in `rust`, and the
+    // documented interactive-mount example fails on exactly that.
+    //
+    // The already-cached path a few lines above resolves it this way; only this
+    // arm did not, which is why the failure needed a cold pull to reproduce.
+    let image_runtime_config = oci_entrypoint_from_cache_path(cache_root, config_path.as_deref())?;
     inject_runtime_and_materialize(super::materialize::MaterializeCall {
         cache_root,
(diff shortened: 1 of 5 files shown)


auser:
Closing as redundant — both fixes are already on main, landed by the session working from the Hetzner box while this was being written.

Their `pull_core.rs` change is substantively identical (`oci_entrypoint_from_cache_path(cache_root, config_path.as_deref())`). Their `volume.rs` change is better than mine: they parameterised the cache constructor so a test can observe *whether* it runs, which matters because `MountImageCache::new` is `#[cfg(test)]`-stubbed — under test, eager and lazy construction were indistinguishable, so my version changed the behaviour without being able to prove it.

No work lost; nothing here that main does not already have.

PR #3194, opened 2026-09-06
release: v0.18.0-rc.1

auser:
Version bump + git-cliff changelog for v0.18.0-rc.1. Merge via the queue, then `just release-tag 0.18.0-rc.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to mvm are documented in this file.
 
-## [0.18.0] — 2026-09-02
+## [0.18.0-rc.1] — 2026-09-06
 
 ### Added
 - **mvm-cli**: Fetch + apply the signed pack revocation list (Plan 213 §I)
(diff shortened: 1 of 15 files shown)


auser:
Amended: dropped the superseded `## [0.18.0]` changelog section.

That release was prepped on 2026-09-02 and never tagged — the last published release is v0.17.0 — so its section described a release that does not exist. git-cliff generates the rc.1 section from `v0.17.0..HEAD`, which covers the same range, and keeping both would have shipped the content twice under two headings.

Checked rather than assumed. Diffing the two sections' entries showed 10 lines in the old one absent from rc.1, so the `--unreleased` range was not simply a superset. Each turned out to be a **pre-squash branch subject** superseded by its squash-merge counterpart:

- `Gate a release on evidence the documented surface ran` → `**release**: Make the documented-surface gate actually gate a release` (#3117, `740d59d482`)
- `Stream host file contents into the image instead of buffering them` → `**fs**: Stop copying every file's bytes into the ext4 plan`

The rest are a merge commit and internal specs/notes entries. Nothing user-visible is lost, and rc.1 is the canonical record of what is actually on main.

PR #3293, opened 2026-09-15
docs(plan): claude code as an mvm workload

aneyzberg:
## Summary

New plan: run Claude Code inside an mvm microVM — an interactive dev workbench lane (accessible tier, `machine console`) and a sealed headless lane (`claude --bare -p` over the stdin input plane) off one `mkGuest` flake, distributed as an in-repo example first and an `mvm-templates` remote-registry entry second.

The plan names the three constraints that shape the design up front (no writable host-directory share; the API key guest-held in the interim; interactive access is dev-tier by claim 15) and carries W5 to move the key out of the guest onto the shipped placeholder-substitution path, with the ADR-023 terminator as the named fallback.

`Backing: preview` / `Validation: none`; `check-plan-names`, `check-declared-backing`, `check-doc-claims` all green.

The W0 spike (research + live HVF smoke runs) has already executed against this plan — findings land in the stacked follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3299, opened 2026-09-16
feat(examples): claude-code guest — interactive and sealed headless profiles

aneyzberg:
## Summary

Stacked on #3294 (which is stacked on #3293). Delivers W1 of the claude-code workload plan: `examples/claude-code/` — one `mkGuest` recipe, two profiles.

- **`default`** — accessible workbench: bash behind `machine console`, `claude` on PATH.
- **`tenant-headless`** (selected as `--profile headless`) — sealed image whose baked entrypoint is `claude --bare -p`, fed by the stdin input plane.

The binary is the first-party native **musl** build, pinned by version + SHA-256 from the release manifest at downloads.claude.ai (both aarch64 and x86_64) and patchelf'd to the Nix musl loader — no Node.js and no glibc in the image. The wrapper bakes the egress fail-loud guard, telemetry kill switches, `NODE_USE_ENV_PROXY=1`, ripgrep/gitMinimal on PATH, workspace-disk state via `CLAUDE_CONFIG_DIR`, and read-only key-file pickup (interim posture; W5 is the placeholder-substitution destination). `mvm.toml` admits the measured three-host minimum and turns on AI token metering.

## Validation

- Both profiles built through the builder VM; sidecars to spec (interactive: accessible shell tier; headless: `sealed: true`, `accessible: false`, recorded argv — what admission needs for the stdin grant).
- Full hermetic BDD suite green, including the s29 doc-example gates: every README command parses against the real clap tree and is recorded in the regenerated coverage ledger.
- `check-declared-backing` green (plan W1 ticked in the same change, with findings: the `tenant-<profile>` flake-attribute convention, and the ledger-regeneration requirement).

After #3294 merges: retarget to main and rebase to drop the base's commits.


examples/claude-code/flake.nix
```diff
@@ -0,0 +1,145 @@
+{
+  description = "Claude Code agent guest — interactive workbench and sealed headless profiles";
+
+  # Two profiles off one image recipe:
+  #   `default`  — accessible (dev-tier) image: `entrypoint.shell` puts an
+  #                interactive bash behind `mvmctl machine console`, with
+  #                `claude` on PATH.
+  #   `headless` — sealed image: `entrypoint.command` runs `claude --bare -p`
+  #                reading its task from the stdin input plane
+  #                (`machine run --entrypoint --stdin -`).
```
auser, line 10: Okay, we're actually missing something here. Like I had `tiers` where everything locally _always_ start with the `dev` tier and `mvmd` is the only authorized client to start things in prod. 

That decision was made to simplify the UX/DX.

examples/claude-code/flake.nix
```diff
@@ -0,0 +1,145 @@
...
+  #
+  # The binary is the first-party native musl build, pinned by version and
+  # SHA-256 from downloads.claude.ai's per-version manifest.json. It is
+  # dynamically linked against musl only (single DT_NEEDED), so the image
+  # needs no glibc and no Node.js — just musl's own loader, supplied from
+  # pkgs.musl via patchelf. Node/npm are NOT in this image; Claude Code's
+  # own runtime is self-contained (Bun-compiled).
+  #
+  # The `github:tinylabscom/mvm` pin is load-bearing: a source-checkout
```
auser, line 19: I think we can simplify this nix template. It might be worth spending some time thinking about how templates are "stacked", e.g.

alpine -> node -> openclaw, et.al.

examples/claude-code/mvm.toml
auser, line 1: I like the `mvm.toml` format, what do you think about having a `Dockerfile`-like experience or think this `mvm.toml` file covers it?

features/suites/s29_doc_examples/docs_coverage.toml
```diff
@@ -259,6 +259,7 @@ uncovered = [
     "mvmctl machine console agent-dev",
     'mvmctl machine console agent-dev --command "uname -a"',
     "mvmctl machine console api-dev",
+    "mvmctl machine console claude",
```
auser, line 262: is `mvmctl machine console <name>` what this means? If it could be other agents, like codex, gemini, deepseek, etc. would that change?

public/src/content/docs/guides/config-secrets.md
auser: This is wrong... we HAVE to remove `--volume`. It was deprecated in favor of `--mount`. Please identify where you got this information

auser (commented):
Love this example set up -- we should follow this for future examples, just wrapping in the [templates](https://github.com/tinylabscom/mvm-templates) repo for extensibility

public/src/content/docs/guides/config-secrets.md
aneyzberg: [Claude] Provenance: that sentence predates this PR — it was added by #1602 (2026-07-09, the --volume→--mount rename) and this PR's rewrite preserved it. The alias is still live in the CLI (visible_alias = "volume", crates/mvm-cli/src/commands/vm/exec.rs). Fixed in 2246dcc93: the guide no longer advertises the alias. Removing the alias itself + sweeping the 8 other doc files that mention it is queued as its own small PR so this one stays reviewable.

features/suites/s29_doc_examples/docs_coverage.toml
aneyzberg: [Claude] That ledger row classifies the CLI verb path 'machine console' — it's the tier for the command itself (parse-gated on every PR, live-witnessed in the s34 scenario), agent-agnostic. Running codex/gemini/etc. inside the guest changes nothing here: they'd attach through the same verb, and only the docs' example commands (which the extractor reads) would add rows.

examples/claude-code/flake.nix
aneyzberg: [Claude] Recorded in the plan (Review feedback section, 2246dcc93): locally everything starts dev-tier and mvmd is the only authorized prod client — so Lane B's --prod framing is an mvmd-lane destination, and the local deliverable is the sealed image + stdin plane. The example's two profiles stay, but the prod promise moves out of local mvmctl scope.

examples/claude-code/flake.nix
aneyzberg: [Claude] Iceboxed in the plan (2246dcc93): template stacking as consumable layers (alpine → node → claude-code) so flakes compose instead of restating the base.

examples/claude-code/mvm.toml
aneyzberg: [Claude] Opinion, also noted in the icebox: mvm.toml covers the declarative surface well and pairs naturally with the flake owning image contents; a Dockerfile-like imperative layer would blur who owns the rootfs (the flake) vs the run config (the toml). If stacking (your other comment) lands, most of what people reach for Dockerfiles for — layering — arrives without the imperative model.

Issue #3283, opened 2026-09-15
An off-the-shelf HTTPS client cannot use credential substitution: the proxy refuses CONNECT


auser:
This is the blocker for "an AI agent boots in the sandbox". A real agent
CLI cannot reach its model API with the key substituted.

**What happens.** When a workload has secret placeholders, the invoke path sets
`HTTP_PROXY`/`HTTPS_PROXY` to the guest forward proxy on port 18080
(`crates/mvm-cli/src/commands/vm/invoke.rs:1339-1350`). That proxy accepts only
absolute-form requests (`crates/mvm-agentd/src/forward_proxy.rs:62-66`). A
client that honors `HTTPS_PROXY` for an `https://` URL — curl, Python
`requests`, most Node HTTP agents — tunnels with `CONNECT`, which gets a 502
(`forward_proxy.rs:135`). Node's built-in `fetch` ignores `HTTPS_PROXY`
entirely. The proxy also serves one request per connection, buffers the whole
response (so a streamed model response arrives at once at the end), and caps
requests at 16 MiB.

**Why.** Substituting a header on a `CONNECT` tunnel means terminating TLS
inside the host endpoint. The terminator exists, but the workload runner never
enables it: `crates/mvm-runtime/src/workload_runner/runner/spawner.rs:107-110`
passes `terminator_listen: None` and `tls_intermediate: None`, so no per-VM
egress CA is provisioned and the CA env variables at `invoke.rs:1289-1332` are
never set on this path.

**Result today.** The only client that works is one hand-written to send
absolute-form requests. No shipped agent CLI does that.

**Do:** wire the SNI terminator and the per-VM egress CA onto the workload
runner path, so `CONNECT` to a host with a bound secret is terminated,
substituted, re-encrypted and audited, and the CA bundle is injected into the
guest env. Stream responses instead of buffering them. Refuse a `CONNECT` to a
bound host rather than relaying it opaquely, so a placeholder can never leave
the host unsubstituted.

**Witness:** an unmodified HTTPS client honoring `HTTPS_PROXY` completes a
request to a bound host, the destination receives the real credential, the
guest never held it, and `secret.substituted` is chain-signed.

Plan: `specs/plans/2026-09-15-agent-sandbox-drive-plane.md`. Epic #3275.

auser:
Design investigation done; recorded in `specs/plans/2026-09-15-agent-sandbox-drive-plane.md` (WS-S) on the branch for PR #3256.

[...]

Issue #3261, opened 2026-09-15
The SDKs must never shell out to mvmctl: one versioned C ABI, and a gate


auser:
`crates/mvm-sdk/sdks/python/mvm/_sandbox.py:728` and
`crates/mvm-sdk/sdks/typescript/src/_sandbox.ts:613` build an argv and run
`mvmctl` once per call. That is a second entrypoint to every verb, it cannot
stream, and it costs a process per call. **This is bad design and it goes.**

It is not an accident.
`specs/adrs/027-cli-surface-consolidation.md:167-174` chose it deliberately,
named the blocker (`mvm-sdk` sits below the runtime, so linking the local
backend would form a cycle), and called the convergence onto `MvmClient`
follow-on work — which nobody did. The cycle is real today:
`mvm-client` → `mvm-hostd` → `mvm-sdk` (`crates/mvm-client/Cargo.toml:43`).

So the fix is **not** a new edge out of `mvm-sdk`.

**Do:**
- Add `crates/mvm-hostlib` at the top of the dependency graph, beside
  `mvm-cli`. It links `mvm-client` and exposes one versioned C ABI over the
  `MvmClient` trait plus the drive verbs. Nothing depends on it, so a cycle is
  impossible by construction.
- Carry an ABI major/minor and a `mvm_hostlib_abi_is_compatible` entry point the
  bindings must call before use, so a mismatched pair fails loudly instead of
  reading a moved struct.
- Rewrite `_LiveTransport` and its TypeScript twin onto that ABI, in-process.
  Streaming becomes possible; the process-per-call cost goes away. The SDK does
  not spawn `mvmctl`, and does not spawn a long-lived helper either.
- Add `xtask check-no-cli-shellout`: no file under `crates/mvm-sdk/sdks/` may
  reference `subprocess`, `spawnSync`, `execFile`, or `$MVM_CLI_BIN`. Prose is
  not what should hold this.
- Amend ADR-027 so it stops reading "still deliberately shells out", and update
  `specs/plans/2026-08-15-sdk-binding-fan-out.md`, whose costing assumes surface
  B is an argv builder.

**Acceptance:** `rg -n 'subprocess|spawnSync|execFile' crates/mvm-sdk/sdks/`
returns nothing, and the gate fails if it ever does again.

Plan: `specs/plans/2026-09-15-agent-sandbox-drive-plane.md` WS2.

auser:
[...]

PR #3333, opened 2026-09-16
feat(run): --secret on the run surface; pin the substitution path's real behavior

aneyzberg:
## ⚖️ Decision needed before this leaves draft

**How should the host insert the API key into real HTTPS traffic?** Substitution itself works (M3 below), but the typed flow can't carry Claude Code's actual `https://api.anthropic.com` traffic yet (M2 below). Two ways to finish, pick exactly one:

**Option A — teach the typed flow an https-upstream leg** *(recommended)*
The guest sends absolute-form `http://api.anthropic.com/...` to the loopback forward proxy; the host endpoint, seeing a flow whose signed plan carries a secret binding for that destination, originates **TLS** upstream instead of plaintext, and the policy gate learns to charge such a flow to port 443 rather than 80.
- Pros: small and local (the endpoint's forward leg + one gate rule); no TLS interception machinery; the guest's request is already parseable so nothing new touches TLS internals; scope is naturally pinned to plan-bound destinations.
- Cons: the guest→proxy hop for that flow is plaintext HTTP on loopback (inside the guest only, and it carries a placeholder, never the key); requires setting `ANTHROPIC_BASE_URL` in the workload env.

**Option B — wire ADR-023's TLS terminator for CONNECT**
The endpoint terminates TLS on CONNECT flows to bound destinations using the per-VM CA the guest already trusts (`NODE_EXTRA_CA_CERTS` injection exists), parses the request, substitutes, re-originates TLS upstream.
- Pros: fully transparent — no `ANTHROPIC_BASE_URL`, works for any client, matches the ADR's design.
- Cons: substantially more machinery (per-VM cert minting, in-endpoint TLS server, MITM correctness surface); the ADR's nft-redirect premise is stale (no guest NIC), so the design needs a revision pass first; larger claim-13/16 review burden.

Decide by commenting here; that choice becomes the next commit series on this branch. Either way the finish line is the same: `--secret anthropic` + the example flake = the guest never holds the raw key.

---

## Summary

W5 of the claude-code workload plan (see #3293), based on current main — **draft** until the #3293 → #3299 stack lands, then this gets rebased so its copy of the plan file becomes an incremental diff.

**M1 — shipped mechanism.** `--secret NAME[:HOST,...]` moves from the unreachable `up::Args` onto the shared `RunArgs`, so `mvmctl run` and `mvmctl machine run` bind a stored secret for host-side substitution. Fail-closed resolution (unknown secret / missing binding / out-of-policy destination refuse before boot), guest env var named by a new optional `env_var` field on the service catalog (`anthropic` → `ANTHROPIC_API_KEY`, …), persistent machines re-resolve per start and pin the reference so `secret rm` refuses while referenced. The guest sees only the minted `mvm-secret-<hex>` placeholder via the endpoint handshake.

**M3 — substitution correctness pinned.** The claim-13 `substitute` witness family gains the Anthropic credential position: `x-api-key` placeholder rewriting (dispatch core + service process path), unbound-destination refusal with nothing forwarded and no value in the refusal text, and a launch-env handshake asserted placeholder-only.

**M2 — the honest blocking fact.** The typed HTTP flow does **not** upgrade `http://` to TLS upstream, and the claim-10 gate keys the scheme's default port — so a 443-only policy refuses absolute-form http before the forward leg. Steering Claude Code's `ANTHROPIC_BASE_URL` onto the substitution path is therefore **blocked today**; tests pin both facts. The decision section above is this fork.

**M4 — streaming boundary recorded.** The guest relay drains the typed flow to `HttpComplete` before answering the workload, so an SSE body arrives whole after upstream close; a test records the boundary and the plan carries what an incremental front would change.

## Validation

`cargo fmt --all` clean, workspace clippy `-D warnings` clean, nextest green for mvm-cli / mvm-hostd / mvm-agentd / mvm-core.


crates/mvm-agentd/src/flowmux_sync.rs
auser: This is one of the heaviest lifts in the refactor -- the synth network traffic over vsock

crates/mvm-cli/src/commands/vm/up/mod.rs
auser: Really happy (I think) for removing the `--secret` from the CLI

crates/mvm-cli/src/commands/vm/up/oci_persist.rs
auser: Does this keep the `--secrets` flag in the CLI for OCI images? Isn't that an anti-pattern in this repo?

crates/mvm-cli/src/commands/vm/exec.rs
auser: I think this is fine as long as it's in the dev tier

crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs
auser: This file needs splitting... it's ENORMOUS

auser:
I think Option A is definitely the way we want to go. An extra hop is already baked in and it's "hop" will be negligible as it's on the host from the microvm.

aneyzberg:
[Claude] Option A acknowledged — but events overtook this draft while it sat: main shipped #3338, which terminates CONNECT to bound hosts (per-VM name-constrained CA, gate/meter/redaction applied, refusal never degrades) — the Option-B shape rebuilt for vsock, and its motivating issue #3283 is the same fact this PR's M2 tests pinned. So the fork is closed by shipped code, and A's rationale (smallest path to substitution) no longer applies: an unmodified client now gets substitution via CONNECT with no base-URL steering. Proposed disposition: salvage pass keeping M1 (the --secret run-surface flag — a CLI convenience --from-workload-ir doesn't replace) and the x-api-key witnesses if upstream lacks them; drop the M2 behavior pins #3338 invalidated. Say the word and I'll rework the branch to that scope.

auser:
Heads-up on a file move that will touch this PR when you rebase. For #3261, boot admission is moving out of the CLI into `mvm-client`, so the CLI and the SDK's host library admit through one path. The moves are plain renames (`git mv`), so rebase rename detection should carry your hunks to the new location:

| was | now |
|---|---|
| `crates/mvm-cli/src/commands/vm/up/admission.rs` | `crates/mvm-client/src/admission/mod.rs` |
| `…/vm/up/audit.rs`, `…/vm/up/policy.rs` | `crates/mvm-client/src/admission/{audit,policy}.rs` |
| `…/vm/policy_resolver.rs`, `…/vm/entrypoint_resolve.rs`, `…/vm/agent_verbs.rs` | `crates/mvm-client/src/admission/…` |
| `…/commands/shared/grants.rs` | `crates/mvm-client/src/admission/run_grants.rs` |

Items the CLI calls are now `pub` and reached as `mvm_client::admission::…`. I'll link the PR here once it is open.

auser:
The move is #3476.

Issue #3302, opened 2026-09-16
Split network_endpoint_proxy.rs: 5,281 lines is why parallel pipelines keep appearing in it


auser:
`crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs` is 5,281 lines with ~197 functions, and `SubstitutionService` is spread across three separate `impl` blocks (`:1218`, `:2728`, `:4802`).

That is well past the 1,500–2,000 line ceiling we hold everything else to, and the size is not incidental — it is why the file keeps growing parallel pipelines that skip controls:

- `handle_https_terminator` (`:1788`) is a second request path beside `process`/`process_body_stream`. It never consults the egress gate, never meters the AI budget, never runs reversible replacement, and buffers whole responses. It was written for a topology that no longer exists and nothing calls it.
- The scan layer in `supervisor/network/stages.rs` is a third one, with no callers at all (#3297).
- The audit of a substitution happens on response completion in one path and nowhere in another (#3286, #3300).

When one file owns the whole decision surface, a new path added to it looks local and is not.

**Do:** split by responsibility, not by line count. A first cut that the existing types already suggest:

- the forwarder and its SSRF-hardened resolution (`HardenedForwarder`, `ForwardResponse`, `ForwardStreamResponse`)
- request preparation and the binding check (`prepare_request`, `prepare_flow`, `PreparedFlow`)
- streaming and redaction (`process_body_stream`, `StreamingRedactor`)
- the AI budget (`AiRequestMeta` and its impl block)
- audit emission (already partly in `secret_audit.rs` — finish the move)
- service assembly (`from_plan`, `FromPlanInputs`, the builder methods)

Each piece should be testable without constructing the whole service. The measure of success is that adding a new request path requires touching the seam that owns admission, rather than being able to add one beside it.

Two things to do in the same pass, because they are the reason the split is worth doing:

1. Delete `handle_https_terminator` and its supporting glue once #3283's termination path is live — it is the old listener-driven pipeline and keeping a second one is how this happened.
2. Make the egress gate a required constructor argument rather than an optional attachment (filed separately).

Related: #3283, #3286, #3297, #3300.

auser:
Design done against `origin/main` at `44799f0cac`. It covers this issue, #3301, and the four issues that land on the same file afterwards (#3300, #3286, #3297, #3339), so they are sequenced together.

## What the file actually is

[...]

Issue #3301, opened 2026-09-16
The egress gate is optional on SubstitutionService, so an endpoint that enforces nothing is well-typed


auser:
`SubstitutionService` is built without its egress gate, and the gate is bolted on afterwards.

`from_plan` (`crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs:1528`) constructs the service with `egress_gate: None` (`:1235`). The gate arrives later through an optional builder call — `with_egress_gate` / `with_shared_egress_gate` (`:1464`, `:1470`) — and only when `projection.gate.is_some()` at the one production assembly site (`network_endpoint.rs:460-462`). The field is `Option<Arc<EgressGate>>` (`:1001`).

The consequence: the claim-10 check inside `prepare_flow` sits behind `if let Some(gate)`. A service without one is a well-typed value that enforces nothing. Nothing in the type system says a VM's endpoint must have a gate; it holds only because the single caller happens to attach one.

This is not hypothetical. The security review of #3283 went looking for a peer-route bypass and found that the backstop it would have relied on — `prepare_flow`'s explicit peer refusal — is inside that same `if let`, so the argument for why the bypass was unreachable had to be made from the caller rather than from the type. The fix there was a guard at a different layer, which is the right fix for that path but does not address why the question was hard to answer.

**Do:** make the gate a required constructor argument, so an endpoint with no claim-10 enforcement is unrepresentable rather than merely not-currently-constructed. `tls_intermediate` should stay optional — whether a VM can terminate genuinely varies — but the gate does not vary: every workload VM has a network policy, and default-deny is still a policy.

Check the test constructors while doing this: several build a service with no gate, which is exactly the shape that should stop compiling. A test that wants no policy should pass an explicit deny-all gate, which is also a more faithful double.

Related: #3283, #3297, #3300.

auser:
Sequenced with #3302. The design, the PR order and the decisions taken for this issue are in https://github.com/tinylabscom/mvm/issues/3302#issuecomment-5697626407.

Issue #3300, opened 2026-09-16
A claim-10 gate refusal is not recorded in the chain-signed audit log


auser:
A claim-10 refusal on the typed HTTP path leaves no entry in the chain-signed log.

`prepare_flow` refuses a destination the egress gate denies and carries its own comment saying so: "Audit of the claim-10 denial is a later increment" (`crates/mvm-hostd/src/supervisor/network_endpoint_proxy.rs:2472,2497`). The neighbouring failures *are* audited — `audit_fail_closed` and `audit_placeholder_dropped` both write — so the gap is specifically the policy denial, which is the one an operator would most want to see. A workload that repeatedly probes destinations it is not admitted to reach produces a clean chain.

This predates the CONNECT-termination work in #3283 and is equally true of the typed HTTP path that shipped before it, so it is not a regression. It matters more once termination lands, because from then on any TLS client in the guest reaches this refusal rather than only the SDK's typed flow.

**Do:** emit a chain-signed refusal when the gate denies a destination, carrying the destination and a fixed reason label and no request bytes, headers or URL — the shape `emit_secret_flow_refused` uses. Then the statement "every refusal on a credentialed path is audited" becomes true, which today it is not.

While there: the terminated flow's 502 arm (`terminator/flow.rs:274-277`) inherits the same gap, since most of what reaches it is audited underneath but this case is not.

Found during the security review of #3283.

Related: #3286 (substitution audited only on completion), #3297 (dead scan layer), #3283.

auser:
Sequenced with #3302. The design, the PR order and the decisions taken for this issue are in https://github.com/tinylabscom/mvm/issues/3302#issuecomment-5697626407.

Issue #3297, opened 2026-09-16
The host-side egress scan layer has no callers, and a placeholder in a request body is ungated


auser:
The whole `supervisor/network/stages.rs` scan layer has no callers outside its own file, and doc comments elsewhere still describe it as enforcing things.

`rg` finds no caller for `build_egress_scan`, `PlaceholderLeakScan`, `MandatoryDenyEgressScan`, `SshBannerProtocolDenyScan`, `L4PolicyScan` or `ScanStage` anywhere outside `crates/mvm-hostd/src/supervisor/network/`. These were stages on the L3 packet path, which went away with the guest NIC, so the layer has been inert since that cutover while its own prose still says "always-on":

> `PlaceholderLeakScan` — always: drops any egress carrying a substitution placeholder out the raw wire. Always-on because the placeholder namespace is host-reserved and the scan is a cheap single-prefix check.

(`stages.rs:704-707`.) Nothing calls it.

**The gap that matters.** `prepare_request` substitutes and checks **headers** only. `PlaceholderLeakScan` was the backstop for a placeholder appearing anywhere else. So a workload that puts its placeholder in a JSON body rather than an `Authorization` header sends the placeholder itself to the destination.

That is not a credential leak — the placeholder is not the secret and the real value never leaves the host — but it is a silent failure of the substitution contract: the request goes out authenticated to nobody, the destination receives an opaque host-reserved token, and nothing in the audit chain records that a placeholder left unsubstituted.

**Two things should happen:**

1. **Decide the contract.** Either bodies are in scope for substitution — `prepare_request` scans and substitutes them under the same binding check headers get — or they are explicitly out of scope, and a placeholder found in a body is refused and audited rather than forwarded. "Silently forwarded" is the one option that should not survive.
2. **Delete or wire the stage layer.** An uncalled security stage is worse than none, because it reads as coverage — the same failure as the flowmux module deleted in #3283's first task, which a gate was "checking" while the compiler never saw it. If the decision in (1) is "refuse", the typed HTTP path is the natural home, and the L3-shaped stages should go. Sweep the doc comments in `crates/mvm-cli/src/commands/shared/resolve.rs:259` and `crates/mvm-contract/src/policy/projection.rs:262,268`, which describe `L4PolicyScan` and `DnsSinkholeScan` as live enforcement.

[...]

PR #3336, opened 2026-09-16
Terminate a CONNECT tunnel to a bound host, and leave the guest one egress path

### LINKED ISSUE #3283: An off-the-shelf HTTPS client cannot use credential substitution: the proxy refuses CONNECT

### LINKED ISSUE #3288: Two guest egress proxies: a CONNECT relay on 1080 and a substituting forward proxy on 18080
A workload guest has two egress entry points with different capabilities:

- `mvm-egress-client` on `127.0.0.1:1080` relays bytes for `CONNECT` and SOCKS5
  and never sees plaintext, so it cannot substitute a credential.
- `mvm-forward-proxy` on port 18080 (`crates/mvm-agentd/src/forward_proxy.rs`)
  parses absolute-form requests and can substitute, but refuses `CONNECT`.

Which one a process uses depends on which proxy variable it reads, and the
invoke path points `HTTP(S)_PROXY` at 18080 only when placeholders exist
(`crates/mvm-cli/src/commands/vm/invoke.rs:1339-1350`). The result is two data
paths with different policy behavior for the same destination.

**Do:** one guest egress entry point that handles `CONNECT`, SOCKS5 and
absolute-form, and hands every flow to the same host endpoint. Terminating TLS
there for bound hosts (see the CONNECT substitution issue) is what makes a
single entry point possible.

Plan: `specs/plans/2026-09-15-agent-sandbox-drive-plane.md`. Epic #3275.

### LINKED ISSUE #3305: check-single-network-path validates an orphan file the compiler never compiles
`crates/mvm-hostd/src/supervisor/flowmux/session.rs` is 1,188 lines containing a second full copy of the connect path — `gate.decide_target` at `:384`, DNS verdicts `:528-545`, ingress verdicts `:695-700`.

**It is not in the module tree.** `crates/mvm-hostd/src/supervisor/flowmux.rs:10-17` declares `http_flow, ingress, registry, resources, tcp_relay, udp_relay, wire` — no `mod session;` — and there is no `#[path]` attribute anywhere under `crates/mvm-hostd/src/`. The compiler never sees the file.

**And `check_single_network_path` enforces invariants on it.** `xtask/src/check_single_network_path.rs:256` lists it in `PEER_BRANCH_CALLERS` and `:346` in `FLOW_AUDIT_SITES`. Both read the file with `std::fs::read_to_string` (`:129`), so the gate passes by reading dead text.

This is exactly the failure mode #3297 names — a gate that proves a *witness exists* without proving anything *calls* it — still live, inside the gate that exists to prevent a second network path.

### Fix

1. Confirm non-compilation with a one-line build check (the evidence here is the absent `mod` declaration and absent `#[path]`; I did not build).
2. Delete `flowmux/session.rs` and its two `check_single_network_path` entries (`:256`, `:346`).
3. **Add an assertion to `check_single_network_path` that every path it names is reachable from a `mod` declaration.** Without this the same class recurs: the gate's whole method is reading files as text, so any file it names can silently leave the crate.

Step 3 is the part that matters. Deleting one orphan fixes one instance; asserting module-tree reachability fixes the class.

No claim is weakened — the file enforces nothing today. It removes false assurance.

Found during the big-cleanup section-A inventory (`specs/plans/2026-09-15-the-big-cleanup.md`).

auser:
Credential substitution shipped, worked, and could not be used by any HTTP client anyone would actually run. This makes it usable, with one egress path instead of two.

## Issues

- Closes #3283 — an off-the-shelf HTTPS client cannot use substitution because the in-guest proxy refuses `CONNECT`
- Closes #3288 — two guest egress proxies with different capabilities
- Closes #3305 — `check-single-network-path` validates an orphan file the compiler never compiles

Advances epic #3275. Related, still open: #3284, #3285, #3286, #3287, #3297, #3300, #3301, #3302.

## The problem

When a workload had secret placeholders, `HTTP(S)_PROXY` pointed at an in-guest proxy that accepts absolute-form requests only. Every HTTPS client that honours a proxy environment variable — curl, Python `requests`, most Node agents — tunnels with `CONNECT` instead, and got a 502. On the host side the refusal was correct: relaying an opaque flow to a host carrying a bound secret would put a placeholder on the wire and skip substitution entirely. So whether a request was substituted depended on which proxy the workload happened to pick, and the working combination was one no shipped agent CLI produces.

## Why the terminator was sitting there unwired

Not a decision. It was built against a topology with a guest NIC and an nftables redirect, and its live glue recovers the original destination from `SO_ORIGINAL_DST`. The vsock-only cutover deleted the NIC and the redirect, so the listener that fed it has been disabled at every call site since.

The TLS half survived that intact — it is generic over `Read + Write` and does not care what carries the bytes. The flow half did not, and is not reused here: it is a parallel pipeline that never consults the egress gate, never meters the AI budget, never runs reversible replacement, and buffers whole responses. A terminated flow is driven through `SubstitutionService::process_body_stream` instead, which is where those four controls already live.

## What holds the boundary

- The forward leg's destination is the **gate-admitted** value, never bytes from the guest.
- A request whose `Host` disagrees with the tunnel authority is refused with 421 and a chain-signed entry — and so is a request with **no** `Host`, because defaulting to the authority would make the absence of the check indistinguishable from it passing.
- Only an `Egress` route is terminable. For a peer route the gate's admitted address *is* the translation, so terminating one would re-originate by name through the ordinary resolver with the real credential attached. The guard is an allowlist match, so a future third route variant is not terminable by default.
- Bytes past the declared `Content-Length` are refused rather than folded into a credentialed body that never passed header substitution; a transfer-coded body is refused rather than mangled.
- Exhausting the per-VM terminated-flow ceiling **refuses**. It never degrades to an opaque relay — that degradation would have been the one path where this became a bypass.
- A tunnel to a destination no secret is bound to is relayed untouched. Mediating all guest egress was rejected in ADR-023 and stays rejected.

## The certificate

Self-signed per VM rather than an intermediate under a long-lived host root: an intermediate handed to a guest as its trust anchor needs partial-chain verification, which rustls supports and older OpenSSL builds do not — and the premise is that an unmodified client works. A review measured that both verifiers that matter enforce name constraints from a self-signed anchor. Deleting the host root lost nothing: the guest only ever trusted the intermediate directly, there was nothing revocable, per-boot minting is stronger rotation than a root provides, and it removed the only long-lived unconstrained signing key on the host.

Delivery and termination are one value, so an endpoint cannot hold TLS material the guest never received. A warm-claimed child inherits its parent's CA — and a child whose plan binds a destination that certificate does not permit is refused at claim time rather than minting a leaf the guest's TLS stack rejects with nothing in the audit log to say why.

The key never enters the guest by any channel: it rides the endpoint's stdin, is persisted 0600 through a create-fresh-then-rename helper, is zeroized, is unserializable by construction, and is forgotten on the stop path.

## One entry point

The load-bearing half of retiring the second proxy was not the environment variables. The `CONNECT`-capable client was gated on a token that was deliberately off for exactly the launches that bind a credential — a leftover from the pre-FlowMux single-stream protocol, when the two could not share the vsock port. They can.

## A module the compiler never saw

`flowmux/session.rs` held 1,188 lines that no `mod` declared: a byte-identical copy of the live `handle_open_tcp`, which a later change then edited, fixing nothing. `check-single-network-path` named it as one of two callers that must route a guest target through the egress gate. A gate asserting a property of a file the compiler does not compile reads as coverage, which is worse than no check.

When the remaining file went over the size ceiling, the split took the same care: the gate's two assertions were separated rather than moved together, so the positive one follows `decide_target` to the new file and the negative one now covers the whole `flowmux/` directory instead of one file.

## Verification

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo nextest run --workspace` (13,314 passing), `cargo test --workspace --doc`, `just check-gated`, and `cargo run -p xtask -- check-all` — 67 gates clean.

Three adversarial security reviews across the two halves. Findings fixed rather than argued: a peer route that would have re-originated with the real credential attached, an unaudited authority-mismatch refusal, pipelined bytes folding into a credentialed body, a warm-claimed child terminating under a certificate that does not permit its bound hosts, a persisted key inheriting a predecessor's file mode, and a delivery failure that would have broken all TLS rather than bound-host TLS.


auser:
Superseded by #3338: identical tree, same commits, with commit-message trailers removed so they do not land on main through the squash. Nothing else changed.

Issue #3371, opened 2026-09-16
Released Linux binaries require glibc 2.39 and do not run on RHEL/Rocky 9


auser:
The released Linux binaries require glibc 2.39, so they do not run on RHEL/Rocky 9 (glibc 2.34) or any other distribution older than roughly 2024.

`mvmctl` and `mvm-host-agent` in v0.17.0 and v0.18.0-rc.1 both link symbols versioned `GLIBC_2.39`: they are `-unknown-linux-gnu` builds produced on a recent runner image, so they inherit that image's glibc as a floor. Found by the distro compat lane (#3273), whose Rocky 9 cells fail with `GLIBC_2.39 not found`.

This is a product decision more than a bug fix, so the options:

1. **Ship `-unknown-linux-musl` builds** for the host binaries. No glibc floor at all; the builder-VM binaries already build musl via zig. Cost: verify nothing on the host path depends on glibc-only behaviour (DNS resolution through NSS, `dlopen`), and measure binary size and performance.
2. **Keep gnu, pin an old floor**, e.g. build with `cargo-zigbuild --target x86_64-unknown-linux-gnu.2.28`. Keeps glibc behaviour, runs on RHEL 8+.
3. **Keep the floor, and document it** — say which distributions are supported and have the installer refuse with a clear message on an older glibc rather than failing at the loader.

Recommendation: option 1 for the shipped host binaries, consistent with producing musl for everything mvm builds; option 3's installer check regardless, so an unsupported host gets a sentence rather than a loader error.

Issue #3370, opened 2026-09-16
The one-line install fails on Apple Silicon: releases through v0.17.0 keep entitlement profiles under resources/


auser:
The one-line install fails on Apple Silicon today.

`curl -fsSL https://runmvm.com/install.sh | sh` resolves the installer's baked default, **v0.17.0**. That release's macOS archive — like every earlier mvmctl release — ships the entitlement profile at `resources/mvmctl.entitlements`. `install.sh` requires `assets/mvmctl.entitlements` and stops with "missing entitlement profile" before installing anything. The installer before #3346 had the same requirement, so this is not a regression from that change; it has been true since the entitlement lookup moved to `assets/`.

Found by the installer back-compat lane (#3273), which installs the newest published releases with the current script.

**Do:** have the installer look for each required profile under `assets/` and then `resources/`, preferring `assets/`; install whichever it found into the release directory's `assets/` so the runtime lookup is unchanged. A test in `tests/install_sh.rs` that installs a fake macOS release carrying only `resources/mvmctl.entitlements` and `resources/mvm-supervisor.entitlements`. Moving the baked default to a release that ships `assets/` is not a substitute — pinning an older release must keep working.

Issue #3277, opened 2026-09-15
Epic: install lifecycle — monitored URL, atomic upgrade, and a way out


auser:
Installing, upgrading and removing `mvmctl` should be first-class, tested
and observable. The integrity posture — mandatory checksum verification and
tag-pinned keyless signature verification — is unchanged or stronger throughout,
never traded for convenience.

We are already ahead on integrity. What is missing is everything around the
download: a stable monitored URL, a baked version so the happy path makes no API
call, versioned install directories with an atomic swap, a real uninstaller, and
CI that runs the installer against older releases and across distros.

Plan: `specs/plans/2026-09-15-install-lifecycle-and-packaging-polish.md`

## Work items

- [ ] #3268 — Install: a stable vanity URL, monitored
- [ ] #3269 — Install: bake the default version instead of calling the GitHub API
- [ ] #3270 — Install: atomic upgrade with rollback
- [ ] #3271 — Install: there is no uninstaller
- [ ] #3272 — Verify the release signature without a host cosign, and close the claim 20 limits note
- [ ] #3273 — CI: installer back-compat and Linux distro compat lanes
- [ ] #3274 — Nix hygiene: version from the manifest, lockfile vendoring, and a stated test boundary

PR #3372, opened 2026-09-16
Run the current installer against published releases and across distros

### LINKED ISSUE #3273: CI: installer back-compat and Linux distro compat lanes
`tests/install_sh.rs` exercises the current script against synthetic assets
only. Nothing runs the current installer against older *published* releases, and
nothing runs the released Linux binary across distros — though we ship
`-unknown-linux-gnu` and glibc drift is unmeasured.

**Do:**
- A workflow running the current installer against the last N published
  releases, on a macOS and a Linux runner.
- A workflow running the released Linux binary in debian, ubuntu, rocky and
  fedora containers. This also quantifies what staying on gnu costs us versus
  musl.
- A cold first-run smoke on Linux executing the exact commands from
  `public/src/content/docs/install/linux.md`, triggered by edits to that page or
  to `install.sh`. The macOS equivalent can only cover install plus `doctor`
  until a self-hosted Apple Silicon runner exists (#3011).

Plan: `specs/plans/2026-09-15-install-lifecycle-and-packaging-polish.md` WS6.

auser:
## Issues

- Closes #3273 — installer back-compat and Linux distro compat lanes

Part of epic #3277. Surfaced #3370 (the one-line install fails on Apple Silicon) and #3371 (released binaries need glibc 2.39).

## Why

Nothing ran `install.sh` against a release anyone could download. `tests/install_sh.rs` drives the script against synthetic archives — which tests it against the archive shape we expect, not the shapes we shipped.

## `.github/workflows/installer-compat.yml`

- **Resolve** — picks the newest three published `v*` mvmctl releases from the releases API each run, skipping `boot-image/*` tags, tags with no assets, and v0.0.1's differently named archives. A release older than the installer's default may be refused; anything else must install.
- **Installer** (`ubuntu-latest`, `macos-latest`) — installs each release into a prefix already holding unrelated files; checks `--version` and that every executable and `assets/` in *that release's own archive* is linked (payloads `install.sh` excludes must not be); on macOS checks entitlement keys and that `doctor` runs; uninstalls with state present — published releases predate the running-machine check, so it must refuse without removing anything, then `--force` removes it — and checks only the unrelated files remain.
- **Upgrade** — walks one prefix oldest → newest; after each step the replaced release must still be complete and runnable and entries the newer release dropped must be gone; then rolls back and uninstalls.
- **Distro** — `debian:stable`, `ubuntu:24.04`, `rockylinux/rockylinux:9`, `fedora:latest` on x86_64 and aarch64 (`ubuntu-24.04-arm`, as `ci.yml` already uses). Every binary goes through `ldd -v` first; the job summary shows host glibc, required glibc and each `GLIBC_x.y not found`; a loader error fails the job.
- **Docs smoke** — `e2e-docs.yml` did not cover the install page. This extracts the One-liner, Pin a version and Verify blocks from `install/linux.md` and runs them as written; on a PR only the installer URL points at the PR's `install.sh`.

**Triggers:** weekly (Monday 10:37 UTC — what changes without a PR is external: new releases and moving container tags; the slot avoids the nightly macOS cluster), `workflow_dispatch`, and PRs touching `install.sh`, `uninstall.sh`, the install pages, the scripts or the workflow. A release does not trigger it, since releases publish with the default token; run it by hand after cutting one.

**Not a required check** — every lane depends on GitHub downloads and public registries. The scripts' own tests run on every PR in `ci.yml`, with a shellcheck entry. A failed scheduled run opens or updates one issue and a green run closes it, like `install-url-watch.yml`.

## Found building it

1. **The documented version pin never pinned.** `MVM_VERSION=v0.16.1 curl … | sh` gives the variable to `curl`, not `sh`, so it silently installed the default. Fixed in `install/linux.md`, `install/macos.md`, `getting-started/installation.md` and `public/src/lib/agent-skill.ts`.
2. **#3370** — the installer cannot install its own default release on Apple Silicon: releases through v0.17.0 keep the entitlement profile under `resources/`, and the installer requires `assets/`. The macOS cells for those releases will be red until it is fixed.
3. **#3371** — the released Linux binaries need glibc 2.39, so the Rocky 9 cells fail. That is what the lane is for.

## Verification

Locally: `scripts/installer-compat/installer-compat.test.sh` — 36 cases through the real `install.sh`/`uninstall.sh` against fake releases served locally, including must-fail cases checked to fail for the right reason; the release picker and archive reader also run against the real releases API and the real v0.16.1 / v0.17.0 / v0.18.0-rc.1 archives; `shellcheck` and `actionlint` clean; installer, workflow-structure and release test binaries 154/155 (the one failure a timing test of an unrelated script, passing in isolation); `check-all` (68 gates).

Only CI's first run proves the container runs and real distro `ldd -v` output, signing and `doctor` on hosted macOS, the issue-reporting job, and two Linux-only test cases that need Docker.


auser:
First CI run, read cell by cell:

- **Lint (required)** failed on three shellcheck SC2015 notes (`A && B || C`) in `scripts/installer-compat/`. CI's shellcheck flags them at default severity; the local version did not. Fixed in `492a34d9bb` by writing them as explicit `if`s.
- **`Install v0.17.0 (macos-latest)`** was the #3370 bug, the installer failing on its own default release. That is fixed on main by #3389, and this PR's CI runs on the merge with main, so the cell should now pass.
- **`Install v0.18.0-rc.1 (macos-latest)` and `Upgrade and roll back (macos-latest)`** are a defect in that **published artifact**. rc.1's macOS `mvmctl` is dynamically linked against Homebrew's libkrunfw:
  ```
  dyld: Library not loaded: /opt/homebrew/opt/libkrunfw/lib/libkrunfw.5.dylib
    Referenced from: …/lib/mvm/1-v0.18.0-rc.1/mvmctl
  ```
  So it cannot run `--version` on a Mac without libkrun installed. The source is already fixed: #3247 removed libkrun from standard builds and releases on 2026-09-15, after rc.1 was cut on 2026-09-09. The next release candidate clears these cells. The installer handled it correctly: it ran the staged `mvmctl` before switching, refused, and left the previous install untouched.
- **`rockylinux/rockylinux:9` (x86_64 and aarch64)** are #3371, the released binaries needing glibc 2.39. That is what this lane exists to show.

None of the compat lanes are required checks, so the known-red cells don't block the merge.

Issue #3373, opened 2026-09-16
repo: measure and gate post-cutover history compaction


auser:
Tracks optional W9 of specs/plans/2026-09-16-image-repository-extraction.md.

After the image repository cutover and compatibility window, measure whether generated image/build blobs in mvm history make a destructive history rewrite worthwhile. Moving paths does not shrink existing history. Do not combine this work with the cutover.

Acceptance:
- Inventory only blobs reachable from the remote default branch and published tags; measure fresh full and blobless clones.
- Record a thresholded go/no-go decision and prefer partial/blobless clone documentation when the savings do not justify a rewrite.
- Any rewrite requires separate owner approval, a merge freeze, a verified permanent read-only archive/object bundle, an old-to-new commit map, and contributor/fork recovery instructions.
- Review an exact git-filter-repo path policy; preserve GitHub release assets and historical release provenance.
- Coordinate branches, tags, signatures, attestations, protected settings, and required fresh clones.
- Verify repository integrity, release-source mapping, build/test gates, and before/after clone size before reopening merges.

This issue authorizes measurement and planning only. It does not authorize force-pushing rewritten history.

Issue #3369, opened 2026-09-16
supply-chain: migrate image publication and consumer trust


auser:
Tracks W6 of specs/plans/2026-09-16-image-repository-extraction.md.

Publish a complete candidate image set from mvm-images and migrate mvm consumers through an explicit old-plus-new trust window. Cover verifier identities, revocation URLs, Stage 0 kernels, default images, image updates, CI, and WebLinux consumers.

Acceptance:
- A clean machine downloads only from mvm-images, verifies the pinned identity and digest, and boots.
- Linux Firecracker and physical Apple Silicon HVF witnesses pass.
- A pin-update workflow opens evidence-bearing PRs but never auto-merges.
- Old trust is removed only after the compatibility window.

Issue #3362, opened 2026-09-16
images: move and reproduce image sources in mvm-images


auser:
Tracks W4 of specs/plans/2026-09-16-image-repository-extraction.md.

Inventory and move the image-owned Nix flakes, locks, assembly scripts, and image tests into mvm-images. Build embedded guest and builder binaries from an explicit public mvm source commit without copying the product tree.

Acceptance:
- x86_64 and aarch64 image sets build and boot.
- New and existing releases are compared for file set, manifest semantics, closure, boot behavior, and explained byte differences.
- Shared Nix helpers are moved only when all consumers are image concerns.
- Every published manifest records the exact mvm input commit.

auser:
Conflict warning for W4, so the move is sequenced with it rather than colliding.

Another session has an **unpushed** branch, `feat/firecracker-builder-image`, closing #3324, and says it will open the PR shortly. It touches the Stage 0 and builder-runner code in `mvm-build` / `mvm-runtime`:

- a Firecracker builder-image resolver (`DriverBuilderVm<FcDriver>` boots the Stage 0 image as-is);
- three builder-spec fixes: egress port `HostDials` → `GuestDials`, console taken from the driver, and no agent port on one-shot builders;
- `ConsoleHaltWatch` in `mvm-runtime`'s `builder_runner`, so a builder guest that **halts instead of powering off** no longer wedges the run — Firecracker does not exit on a guest halt, and a live run hung ~27 min at `Power off not available: System halted instead`;
- `FcRunningVm::kill` able to terminate an agentless guest.

It was validated live on x86_64 Firecracker v1.14.1 with a cold `MVM_HOME`: auto-detect picked Firecracker, Stage 0 exited 0 in ~17 min, and a builder shell job exited 0 including a fetch from `cache.nixos.org` over egress.

Any W4 move of the Stage 0 or builder-runner code should land after that PR, or rebase onto it. Relatedly, the plan should state where the Stage 0 *driver* lives after extraction: the plan keeps "explicit source-build integration used by contributors" in `mvm`, which suggests it stays, and #3360's Stage 0 store fixes depend on that answer.

auser:
The Firecracker builder work this issue's W4 sequences after is now open as #3376 (closes #3324). It changes host-side code only: the builder runner, the Firecracker driver's kill path, and the builder spec. It moves no image sources, so W4 can start after it merges.

auser:
[...]

Issue #3365, opened 2026-09-16
artifact: define signed image-set manifest and mvm lock


auser:
Tracks W3 of specs/plans/2026-09-16-image-repository-extraction.md.

Define one versioned, signed, atomic image-set manifest and one checked-in mvm image lock. Include producer and mvm source commits, guest architectures, artifact roles, boot protocols, formats, required capabilities, digests, sizes, protocol compatibility, Nix inputs, SBOM references, signing identity, and revocation channel. Keep guest artifact selection independent of host OS so Firecracker, HVF, and a future Windows backend can select compatible packs without host-named images.

Acceptance:
- Round-trip and default tests cover all new types.
- Tampering, wrong identity, wrong architecture, partial sets, incompatible protocols/capabilities, replay, and revocation are rejected.
- Offline verification works from the manifest, signature bundle, and artifacts.
- Scattered repository and tag pins can be generated or replaced from one lock.
- Windows remains schema-ready rather than claimed as supported until a backend and native Windows witness exist.

auser:
Host↔guest contracts that cross the `mvm` ↔ `mvm-images` boundary as of today, from recently merged work. These should be pinned by the image-set manifest / compatibility contract rather than treated as files that move.

1. **Per-VM egress CA delivery (#3338).** The host writes the per-VM CA certificate as an optional `egress-ca.crt` node on the per-boot FlowMux identity drive (`crates/mvm-vmm/src/host/flowmux_identity.rs`, `IdentityDriveContents`; the writer refuses a slot containing private key material). The guest side:
   - `nix/lib/mk-guest.nix`, Stage 2.485, runs **after** `provision-identity-for` and calls `mvm-egress-client install-egress-ca-trust`, which builds `/run/mvm/ca-bundle.crt` from the image's baked roots plus that certificate — always writing a bundle, roots alone when no certificate arrived;
   - the host's invoke path exports `SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE` and `NODE_EXTRA_CA_CERTS` pointing at `/run/mvm/ca-bundle.crt`.

   The drive node name, the verb, the bundle path and the ordering are one contract across both repos. The `mvm.egress_ca` kernel-cmdline token was deleted on both sides; an image that still parses it is stale.

[...]

PR #3375, opened 2026-09-16
docs: plan image repository extraction

auser:
## Summary

- define the incremental extraction of the builder, workload, overlay, SDK, and smoke image train into one public mvm-images repository
- preserve signed, digest-pinned production consumption while adding an explicit dev-only sibling-checkout workflow
- make the artifact contract portable across Linux/Firecracker, macOS/HVF, and a future Windows backend
- isolate paired worktrees and document current migration overlaps
- add a separately gated post-cutover history-compaction decision
- register the work in the sprint and refactor rollups

## Workstreams

Issues #3362 through #3369 track W1-W8. Issue #3373 tracks the optional history measurement and go/no-go decision.

## Validation

- cargo run -p xtask -- check-plan-names
- cargo run -p xtask -- check-declared-backing
- cargo run -p xtask -- check-sprint-append
- git diff --check


auser:
One line to add under "`mvm` continues to own": **the Stage 0 driver and the builder runner** (`mvm-build`'s Stage 0 orchestration, `mvm-runtime`'s `builder_runner`, the VMM drivers). The plan moves the Stage 0 *seed inputs*, but it does not say where the code that boots Stage 0 lives, and two things depend on the answer: #3360's Stage 0 store GC, store drop and virtio-blk discard fixes, and #3376 (Firecracker builder + halted-guest fix), which W4 sequences after. Keeping it in `mvm` matches "explicit source-build integration used by contributors" above.

auser:
Boundary note for W3 (#3365): the host↔guest contracts that recent merged work put across the mvm ↔ mvm-images line — the per-VM egress CA on the identity drive and its trust-bundle verb, the single guest egress entry point, the `importTOML` version reads, and `check-deferrals` coverage of `nix/` — are listed in https://github.com/tinylabscom/mvm/issues/3365#issuecomment-5706360083, along with three upcoming changes (#3355, #3371, #3265) that will need coordinating across both repos.

auser:
Four points from a read-only comparison of image-build models against `origin/main`, which bear on what moves to `mvm-images`:

1. **The runtime overlay, initramfs and SDK sidecar may not belong in `mvm-images`.**
   - The overlay resolver requires `VERSION == mvmctl version` exactly (`crates/mvm-fs/src/overlay.rs:360-365`). That contradicts the plan's compatibility-range model (trust invariant 6 / W3), so moving it would mean a cross-repo image release for every guest-agent change.
   - A source checkout already builds the overlay on the host with no Nix and no VM (`crates/mvm-build/src/runtime_overlay.rs:643-656`).
   - The universal initramfs is already a cargo-built, epoch-zero cpio (`crates/mvm-build/src/initramfs.rs:1-13`).

   These are host-assembled and version-locked to `mvmctl`, which argues for keeping them in `mvm`.
2. **Inventory mismatch.** `nix/images/sdk-sidecar/` (plan ~line 129) does not exist; the sidecar is built by a builder shell job (`crates/mvm-cli/src/commands/env/builder_vm/sdk_sidecar.rs:111-123`). `nix/images/kernel/` and `nix/images/initramfs/` are not listed at all.
3. **`nix/lib/mk-guest.nix` is public API, not an image helper.** User flakes consume it (`examples/*/flake.nix`, `public/src/content/docs/guides/nix-flakes.md`), and `default-tenant` would consume it from the other repository. W4 should version it as an interface.
4. **The contributor-source rule needs restating.** Once image sources leave the checkout, CLAUDE.md's "when matching source is available" effectively means "with an explicit sibling checkout" (plan ~lines 30-33, 235).

Separately, for W1's measurement: `--prod` OCI runs currently boot the builder VM on macOS only to reseal verity with `veritysetup` (`crates/mvm-build/src/run_image.rs:152-153, 254-316`), although pure verity sidecars are already written a few lines earlier (`:144-147`) and the guest accepts both layouts. Removing that reseal will cut builder boots from `--prod` scenarios independently of the extraction. Measure it separately, so W1's gain isn't overstated or double-counted.

auser:
Coordination: I am about to fix #3399 — the in-guest DNS stub in `crates/mvm-agentd/src/flowmux_egress.rs` indexes `query[0], query[1]` with no length check, so a 0- or 1-byte datagram to `127.0.0.1:53` panics it (UDP and TCP paths). The change is confined to that file in `mvm-agentd` (length check before indexing, plus tests); it changes no image recipe, drive layout, verb or path. Once it lands, the next guest image built from `mvm` picks it up — flagging so the extraction can account for which `mvm-agentd` revision a pinned image set carries.

Separately, #3406 (refuse raw UDP/53 under an allow-list, #3355) turned out to be host-only: guest resolution already goes loopback stub → FlowMux `Resolve` → `EgressGate::dns_verdict`, so no guest-image change.

Issue #3433, opened 2026-09-18
Firecracker-served builder and Stage 0 stores cannot be trimmed: block discard needs Firecracker 1.17.0, we pin v1.14.1


auser:
Split out of #3360.

#3360 part 2 adds `VIRTIO_BLK_F_DISCARD` to the in-house VMM device model, so a guest `fstrim` (or the Stage 0 `FITRIM`) now returns space to the host — **on HVF**. The Linux side does not benefit, and the reason is not in our device model.

Firecracker serves the Linux builder and Stage 0 disks with its own virtio-block, which gained opt-in discard only in **1.17.0** (writable drives, Sync IO engine). `FC_VERSION_DEFAULT` in `crates/mvm-core/src/config.rs` pins **v1.14.1**. Below that version the device never advertises the feature, so the guest block layer sets `discard_max_bytes` to 0 and every discard the filesystem generates is dropped before it reaches the host. Both trims return `EOPNOTSUPP` and the sparse images keep their high-water mark.

## What closing this needs

- Bump `FC_VERSION_DEFAULT` past 1.17.0, with whatever else that bump drags in (the version is also pinned in `nix/ops/hetzner/cloud-init.yaml`, which the comments there say to bump first).
- Opt the builder and Stage 0 drives into discard in the Firecracker drive config, and confirm the IO engine they use is the one that supports it.
- Prove it live on the KVM box: allocated blocks of the store image before and after a collect + trim, not just the absence of an error.

## Why it is separate

It is a dependency bump plus a drive-config change with its own live-validation cost, and it does not touch the device model. Folding it into the discard PR would have put a Firecracker upgrade inside a virtio change.

## Do not claim

Until this lands, builder-store reclaim is HVF-only. libkrun's block device was never surveyed for discard — unknown, not "no".

Issue #3360, opened 2026-09-16
Builder Nix stores only grow: Stage 0 is never GC'd, virtio-blk has no discard, and 53 GB has no targeted reclaim


auser:
On a contributor Mac, `~/.mvm/cache/builder-vm/` holds **~115 GB**, and two sparse Nix-store images account for 85 GB of it:

| Image | Real on disk | Apparent |
|---|---|---|
| `nix-store-stage0-aarch64.img` | **53 GB** | 64 GB |
| `nix-store-aarch64.img` (steady-state builder) | **32 GB** | 64 GB |
| `builder-vm/hvf` | 27 GB | — |
| `builder-vm/jobs` | 1.8 GB | — |

Nothing in the design stops these growing, and nothing gives the space back. Three causes compound.

## 1. The Stage 0 store is never garbage-collected

The builder GC cap — `DEFAULT_BUILDER_STORE_GC_GIB = 24` (`crates/mvm-build/src/builder_vm_runtime.rs:124`), which runs `nix-collect-garbage --delete-older-than 14d` past the cap — is applied in exactly one place: `render_flake_cmd_sh` (`builder_vm_runtime.rs:420`, cap at `:446`, script at `:693`), the **steady-state** builder's flake build script.

Stage 0 never runs it. So every Stage 0 run's intermediates stay in `nix-store-stage0-<arch>.img` until the image reaches its 64 GiB ceiling (`DEFAULT_NIX_STORE_MIB = 65536`, `crates/mvm-build/src/libkrun_builder.rs`). The 53 GB image above is consistent with that: well past the 24 GiB cap the steady-state store gets.

## 2. Freed guest space never returns to the host

The in-house virtio-blk device (`crates/mvm-vmm/src/vmm/virtio.rs:75-81, :300-301`) advertises only `VIRTIO_BLK_F_RO` or `VIRTIO_BLK_F_FLUSH`. There is no `VIRTIO_BLK_F_DISCARD` or `VIRTIO_BLK_F_WRITE_ZEROES`, no `discard` mount option, no `fstrim`, and no host hole-punching (`F_PUNCHHOLE` / `FALLOC_FL_PUNCH_HOLE`) anywhere in the builder path.

[...]

PR #3434, opened 2026-09-18
fix(builder): bound the Stage 0 Nix store and let cache prune reclaim it

auser:
Parts 1 and 3 of #3360: garbage-collect the Stage 0 Nix store, and let `cache prune` remove it on its own. Part 2, the virtio-blk discard that lets a trim return host disk, landed in #3440.

The Stage 0 store (`nix-store-stage0-<arch>.img`) was never collected, so every bootstrap's intermediates stayed until the image reached its 64 GiB ceiling. A long-lived contributor home measured 53 GB of it, against a 12 GB clean closure. The only way to remove it was `cache repair`, which also deletes the builder image Stage 0 produced.

## Collection

After a successful build, `stage0-init` measures the store filesystem. Past the cap it collects:

1. **Registers the seed.** The seed's own paths (the `nix` that runs Stage 0 and its CA bundle) were never loaded into the Nix database. Nix ignores a root that points at an unregistered path and deletes unregistered paths as garbage, so an unguarded `nix store gc` would take the seed. The seed's `.reginfo` is copied to `/run` before the persistent store is bound over `/nix`, then loaded with `nix-store --load-db`.
2. **Roots** the seed `nix` and CA bundle, plus the last output of each Stage 0 output mode, so an unchanged next bootstrap is still a cache hit and building the kernel does not unroot the image.
3. Runs `nix store gc`.
4. **Trims** the store with `FITRIM`. A disk without discard answers `EOPNOTSUPP`, which is logged and ignored.
5. If the seed is gone afterwards anyway, drops the store marker so the next run reseeds instead of booting a store with no `nix`.

Collection never fails Stage 0: the artifacts are already on the output disk. Every decision logs one line, including why it did not collect (no persistent store, within cap, no seed registration).

**The cap** is the steady-state builder's (`MVM_BUILDER_STORE_GC_GIB`, default 24 GiB). It reaches the guest as `mvm.store_gc_kib=<KiB>` on the Stage 0 kernel cmdline, added by `stage0_spec` for every backend. It is not in `stage0-build.conf`, because the ordinary builder-image bootstrap writes no conf. A cmdline without the token gets the default, and a zero or unparseable value is ignored rather than collecting on every run.

## Console drain

`power_off()` drains stdout and stderr before `reboot(2)`. The halt was truncating the `stage0-init: done; halting` marker the host looks for.

## `cache prune --stage0-store`

Removes only `nix-store-stage0-<arch>.img`, keeping the builder image and the steady-state store. `--deep` implies it. An explicit `--stage0-store` refuses the whole prune while a Stage 0 bootstrap holds its lock; `--deep` skips the store with a warning and reclaims everything else. The freed figure is allocated blocks, not the 64 GiB sparse length, and `cache repair --store-only` now reports the same way.

## Layout

The guest code is in `stage0-init/store_gc.rs`: the pure policy (cap parsing, roots, store-path resolution, the `FITRIM` encoding with a compile-time layout contract against `linux/fs.h`), plus a Linux-only `collect` module that does the work. The seed lookups moved unchanged to `stage0-init/seed.rs`. That keeps `stage0-init.rs` under the file-size cap.

## Live evidence

Hetzner KVM box, x86_64, Firecracker 1.14.1, cold `MVM_HOME`, `MVM_BUILDER_STORE_GC_GIB=1`. Built with `MVM_EMBED_NO_CACHE=1`, with the new log strings confirmed in the embedded init before any result was trusted.

- **Run 1:** `Nix store 9307936 KiB -> 2157480 KiB after garbage collection`, then `the store disk does not support discard` (expected on Firecracker 1.14.1). Done marker intact, exit 0.
- **Run 2** (builder image deleted, store kept): reused the store, built 0 derivations, collected again (2620968 -> 2156280 KiB), exit 0.
- **`cache prune --stage0-store`**, dry run then real: removed only the Stage 0 store, 11.2G freed.

Those runs used the tree before the final commit, which only moves code between modules and adds the layout assertions. I did not repeat the 35-minute live run for it.

## Not done

- **Firecracker does not shrink the file.** The trim returns host disk only where the store disk honours discard, which since #3440 is the in-house device model (HVF). Firecracker gained block discard in 1.17 and is pinned at 1.14.1 (#3433). On Linux, collection stops the growth but the image file keeps its high-water mark.
- **Default persistence is undecided.** The Stage 0 store still persists by default (now bounded); `--stage0-store` is the reclaim.

## Validation

`cargo fmt --all -- --check`; `RUSTFLAGS="-D warnings" just check-gated` (the only local build that compiles `stage0-init` for Linux); Linux-target clippy on `mvm-build`; workspace clippy `-D warnings`; `cargo nextest run -p mvm-build -p mvm-cli -p mvm-runtime -p mvm-backends` (4577/4578; the one failure, a vCPU CPU-time test in untouched HVF code, failed at load average ~280 and passes on rerun); `cargo test --workspace --doc`; `xtask check-all` (69 gates); the `--features test-support` library lane (4693 passed).


auser:
Marking draft: the two commits pushed here carry the Stage 0 GC cap in `stage0-build.conf`, and that mechanism does not reach the guest on the path that needs it — the ordinary builder-image bootstrap writes no config at all; only the SDK-sidecar and external-kernel paths do. So the cap would land on exactly the paths that did not need it.

The originating session had already caught this and reworked it onto the kernel cmdline, but the session ended before that was committed. Picking that up now, along with the console-settle fix it carried. Re-verifying and undrafting once both are in.

Issue #3477, opened 2026-09-19
spawn_scope unresponsive-manager test flakes: fake systemd-run launcher not found


auser:
`mvm-core` `spawn_scope::tests::an_unresponsive_manager_fails_the_launch_instead_of_hanging_it` failed on the aarch64 workspace lane of an unrelated PR (#3474, which touches only `mvm-hostd`). Run 35417815707, job 105830235277:

```
panicked at crates/mvm-core/src/spawn_scope.rs:1892:14:
a scope that is never created must fail the launch: Child { stdin: None, stdout: None, stderr: None, .. }
/bin/sh: 0: cannot open /tmp/.tmpP0tkzZ/bin/systemd-run: No such file
```

The test writes a never-returning fake `systemd-run` into a scratch directory (`pretend_mechanism_with_launcher`) and expects `bind_spawn(...).spawn()` to fail at the 300 ms creation deadline. Here `/bin/sh` was handed a launcher path that did not exist, so the child exited at once, `spawn` returned `Ok(Child)`, and `expect_err` panicked. The rest of the lane passed: 13,828 of 13,829.

The launcher path is resolved through the environment `TestEnv` sets up. So one likely cause is the path being resolved or written somewhere other than where it is executed from (for example, another test's environment, or a scratch dir dropped early). A second possibility is the file not being fully visible yet when `sh` opens it. It is intermittent: the same code passed on the x86_64 lane of the same run and on other PRs.

**Do:** make the fixture write and `chmod` the launcher, then verify it exists before `bind_spawn` resolves it. Pass its absolute path explicitly instead of relying on `PATH` from a process-wide environment. And assert on the refusal message, not only on `is_err`, so a launcher that never ran cannot satisfy or fail the test for the wrong reason.

auser:
Reopening: `an_unresponsive_manager_fails_the_launch_instead_of_hanging_it` still fails after #3486. For example, run 35452802258 on #3482, with the panic now at `spawn_scope.rs:1954`: `spawn` returned `Ok(Child)` where the test expects the timeout.

#3486 fixed a real race (a launcher resolved through a `PATH` that another test was changing), but this is a second one, and it is in `BoundCommand::spawn` itself rather than in the test.

[...]

PR #3496, opened 2026-09-19
feat(sdk): load the host library from Python, with generated error classes

auser:
Part of #3261: tasks 3 (error codes in the SDK taxonomy) and 4 (`MVM_HOSTLIB_PATH`), and the foundation of task 8 (the Python binding). #3261 stays open.

## Change

**One definition of the error codes.** The codes a programmatic caller branches on (`NOT_FOUND` … `UNAVAILABLE`, plus the host library's `INVALID_INPUT`, `ABI_NOT_NEGOTIATED`, `EMBEDDER`, `INTERNAL`) are now constants in `mvm_core::error_codes`. The module sits outside the `client` feature so `mvm-sdk` can name them. `MvmError::code` and the host library's error bodies use them.

**Generated error classes.** The SDK taxonomy gains a host-library family: a `HostLibraryError` base plus one type per code. The family is keyed by the `code` an error body carries, not by status number, because host-library statuses reuse the host-services numbers. `gen-stubs` emits a `CODE_ERRORS` map beside `STATUS_ERRORS`, only for a surface that has code-keyed types. So Python gets it now, and TypeScript will when its binding raises these. `gen-stubs` changed only `schema/sdk-errors-v0.json` and Python's `types.py`, additions only.

**The Python loader**, `mvm/_hostlib.py`:
- Finds the library through `MVM_HOSTLIB_PATH`, else beside `mvmctl` on `PATH`, else `MvmTransportError`. It locates `mvmctl` but never runs it, and follows a symlinked `mvmctl` to its real directory. An override naming a missing file is refused, not ignored.
- Negotiates the ABI once per process. A mismatch raises `HostLibraryAbiError` naming both versions.
- `call(method, request)` raises the type the body's `code` names, with `code`, `retryable` and `status` set, and the base type for an unknown code.
- A test asserts the module references no process API.

`_LiveTransport` is not switched over here. That needs the library's launch method, which waits on #3476 and the stack behind it.

## Validation

- `cargo nextest run -p mvm-core -p mvm-hostlib -p mvm-sdk -p xtask`: 3304/3304 passed. `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `xtask check-all` (including `check-stubs`), and `scripts/check-crate-readmes.sh`: pass.
- `pytest tests/test_hostlib.py`: 15 passed, including a live test against a built `libmvm_hostlib.dylib`. `machine.list` answered, and an unknown method raised `HostLibraryInputError`.
- The rest of the Python suite passes, apart from 8 `test_schema_derive.py` tests that need the optional `pydantic` package, which my local environment lacks. Those don't touch this code.
- New Rust tests:
  - every host-library code maps to exactly one type;
  - no type carries both a status and a code;
  - `CODE_ERRORS` is emitted only where it has entries;
  - a manifest without codes still parses.

This touches string literals in `mvm-hostlib/src/lib.rs`, which #3482's follow-up also edits. Whichever lands second rebases.


auser:
Closing and reopening for a fresh CI run: aarch64 failed on the spawn_scope flake (#3477, fixed by #3486 in the queue), unrelated to this change.

Issue #3272, opened 2026-09-15
Verify the release signature without a host cosign, and close the claim 20 limits note


auser:
`verify_signature` on the self-update path returns `Ok` with a warning when
`cosign` is absent (`crates/mvm-cli/src/update.rs`), and `install.sh:95-119` does
the same. So one of claim 20's three consuming paths is best-effort and the
SHA-256 pin is what holds — which is exactly what the "Claim 20 limits" note in
ADR-001 records.

**Do:** verify the Sigstore bundle in-process in Rust for `mvmctl update`; have
`install.sh` prefer `mvmctl verify-release` once a binary exists on disk, falling
back to `cosign` and then to the current warning; delete the limits note once the
third path refuses an unsigned artifact like the other two.

Weigh the closure cost first — a bundle verifier is a real dependency. If the
cost is unacceptable, record *that* decision in the ADR rather than leaving the
limit unexplained.

Plan: `specs/plans/2026-09-15-install-lifecycle-and-packaging-polish.md` WS5.

PR #3504, opened 2026-09-19
fix(cli): verify the release signature in-process on self-update

auser:
Addresses #3272 (the self-update half; `install.sh` stays open).

## What was wrong

`verify_signature` in `crates/mvm-cli/src/update.rs` shelled out to `cosign verify-blob` and returned `Ok` with a warning when cosign was not on `PATH`. On the one path that replaces the `mvmctl` binary, the publisher check depended on what the host had installed, and the SHA-256 it fell back on comes from a manifest fetched over the same channel as the archive.

## Fix

It now calls `mvm_build::release_signature::verify_release_archive_signature` with `ReleaseTrain::Cli`. That is the verifier the fetch path and the runtime overlay already use, against the same embedded trust root. A missing, unparseable, or foreign-signed bundle refuses the update. No dependency is added, because release builds already carry `manifest-verify`. A build without that feature refuses to self-update, and the refusal says why.

`--skip-verify` keeps its meaning (skip the signature, keep the SHA-256). Its help text used to say "Skip checksum verification", which was wrong.

## Claim 20

The statement now names the build, fetch, and self-update paths. This is updated in ADR-001 (both tables), `model/claims.toml`, the generated `CONFORMANCE.md`, and CLAUDE.md. The "Claim 20 limits" note is rewritten to keep the two things still outside the claim: a build without `manifest-verify`, and `install.sh`, which runs before any `mvmctl` exists.

The `manifest-verify`-gated tests on these paths never ran in CI, because every use of that feature there was `cargo run --example`. A new step in `Lint feature coverage` runs them, which is what makes the new witness citable.

## Evidence

- `an_archive_without_a_bundle_is_refused`
- `an_archive_with_a_garbage_bundle_is_refused`
- `a_real_release_bundle_verifies_under_its_tag` checks the committed `v0.18.0-rc.1` bundle. It verifies under `v0.18.0-rc.1` and is refused under `v0.18.0`, which shows the tag's `v` and the identity template's `v` are not doubled.

## Gates

- `cargo nextest run -p mvm-cli --features manifest-verify -E 'test(/^update::/)'`: 27 passed
- the same run without the feature: 26 passed
- `xtask check-all`: 69 gates clean
- `check-claim-catalog`, `check-conformance`, `check-witness-citations`, `check-asserted-absence`: clean
- mutation-witness baseline re-pinned, because `update.rs` joined the surface
- pre-commit hook (fmt, clippy, actionlint): pass


auser:
Closing and reopening for a fresh CI run: aarch64 failed on the known spawn_scope flake (#3477), unrelated to this change.

Issue #3364, opened 2026-09-16
dev: support explicit sibling-checkout image development


auser:
Tracks W5 of specs/plans/2026-09-16-image-repository-extraction.md.

Make paired mvm and mvm-images worktrees a first-class development workflow. Add an explicit local image source, shared manifest parser and resolver, content-addressed cache identity, trust-tier reporting, worktree-pair-scoped mutable state, and contributor documentation. Do not use automatic sibling discovery or a Git submodule.

Acceptance:
- A developer can change an image in a sibling worktree and boot it without publishing.
- A second run reuses the content-addressed result.
- Both repository commits and dirty fingerprints participate in cache identity and doctor output.
- Mutable build/runtime state and VM/TAP/socket identities are isolated per paired worktree; two paired changes can run concurrently.
- Only atomically published immutable cache entries are shared.
- Release binaries and production admission refuse unsigned local packs.
- Path traversal, symlink substitution, stale manifest, and wrong-architecture tests fail safely.

auser:
W5 is complete on main. Every W5a-W5n slice and the W5m acceptance witness is checked in specs/plans/2026-09-16-image-repository-extraction.md. PR #3618 landed the contributor-default sibling checkout and paired-CI witness, while PR #3623 closed the kernel-selection defect found by the live witness. The plan records cache reuse/invalidation, concurrent-pair isolation, stale/wrong-architecture refusals, production trust-tier refusal, and the live sibling boot evidence. Closing this tracking issue as completed.

PR #3511, opened 2026-09-19
Read a locally built image set at the local-dev tier (W5c, #3364)

auser:
Slice W5c of the image-repository extraction (#3364): mvm now reads a set built from a local `mvm-images` checkout paired with a local mvm checkout. It uses the same schema and parser as a released set, classifies the result `local-dev` (never `verified-release`), and refuses it when it no longer describes what is on disk. The emitter side is tinylabscom/mvm-images#6 (W5b). No image consumer reads a local set yet; that work is W5f–W5k.

> **Stacked on #3509 (W5a).** It is not merged yet, so this branch is based on `feat/3364-w5-image-source-selector` and the diff below includes W5a's commit. I will rebase onto `main` once #3509 lands, and auto-merge is not enabled until then.

## Design

- **Producer identity.** `ImageSetManifest.producer` is now one of two shapes. A release keeps its existing flat form: repository, workflow, tag and source commit, so release manifests serialize byte-for-byte as before and the pack-signing smoke fixture parses unchanged. A local set's producer is `{"local_checkouts": {"images": {commit, worktree}, "mvm": {commit, worktree}}}` and nothing else. The two shapes share no field. A producer that names both, or neither, or half a release, does not parse. A local producer therefore cannot carry a workflow or a tag, and cannot be read as a release.
- **Release-only fields.** `revocation_channel`, `supersedes`, and each member's `pack_hash` and `sbom` are optional in the schema. `validate_structure` enforces them: a release must have them and a local set must not. A local set's `mvm_source_commit` must equal its recorded mvm commit.
- **Two paths, one parser.** `verify_image_set` refuses a local producer at a new `provenance` stage, before comparing anything with the lock. `verify_local_image_set` runs the same parse and structure checks as the release path, then refuses the following, in order:
  - a release producer (`provenance`);
  - identities that differ from the checkouts as re-read now (`freshness`, the image checkout first);
  - any member for another architecture (`selection`);
  - a missing required role (`completeness`);
  - any artifact that is not a regular file in the set's directory (`artifacts`). The release path now makes the same check.

  Artifact names already cannot leave the directory.
- **Host side.** `LocalImageCheckout::read_local_image_set` re-runs W5a's `reverify`. It then reads the paired mvm checkout's identity, which must be the canonical root of its own work tree, and reads `image-set.json` only if it is a regular file. `xtask/dormant-controls.toml` declares it dormant until a consumer moves onto it.
- **Shared identity type.** `RepoIdentity` and `WorktreeState` moved from mvm-build into `mvm_core::image_set`, so a manifest and a fresh probe compare as one type. The working-tree fingerprint now spells out every diff and status option that git config could change, because the Python emitter in mvm-images must produce the same bytes on another host. Both repositories pin the same fingerprint for a fixed dirty tree.

## Negative tests

- **mvm-core:** mixed, partial and empty producers; a clean worktree state carrying a fingerprint, a dirty one without, and an unknown state; a release missing its revocation channel, pack hash or SBOM; a local set carrying a revocation channel, lineage, pack hash or SBOM; a local set whose mvm commit mismatches; the release verifier given a local set, even with a lock pinning its bytes; a local file claiming a release; a stale image commit; a stale mvm fingerprint; a dirty tree that is now clean; the wrong architecture; mixed architectures; a missing role; traversal names (`../escape`, `/etc/passwd`, `nested/file`, `..`, `.`); a symlinked artifact on both paths; a directory in place of an artifact; a changed artifact; a missing artifact; unparseable bytes; a disjoint protocol range.
- **mvm-build, on real git checkouts:** stale after an mvm edit, after an mvm commit, or for another recorded image state; an image edit, caught by `reverify`; a missing role; the wrong architecture; a symlinked manifest; a symlinked artifact; a missing manifest; an mvm path that is not a checkout root.
- **Emitter contract:** a literal copy of the W5b emitter's output parses and verifies.

## Evidence

- `cargo nextest run -p mvm-core -p mvm-build -p mvm-cli`: 5541 passed.
- `cargo clippy -p mvm-core -p mvm-build -p mvm-cli --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- `cargo run -p xtask -- check-all`: 68 of 70 gates passed on the first run. `check-stubs` and `check-ir-parity` failed on ENOSPC (the shared host's disk filled up) and passed when rerun. `check-declared-backing`, `check-dormant-controls` and `check-witness-citations` are clean.
- End to end: `runtime-overlay.default` was built on Linux against a clone of this tree at `11e5ff4` through `--override-input`, then emitted, then read on macOS by this code. It was accepted at `local-dev`, and each variant was refused at its stage: wrong architecture, missing role, a rewritten release producer, a symlinked artifact, and a stale mvm tree. A set emitted on Linux from a dirty mvm tree was accepted on macOS against the identical edit, so the Python and Rust fingerprints agree across git 2.43 and 2.53.

Records: the plan ticks W5b and W5c and gains a local-manifest design bullet. Also updated: `specs/REFACTOR-STATUS.md`, `specs/SPRINT.md` (next to the W5a sentence), and `specs/sprint/delivery/3364-w5c-local-manifest.md`.


auser:
Superseded: every commit here landed in main through #3521 (W5d), which was stacked on this branch and merged first. `crates/mvm-core/src/image_set/local.rs`, `image_set/tests/local.rs` and the W5c delivery note are byte-identical on main; `crates/mvm-build/src/image_source/local_set.rs` is there too, extended by W5d. Nothing in this PR is missing from main, so closing it rather than re-queueing an empty change.

PR #3515, opened 2026-09-19
test(core): keep the scope timeout launcher identity

auser:
## Summary
- keep the executable blocking systemd-run fixture at the absolute path already selected by the scope binder
- stop overwriting it with an unrelated program whose process name is interpreted as a successful payload exec
- correct the sprint, delivery, and refactor records after both Linux workspace lanes exposed the regression

## Testing
- cargo fmt --all -- --check
- just dev-cargo test -p mvm-core spawn_scope::tests:: (47 passed)
- just dev-cargo clippy -p mvm-core --all-targets -- -D warnings
- pre-commit stable cargo clippy --workspace -- -D warnings
- pre-commit stable cargo clippy -p mvm-core --all-targets -- -D warnings

Follow-up to #3477 and #3486. The Linux-only process-table witness runs in CI.


auser:
Superseded by #3516, which fixes the underlying pre-exec command-name race and includes the Linux stress evidence. The fixture overwrite here is deliberate: it provides an executable named systemd-run whose process stays alive for the deadline witness, so this narrower fixture change should not merge.

